PR #130, opened 2015-03-13
Fccubic

ceriottm:
We have worked a lot on the FCC cubic order parameter, and in the process also sped up some aspects of the general evaluation of this class of order parameters. In particular we have
1) introduced more parameters in fcc cubic, making it compliant with the definition in Angioletti et al
2) introduced the possibility of rotating the reference frame with Euler angles
3) accelerated the evaluation of PBCs


GiovanniBussi:
Hi @ceriottm I had a look at the changes on the core ( @gtribello can certainly judge better the multicolvar change). A few questions:
1. applyPbc() just perform the same as distance() except that it does not take the difference and do it on an array. Is it faster than a sequence of calls?
2. Would it be possible to implement the applyPbc() in case of non orthorombic as a sequence of calls to distance()?
3. Do you actually exploit the valarray functionalities? If not, a std::vector could be better since we use mostly vector around in plumed.
4. Do you get better performance with PBC_WHILE? I was trying it some time ago with no actual speedup.
5. Same as 4: do you find the specialized operations on vector<3> faster?
6. Are you fine with the license in case we pull your request?
Finally, I do not know why your pull request has not been built on travis. I will try to understand why.
Thanks!
Giovanni

GiovanniBussi:
@ceriottm I compiled it on travis (by pushing a fccubic branch on plumed/plumed2 repo) and it failed (https://travis-ci.org/plumed/plumed2/jobs/54268158) can you check? Thanks!

gtribello:
Hello guys

In terms of the change to multicolvar I think this is one change to the way the FCC cubic order parameter works.  This is an extension to that bit of code and I think everything is compatibility with the old way this order parameter worked.  It is my understanding that you just add some extra keywords if you want to access Michele's extension.  If the default is used then the calculation is the same as it ever was.  As such I am more than happy that this change be merged.  In fact I would like this change to be merged as it is being used in some work that Michelino and I are doing on surface tensions.  Hence, if we want others to reproduce our work its important that everything be there.

To be clear I am not saying this is urgent but I would like to get this done soon.  I guess the cleanest way to do it is via the pull request but if you have concerns about some of the pbc stuff I can just merge the changes in multicolvar.  What do you think @GiovanniBussi

GiovanniBussi:
I would prefer not to merge changes that are not necessary. I guess @ceriottm optimized PBC but I was asking for a confirm since I tried to do the same in the past with no success. Namely, the specialized vector methods were slowing down the calculation on some compiler (perhaps on my mac). Also, I would avoid the valarray in Pbc since we do not use valarrays anywhere else, unless there is a reason to do it. @gtribello can you easily merge only the important part? Additionally, can you double check the regtests? One of them is failing on travis-ci. Thanks!

ceriottm:
Heya,  sorry for being unresponsive on the regtest fail. It is probbly a
trivial check, but learning how to run regtests with plumed is not
completely trivial, and will not have time to look into that for another
couple of weeks.
Just for the record, the FCCCUBIC patch also changes what is computed: the
implementation was "wrong" (i.e. not consistent with the Angioletti et al.
PRB). Now it is. The previous implementation was also giving ugly "0.002"
numbers for a perfect fcc. Now this yields 0 for a liquid, and one for fcc
as it was meant to be. However this might explain why the regtest fails
since the numbers have changed.
Re the PBC, I checked it quite extensively, and it is much faster, because
it applies them to a whole bunch of deltaij at the same time. Re the
valarrays, arguably almost all of the std::vector<primitive type> should be
changed into valarrays - they can be considerably faster on some
implementations of the STL, but I can understand if you prefer to sacrifice
speed for internal consistency.
Ciao!

GiovanniBussi:
I did some test and the "while" implementation is actually slightly faster for orthorombic boxes.
I made the change directly in master (b46d0793da). I will merge the rest in the following days.

ceriottm:
Good. I am discussing with Gareth about several other changes that make the
multicolvar implementation more efficient. Also, I tested that in actuality
std::vector is exactly as fast as std::valarray, so I have made that change
also if you prefer to use vectors for consistency.
Thanks,
Michele

gtribello:
Does that mean you are planning to merge this whole pull request?  I was trying to put in the modifications manually to Fcccubic.cpp but I am having some difficulties.

GiovanniBussi:
I made a rebased branch (fccubic-rebase1) which contains four commits:
33c5d00 "while" implementation
perhaps we should use a macro such as __PLUMED_PBC_WHILE instead of PBC_WHILE

2a18cfb1 all the rest
b286de valarray -> vector
these two can be merged

9f27da changes in Vector
I would merge this only if it actually improve speeds

Notice that one of the regtests fails, please double check it and (maybe) reset it before merging.

Thanks!

GiovanniBussi:
@gtribello the failing regtest is crystallization/rt-simplecubic

gtribello:
Right I had a look at this.  I think the failing regtest happens because the way that the fcccubic cv is defined has change.  At the same time though I think the new version is more in line with what was done in the Angelotti paper this came from so the change is probably a good thing.  @ceriottm is that correct?  I checked the numerical derivatives for this revised CV and they come out correctly.  I'll come back to that in a moment though.

I made a commit in this branch.  This is a small cosmetic change but it ensures that the default values of some of the parameters in this CV are mentioned in the manual.

On the use of the pbc apply functionality I have a couple of comments:

(1)  With the change as made it is not possible to calculate the derivatives numerically as when you change the box there is no way to apply the pbc correctly.  To calculate the numerical derivatives I had to revert back to the old way this was working (I did not put this reversion in my commit)  

(2)  There is now no way to run this without PBC.  PBC are always applied.  At the same time, because this is a multicolvar, there is a nopbc flag.  This one can be resolved by either removing the nopbc flag or 
by ensuring that applyPbc is not called when you run without pbc.

ceriottm:
Hello,
  so (1) I confirm that the FCCCubic is now consistent with Angioletti et
al, which was not in the previous implementation. (2) Can we modify
applyPbc so that it returns doing nothing if PBCs are disabled? (3) I had
benchmarked the template specializations in Vector, and with g++ -O3 they
made quite a difference. Actually I only specialised the calls that had a
significant weight in the cost of a (multicolvar+fcccubic) calculation -
perhaps it would be good to do that throughout, as I can't see how it can
hurt.
Ciao, and thanks for helping out integrating these changes. I am warming up
a lot to plumed2 so I hope to be able to give more substantial
contributions in the future.

GiovanniBussi:
Wow, it's incredible that compilers cannot optimize a loop with constant number of iterations... Probably I tried with clang and intel, which are a bit smarter. Anyway you are right, it could be better to make explicit specializations for all the methods. Thanks!

GiovanniBussi:
This has been merged already so I close the request

PR #177, opened 2015-11-09
New "switch function" to support native contact biasing

jandom:
hi guys, 

this is my first contribution to plumed2. it allows to use fraction of native contacts biasing in plumed, by adding an new switch function with the CMAP. this couldn't be implemented with matheval, i think. the native contacts biasing is fairly popular and i think could attract some more people to plumed2, which is a great and easy to use tool. 

thanks,

- Jan


GiovanniBussi:
Hi @jandom (cc @all)

I think this could have been done with matheval, though of course the c++ implementation is going to be faster and easier to use. A few things that should be done perhaps before inclusion:
1. Add a regtest, otherwise it is difficult to know when someone else will break that feature. I can help you if you don't know how to do it. Perhaps the easiest way is to modify slightly one of the tests already available.
2. I think the way you treat cases with r<0 are not consistent with the rest. All the other switching functions return 1 one in that case. I guess this is not possible in this case, perhaps we should add some comment in the manual.
3. Do you have a reference for this choice?

Giovanni

GiovanniBussi:
Great! About the behavior when r<d_0, perhaps it is better to keep the usual plumed logic stating that for r<d_0 the function is flat. Looking at the paper you mentioned, it looks like a better choice for syntax would be:
Q BETA=5 LAMBDA=1.2 R_0=0.1 REF=XXX
here REF is the native distance, and R_0 would allow setting parameters in Angstrom, and XXX should be the native distance in Angstrom. This would also fix the issue with r<d_0, and would allow perhaps BETA and LAMBDA to have reasonable defaults.
What do you think?

jandom:
This is an excellent idea!

jandom:
hi Giovanni, quick question:

Introducing the lambda/beta and ref as new class properties was trivial, and so was the read-in of inputs. But it still find a clean way to around this block of code SwitchFunction.cpp

```
  const double rdist = (distance-d0)*invr0;
  double result;
  if(rdist<=0.){
     result=1.;
     dfunc=0.0;
  }else{
```

The 'rdist' and subsequent if-check on it has no equivalent in the native-q land, as far as I can see. What's your thinking?

GiovanniBussi:
I think that within Q d0 should be always set to zero. You could also add a check and report an error if someone does it differently, for safety.

Additionally, it could be safer to replace the test "rdist<=0.0" with "rdist<0". All the current switching function should be already regular in the limit rdist->0, and this would make also Q work properly in the (possibly stupid) case of two atoms at distance zero.

Does this make sense?

jandom:
Hi there - yeah, that was very clear, thanks! Let's see if the commit i just submitted covers that.

jandom:
Woops, just noticed broken regressions - will fix asap

jandom:
hi there sorry to bug - is there anything else (aside from documentation) that i need to do in this thread? it's nothing urgent but it's just not clear to me if i can sit back and relax? ;)

jandom:
This regression test is broken but it wasn't me
```
basic/rt29/report.txt
```
All is done here, I think, including changes to the documentation  and citation added

jandom:
Thanks for all the help Giovanni!

PR #201, opened 2016-10-20
PBMetad with HILLS written on disk @maxbonomi

marcodelapierre:


marcodelapierre:
Hi guys, I don't know much about regtests/Travis. 
If you tell me what I have to do, or where to find relevant information, I will complete the required actions.
Cheers, Marco

GiovanniBussi:
@marcodelapierre Whenever a pull request is opened, travis automatically performs the regtests. In this specific case, you should see a red cross close to your commit (in this page) indicating that some test is failing.

Results for pull requests are [here](https://travis-ci.org/plumed/plumed2/pull_requests). Your commit seems to break some regtests running with MPI, see [here](https://travis-ci.org/plumed/plumed2/builds/169132703)

marcodelapierre:
Thanks @GiovanniBussi , 
I had modified the keyword for MPI Multiple Walkers for PBMetad from MULTIPLE_WALKERS to WALKERS_MPI (uniform with normal Metad, permits to avoid confusion with disk writing version). 
This change is now reflected in the regtests.

marcodelapierre:
Hi @maxbonomi , all the required regtests have now been completed successfully. (sorry if I am cross-posting you, I don't know whether GitHub had already sent you a notification or not)

carlocamilloni:
Hi @marcodelapierre, could you also add a regtest specific for this? maybe you have already done it, but I cannot find it.

GiovanniBussi:
Hi @carlocamilloni @maxbonomi @marcodelapierre 

Currently it is not possible to add a reproducible regtest for file-based multiple walkers. Also METAD does not have it. The reason is that there is no way to be sure that processes wait for each other in writing.

I was planning to make a "type=make" regtest whereone can write a small C++ program (without using driver) in the spirit of basic/rt-make-7. This would allow to:
1. Initialize two Plumed objects
2. Run one for one step
3. Flush all files and perhaps "sleep" a couple of seconds
4. Run the second object for one step.

This should allow to regtest also METAD and PBMETAD with walkers.

If the things Marco added do not break other features and if Carlo and/or Max can confirm they checked the code, I would suggest to merge it even without a specific regtest. When I will make the test for METAD I will ask then Marco to make an equivalent test for PBMETAD.

Does this make sense?

marcodelapierre:
Hi @carlocamilloni @maxbonomi @GiovanniBussi ,
I would just add that tests under regtest/basic/rt-mpi6c/ seem to make use of file-based METAD.
However, as I am not familiar with the suite of regtests that ship with plumed, I have no idea whether these tests are actually run or not by Travis.
If they are not, then just contact me when the time comes of adding the new regtest as mentioned by Giovanni.
On the other hand, if it is the time now of adding the test, I just have a quick question: is there a rationale for putting it in any specific folder under regtest?
Thanks, Marco

GiovanniBussi:
@marcodelapierre @carlocamilloni @maxbonomi 
Marco you are right but that test is not actually testing anything useful (just a colvar file).
The reason is that if you try to make a regtest with multiple walkers without ensuring synchronization it will fail from time to time.

About regtests organization, it is mostly historical. In v2.2 specific optional modules (src/xxx) needed specific regtests (regtext/xxx). Now the choice is more free, in the sense that you should specify the modules that you need in the regtests config file. This allows a single regtest to specify that it needs two modules to be active (see e.g. crystallization/rt-urea/config). Anyway, I guess we will try to keep it more or less consistent with the names of the optional modules.

marcodelapierre:
@marcodelapierre @carlocamilloni @maxbonomi 
Thanks for the clarifications Giovanni. Then just let me know when the time for the new regtest comes. Also, do not hesitate to come back to me in case you have any queries about the code I wrote.

GiovanniBussi:
@marcodelapierre @carlocamilloni @maxbonomi

Related to this old (merged) pull request, I wanted to let you know that I added a regtest for METAD with file based multiple walkers, perhaps you want to do the same for PBMETAD. I did it in v2.2 then merged to v2.3 and master, but for PBMETAD it might require master branch to have multiple walkers enabled.

A few notes below:

- I added a few lines to MetaD such that hills files are flushed with stride MWALKERS_RSTRIDE (26152b244505221b3257de6e8e651acf42f057bf). I think this is safer (before this it was necessary to use FLUSH explicitly). You might want to do the same in PBMETAD.

- The relevant regtest is basic/rt-make-mw. To allow explicit synchronization between different walkers it does not use the driver but a short, ad hoc main.cpp file. However, it should be easy to adapt it. To check whether hills are correctly read, it prints the total bias in each walker on a COLVAR file.

Notice that the flush is required to make the test reproducible. So, either you add the flush to PBMETAD (recommended) or you put a FLUSH command in the plumed input file.

Giovanni

PR #205, opened 2016-10-31
Switchfunction none

jandom:
Hi guys, 

This is a feature that i found pretty useful:  when i started with intermolecular DRMSD, I wanted to bias on the interhelix distances in glycophorin helix-helix dimer. This takes a PDB with 2 TER blocks, one per monomer. 

Then I realized that some people might want to specific a distance list, rather than pass structures, for this feature. Switchfunction-none is how i approached it. 

- [ ] I verified that all regtests are passed succesfully on Travis.


src/tools/SwitchingFunction.cpp
```diff
@@ -305,6 +308,9 @@ double SwitchingFunction::calculate(double distance,double&dfunc)const{
       double tmp1=std::tanh(rdist);
       result = 1.0 - tmp1;
       dfunc=-(1-tmp1*tmp1);
+    }else if(type==none){
+      result = rdist * rdist;
+      dfunc = rdist;
```
OleinikovasV: I believe on line 313 it should be:
dfunc = 2.0 * rdist;

OleinikovasV: Also, this won't behave "like DRMSD" whenever distance is less than the reference:
rdist=0 as well as rdist<0. return 1. immediately. instead of desired zero and a (negative side) parabola, respectively.
Also, if(distance>dmax)  - 0.0 is returned rather than high value of DRMSD component.


Would be very useful if one could to that though...

GiovanniBussi:
Hi! To me it is a bit difficult to understand the purpose of this, for the following two reasons:
- you can do this with MATHEVAL
- in plumed there are assumptions that switching functions go from 1 to 0, that are obviously violated in this case.

A better choice would be
1-(d/dmax)**2

Differences in this function are proportional to differences in d**2 by a factor 1/dmax**2

Would this fit?

Giovanni

src/tools/SwitchingFunction.cpp
jandom: That's probably correct, yes ;-)

jandom: Agreed – this is a valid comment

jandom:
hi guys, thanks for reviewing this. i'm not super-familiar with matheval, so can't comment. all that matters to me is that this is used for this useful native-contact thing, so which ever way it's presented I'm cool with it.

GiovanniBussi:
By linking libmatheval you can use arbitrary switching functions, e.g.:

     SWITCH={MATHEVAL FUNC=x^2}

In any case, I would recommend a switching function that goes from 1 to 0. E.g., as I wrote above:

     SWITCH={MATHEVAL FUNC=(1-(x/100.0)^2) D_MAX=100.0}

Here, 100.0 is the maximum distance considered, at which switching function will be zero.

If you compute a difference between switching functions like these ones, the value will be what you want scaled by a factor 1/100^2. So you will just have to scale the reference values by that factor.

If you think this is useful, we might include a function (1-(x/dmax)**2)

Giovanni

GiovanniBussi:
Actually, you can already obtain it using

    NN=2 MM=100000

That is what you should do is:

    CONTACTMAP ...
    ATOMS1=1,2 REFERENCE1=0.1 WEIGHT1=0.5 
    ATOMS2=3,4 REFERENCE2=0.5 WEIGHT2=1.0 
    ATOMS3=4,5 REFERENCE3=0.25 WEIGHT3=1.0     
    ATOMS4=5,6 REFERENCE4=0.0 WEIGHT4=0.5 
    SWITCH={RATIONAL R_0=100.0 NN=2 MM=1000 D_MAX=100.0} 
    LABEL=cmap
    CMDIST
    ... CONTACTMAP

The reference values should be obtained as:
    
    1-refdist/100**2

where refdist is the reference distance. Here, 100.0 is a large number. You can use a smaller number to ignore differences in distances larger than this.

So I close this pull request for the moment, feel free to reopen it if you think that it is important to have an additional feature for this.

Giovanni

PR #208, opened 2016-11-23
Marco De La Pierre: added WALKERS_DIR option for WALKERS_MPI

marcodelapierre:

### DIFF
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -432,7 +432,7 @@ void MetaD::registerKeywords(Keywords& keys){
   keys.add("optional","INTERVAL","monodimensional lower and upper limits, outside the limits the system will not feel the biasing force.");
   keys.add("optional","SIGMA_MAX","the upper bounds for the sigmas (in CV units) when using adaptive hills. Negative number means no bounds ");
   keys.add("optional","SIGMA_MIN","the lower bounds for the sigmas (in CV units) when using adaptive hills. Negative number means no bounds ");
-  keys.addFlag("WALKERS_MPI",false,"Switch on MPI version of multiple walkers - not compatible with other WALKERS_* options");
+  keys.addFlag("WALKERS_MPI",false,"Switch on MPI version of multiple walkers - not compatible with WALKERS_* options other than WALKERS_DIR");
   keys.addFlag("ACCELERATION",false,"Set to TRUE if you want to compute the metadynamics acceleration factor.");  
   keys.use("RESTART");
   keys.use("UPDATE_FROM");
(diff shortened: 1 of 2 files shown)


marcodelapierre:
Hi guys, this is a small but useful mod in my opinion. 
I don't know anything about typical use of PLUMED with other MD codes, but in the case of LAMMPS we usually run replicas for MW-MetaD/PBMetaD in distinct folders.
Now, when WALKERS_MPI is used, all replicas need access to the HILLS files at restart time. Then it comes handy to be able to specify their path in the input file.
Marco

GiovanniBussi:
Hi,

I have two comments:

- I am not sure this is needed. Can't you just use the following?

METAD FILE=/path/HILLS

- I think this is going to break cases where FILE is specified with full path.
   In other words, if you use

METAD FILE=/path/hillsfile

PLUMED will check for ./path/hillsfile which does not exist.
Notice that we have currently no way to regtest with absolute paths
since this would require the input file to be adjusted based on the
location of the regtest directory. That's why you don't see failures.

Users should change the input file to the following:

METAD FILE=hillsfile WALKERS_DIR=/path

This is counterintuitive since you should do it even when you don't
have multiple walkers.

Given these two points, I would prefer not to merge.

@marcodelapierre what do you think?

Giovanni

marcodelapierre:
Hi @GiovanniBussi , thanks for the reply.

- The reason why we need to specify a path for WALKERS_MPI is that when using Parallel Bias MetaD there is a HILLS file per CV, and one can in principle use a relative large number of CVs, so that having to specify the path for each file name becomes annoying.
- On the other hand, I agree that when there is just one replica/one hills file one might want to specify the full path in the fill name. Also, we are fans of relative paths, but I have just realised that other users might be absolute path fans.

Therefore, I have updated the pull request with a new implementation for both MetaD and PBMetaD, where:
- when a single replica is adopted, the path is always absolute (as it was before);
- when multiple walkers are used, either disk or MPI, the default is absolute path, but using WALKERS_DIR permits to switch to relative paths.

Let me know whether this sounds more sensible (and useful) to you.

Marco

marcodelapierre:
The 4 Mac builds on Travis are failing, but from the log file it looks like a machine issue, as the github server is not reachable.

marcodelapierre:
Hi @GiovanniBussi ,
I had forgotten about the failed tests; they have now all passed.
Let me know if the branch is now good for merging, with the implementation modified according to your comments, as described above.
Marco

GiovanniBussi:
@maxbonomi I wanted to merge this. I am ok with the changes to MetaD. Could you check if the changes in PBMetaD are also fine? Thanks! Giovanni

PR #216, opened 2017-03-03
Dimer CV

marckn:
CV for Dimer Metadynamics (JCTC J. Chem. Theory Comput., 2017, 13 (2), pp 425–430).
Gromacs can be used as MD driver for the simulation and I've built a python script (available on a different repository) that takes a standard Gromacs simulation setup and produces a dimerized one. 

- [X] I verified that all regtests are passed succesfully on Travis.
- [X] I agree to transfer the copyright of the code I have written to the PLUMED developers.


GiovanniBussi:
Hi @marckn excellent job!

I have a small doubt. You compute distances with pbcDistance() and later use setBoxDerivativesNoPbc(). The latter assumes that PBCs are not used, so I think you might have errors in the virial calculation when atoms interact through images. Perhaps you can try to generate test positions to probe this: just shift atoms randomly by multiples of the cell box and redo the calculations. Box derivatives will change, though they shouldn't.

Can you double check?

Then I have a minor suggestion: you might use the standard way to incorporate citations, that is add something like:

log<<"  Bibliography "<<plumed.cite("Laio and Parrinello, PNAS 99, 12562 (2002)")<<"\n";

In this manner all the references are written together at the end of the input.

marckn:
Thank you @GiovanniBussi 
The citation has been fixed (I've yet to push). For the box derivatives it's not clear to me how to use the setBoxDerivatives*() methods so I checked how they are used in the distance CV. This CV is basically a function of the distance so I assumed that it should work the same for me. 
I tested by comparing analytical and numerical derivatives for some configurations coming from a trajectory, I'll surely check whether it works also when the dimer is crossing a boundary.

GiovanniBussi:
@marckn If you use pbcDistance() then you cannot use setBoxDerivativesNoPbc().

setBoxDerivativesNoPbc() sets virial to:
-x_i * ds/dx_j

whereas if you use pbcDistance's you should, for each distance that you evaluated, contribute
-x_ij * ds/dx_ij

This is done for instance in colvarCoordinationBase.cpp

Giovanni

marckn:
@GiovanniBussi O.K. it is related with the use of pbcDistance() rather than the fact that there might be pbcs. In Distance.cpp setBoxDerivativesNoPbc() is used and in fact pbcDistance() is not. 
I'll make the necessary modifications and check the derivatives in a more thorough test. Thanks for the assistance.

PR #217, opened 2017-03-20
Experiment Directed Simulation

whitead:
This adds the Experiment Directed Simulation (EDS) method described in [An Efficient and Minimal Method to Bias Molecular Simulations with Experimental Data](http://pubs.acs.org/doi/abs/10.1021/ct500320c). It is written as a contributed module according to the developer documentation.


GiovanniBussi:
Thanks for fixing the namespaces.

Just a tiny extra change: you should add a file named COPYRIGHT (perhaps containing the header that you used in Eds.cpp) in the src/eds directory. Otherwise, next time one of us updates the headers your one will be overwritten.

whitead:
No problem! Committed the change.

PR #224, opened 2017-04-10
EDS: Added ability to have other CVs/Actions as centers

whitead:
Made a change to EDS to allow centers specified as actions instead of only doubles. Also removed duplicate line in restart files at step 0.


whitead:
OK good idea. I'll update that. Is there a place in plumed now I can see an example of best practice? Maybe the write grid file?

GiovanniBussi:
If you do

     cd regtest
     grep FMT */*/*.dat

you should see several examples. To be sincere there are many existing tests where we did not fix this issue, and from time to time they fail (with newer compilers). I already noticed that several of the current tests fail with gcc6 or intel and -O3, at some point we should probably fix them. If you fix yours now it will be easier later

Giovanni

whitead:
I added a FMT specifier to the restarts, so hopefully the precision issue is resolved.

PR #230, opened 2017-04-30
Coordination Number R_POWER flag

whitead:
This adds the coordination number moments code from [this recent article](http://aip.scitation.org/doi/abs/10.1063/1.4974837) to the existing COORDINATIONNUMBER CV. Thanks to @gtribello for writing the initial implementation.


GiovanniBussi:
Just a short comment. @whitead do you know that switching functions are stretched to zero by default? I think there should be no discontinuity anyway. Additionally, I am not sure it is a good idea to change D_MAX. I guess @gtribello has a clearer idea on this.

whitead:
The switching functions are set to stretched to 0.000001 at their endpoint. If you have r^p * 0.000001 it can be large depending on your unit system. Happy to hear an alternative approach though.

GiovanniBussi:
As far as I remember switching functions are set to be _exactly_ zero at the point where they would be 0.00001 (unless you use NOSTRETCH). So, there should be no discontinuity also with the r**p factor. Can you double check if this is correct?

A different point is that maybe you want to have a larger d_max so as to not to discard the peak of the f(r)*r**p function. Is this the reason perhaps?

GiovanniBussi:
@whitead as an additional note on your pull request, the travis-ci job associated to cppcheck is failing (see [log here](https://travis-ci.org/plumed/plumed2/jobs/227256022)).

One warning is about a variable not initialized in a constructor and should be easy to fix. About `astyle`, it is explained [here](https://plumed.github.io/doc-master/developer-doc/html/_code_formatting.html). You should just type `make astyle` from the root directory then `git commit`.

About my comment on stretched functions above, I am afraid that the default is still *not* to stretch when the implicit syntax is used. However, when you add an explicit switching function ({RATIONAL ...}) the default should be to stretch it to zero. Perhaps we should make it homogeneous in the future (and make stretch the default everywhere). Anyway, I saw that in the regtests that you submitted you used the explicit syntax, so they should have no discontinuity irrespectively of the value of the power.

whitead:
I see what you mean about using stretch on the end-point; it will be smoothly approaching zero. I'm not as concerned about the behavior at exactly zero though, but I can see how what I wrote could interfere with the stretching calculation. 

Let me give you a specific example. Let's say I choose the rational switch function with n = 3, m =6. If I then choose a power of p=4, the CV will definitely not be approaching zero at the endpoint (since r^3 * r^4 / r^6 ~= r). This case is easy, but if you have p=2 you have slower decay and there are some sets of parameters (D_0, R_0) where the switching function will not be as close to 0. I was trying to account for this by extending (via the set_dmax), but I see your point about how that can interfere with the stretching.

Maybe the best solution is just to get sw(dmax * 0.999) * r^p and report to the user so they can make a decision about if it's too large.

whitead:
I think I got astyle to finally work, but now it looks like there is an unrelated issue on Travis.

GiovanniBussi:
Hi @whitead don't worry about the failure on travis-ci (that's a problem on the MacPorts installation that I am trying to solve).

About discontinuities, notice that the function `r^p*f(r)` is still not discontinuous. It just has a large first derivative.

Finally, could it make sense to use `(r/r0)^p` instead of `r^p`? `r0` could either be taken from the switching function or be a separate keyword for `COORDINATIONNUMBER`. I was wondering if this could make the dependence of the results on the employed units easier to track. Anyway, I am not sure if this is necessary or even useful.

About merging, better if @gtribello takes care of that since he is more familiar with the multicolvar framework.

gtribello:
Hello.  Sorry for my slow replies on this.  I had a cold over the weekend that was really kicking my arse.  

Anyway, I think most of this looks fine.  The only thing that worries me slightly and the thing I would like to know before we merge is how much this is affecting the performance of the CoordinationNumbers Action?  I think this change will slow it down and I was wondering by how much?  Have you done any tests on the timings Andrew?  If it slows it down a lot maybe it is better to put this in a separate action.  Thanks

whitead:
You guys drive a hard bargain for this PR ;) I've added another commit so that you only lose 0.1% performance instead of 2% (see timings below)

Old CN:
```
PLUMED:                                               Cycles        Total      Average      Minumum      Maximum
PLUMED:                                                    1    15.252420    15.252420    15.252420    15.252420
PLUMED: 1 Prepare dependencies                         14978     0.012712     0.000001     0.000001     0.000012
PLUMED: 2 Sharing data                                 14978     0.053160     0.000004     0.000003     0.000025
PLUMED: 3 Waiting for data                             14978     0.011048     0.000001     0.000001     0.000019
PLUMED: 4 Calculating (forward loop)                   14978    14.835758     0.000991     0.000926     0.002317
PLUMED: 5 Applying (backward loop)                     14978     0.096334     0.000006     0.000006     0.000024
PLUMED: 6 Update                                       14978     0.056798     0.000004     0.000003     0.000023

```

New CN (e8d05e0):
```
PLUMED:                                               Cycles        Total      Average      Minumum      Maximum
PLUMED:                                                    1    15.579044    15.579044    15.579044    15.579044
PLUMED: 1 Prepare dependencies                         14978     0.013564     0.000001     0.000001     0.000012
PLUMED: 2 Sharing data                                 14978     0.050704     0.000003     0.000003     0.000028
PLUMED: 3 Waiting for data                             14978     0.011014     0.000001     0.000001     0.000012
PLUMED: 4 Calculating (forward loop)                   14978    15.162471     0.001012     0.000935     0.002750
PLUMED: 5 Applying (backward loop)                     14978     0.098153     0.000007     0.000006     0.000028
PLUMED: 6 Update                                       14978     0.060470     0.000004     0.000003     0.000027
```

Adding `if` so new code is only called when R_POWER > 0 (latest change 940936d0c):
```
PLUMED:                                               Cycles        Total      Average      Minumum      Maximum
PLUMED:                                                    1    15.279786    15.279786    15.279786    15.279786
PLUMED: 1 Prepare dependencies                         14978     0.013802     0.000001     0.000001     0.000018
PLUMED: 2 Sharing data                                 14978     0.052827     0.000004     0.000003     0.000063
PLUMED: 3 Waiting for data                             14978     0.011285     0.000001     0.000001     0.000061
PLUMED: 4 Calculating (forward loop)                   14978    14.854964     0.000992     0.000920     0.002875
PLUMED: 5 Applying (backward loop)                     14978     0.099443     0.000007     0.000006     0.000082
PLUMED: 6 Update                                       14978     0.063753     0.000004     0.000003     0.000054

```

gtribello:
Thank you.  I have merged these changes now

PR #233, opened 2017-05-08
Adding Transition Tempering to METAD

JFDama:
These commits add transition-tempered metadynamics as an option to the METAD action. See http://pubs.acs.org/doi/abs/10.1021/ct500441q for the paper introducing the method. It is designed as a robust complement to well-tempered metadynamics.

The algorithm builds on a base option, CALC_TRANSITION_BIAS, that enables the METAD action to track estimates of the amount of bias added to transition barriers over the course of simulation. This is implemented via a barrier finding algorithm adapted from Dijkstra's algorithm added to Grid.h and Grid.cpp. There is a special branch of this algorithm for the case of 1 dimension because pathfinding is trivial in 1 dimension. 

Given that barrier bias estimate, call it V*, transition tempering rescales new hills by a factor dependent on the global V* instead of the local bias as in well-tempered metadynamics. There are three parameters. The most important and the only one required is TTBIASFACTOR, which is just like BIASFACTOR for well-tempered. The other two, TTBIASTHRESHOLD and TTALPHA, are optional. TTBIASTHRESHOLD is a delay parameter; when it is nonzero, V* is replaced by max(0, V* - TTBIASTHRESHOLD) everywhere in the algorithm. TTALPHA is a caution parameter to prevent false apparent convergence. An exp(-V*/Delta T) rule makes hill heights scale as 1/t, and setting TTALPHA between 1/2 and 1 changes this to 1/t^TTALPHA. A lower TTALPHA improves robustness against false convergence by shrinking the hills less aggressively.

These commits include regtests for the transition bias tracking in 1D and 2D (both primary branches of the pathfinding algorithm) and for transition tempering with each of the three parameters set to non-default values.


carlocamilloni:
@GiovanniBussi @gtribello @maxbonomi @JFDama  I have checked the code and to me it looks ok. I am happy to merge it. For the future we should think about how to reorganise the MetaD also to minimise the code duplication with PBMetaD and make it overall more readable

JFDama:
Thank you, Carlo. I was surprised the tests now fail after merge but the only test that is failing is the astyle check. This may be due to the added trailing space on line 398 of MetaD.cpp.

carlocamilloni:
Hi

Yes that is now fixed

Carlo

Sent from my iPhone

JFDama:
Perfect, all tests are passing now. Thanks again.

GiovanniBussi:
@carlocamilloni @gtribello @maxbonomi I did not have time to go through the code but if Carlo is ok with it I am happy to merge it as well

gtribello:
This is also fine with me.  If @carlocamilloni  is happy then we can merge.

carlocamilloni:
Thanks!

JFDama:
Thank you all! Great to have this in there.

PR #234, opened 2017-05-17
METAD ACCELERATION keyword fixes

JFDama:
The proposed commits make three fixes to the "acc" component of the METAD action. First, the initial value is set to 1.0 rather than uninitialized or 0.0. Second, calculating acceleration now works with multiple time stepping. Finally, RESTART functionality is added in the form of an ACCELERATION_RFILE keyword that allows the "acc" component to be read from the end of a previously written COLVAR file.

A new regtest, basic/rt-acc-1, tests that acceleration factors are calculated correctly with and without restart.


GiovanniBussi:
To me it looks fine. @carlocamilloni can you double check? Thanks!

carlocamilloni:
Thanks!

PR #248, opened 2017-06-19
Fix 241

GiovanniBussi:
I open a pull request so as to allow comments into changed files


src/core/ActionAtomistic.h
```diff
@@ -42,6 +42,7 @@ class ActionAtomistic :
 
   std::vector<AtomNumber> indexes;         // the set of needed atoms
   std::set<AtomNumber>  unique;
```
GiovanniBussi: @carlocamilloni I would add a comment here:
/// unique should be an ordered set since we later create a vector containing the corresponding indexes

To avoid that in the future we replace it by mistake with an unordered_set

src/core/ActionAtomistic.cpp
```diff
@@ -77,7 +77,8 @@ void ActionAtomistic::requestAtoms(const vector<AtomNumber> & a) {
 // only real atoms are requested to lower level Atoms class
     else unique.insert(indexes[i]);
   }
-
+  updateUniqueLocal();
+  atoms.unique.clear();
```
GiovanniBussi: @carlocamilloni why is it necessary to clear the atoms.unique set at this point? I've seen a number of unique.clear() around that I did not understand

src/core/Atoms.cpp
```diff
@@ -138,54 +137,81 @@ void Atoms::setForces(void*p,int i) {
...
 // Notice that this works with the assumption that charges and masses are
 // not changing during the simulation!
   if(!massAndChargeOK && shareMassAndChargeOnlyAtFirstStep) {
     shareAll();
     return;
   }
-  for(unsigned i=0; i<actions.size(); i++) if(actions[i]->isActive()) {
-      if(dd && shuffledAtoms>0) {
-        unique.insert(actions[i]->getUnique().begin(),actions[i]->getUnique().end());
+
```
GiovanniBussi: @carlocamilloni it looks you did everything correctly, but I am always confused by the shuffledAtoms flag. Maybe @gtribello can also have a look (if I recall correctly it was him finding the problem once when dd is used with all atoms but atoms are out of order, perhaps with LAMMPS and a single processor)

src/core/ActionAtomistic.h
carlocamilloni: yes no problem

src/core/ActionAtomistic.cpp
carlocamilloni: the idea is that clearing unique at every time step is much slower then just updating it, so it is cleared only upon domain decomposition update, neighbour list update, replica-exchange request. In this way there could be cases in which the number of unique atoms is higher than those actually required but is never less than that. In all cases I tested this is faster, it could be slower if we bias very few atoms every step and monitor all the atoms more frequently than neighbour list update.

src/core/Atoms.cpp
carlocamilloni: shuffledAtoms means that the local index of the atoms does not correspond to the global one

src/core/ActionAtomistic.cpp
GiovanniBussi: I see, so you removed the clear at every step, that's why you added several more clear around in the code. You say "neighbour list update" but it is `requestAtoms` right? Since in principle someone might implement his/her own criteria to request a new list of atoms

carlocamilloni: Yes is requestAtoms so that is as general as possible

GiovanniBussi:
@carlocamilloni to me it looks it can be merged. Great job Carlo!

By the way, I am happy that someone else was able to understand how this class works. It's probably one of the worst parts of plumed2.

GiovanniBussi:
@carlocamilloni sorry the pull request is on master branch (my mistake). I update it right now

PR #262, opened 2017-08-02
Python interface

gtribello:
Hello @GiovanniBussi 

This is my first pull request.  I am slightly scared as I think github read my mind!  It seems it is even more invasive of privacy than gmail, which just reads my emails.

Anyway, this is the stuff that I have been working on with @ceriottm to have an interface between PLUMED and python.  It is not perfect at the moment but it would be great if we could merge it relatively quickly as I am going to a meeting organised by Peter Bolhuis on the 15th August.  He is interested in using PLUMED to calculate CVs for his TPS code, which is written in python, if we can do that directly from the python I think we will look super organised!

There are five things that we would like some help/advice with if possible:

(1)  The Makefile:  At the moment you make the python wrapper by compiling PLUMED with configure and make as usual.  I then do:

cd plumed2/src/python
CFLAGS="-I/Users/gareth/Projects/CVception/COSMO/plumed2/src/wrapper" LDFLAGS="-L/Users/gareth/Projects/CVception/COSMO/plumed2/src/lib" python buildPythonInterface.py build_ext -i

If that could all be done in Makefile if suitable configuration options are chosen that would be great.  I really need some help and advice here though as I am very lost.

(2)  Passing the bias: We have an implementation for passing CVs that looks like this:

plumed.cmd("setMemoryForData d1",data)

this sets some memory to copy the value named d1 to at the end of the calculation.   We wondered if the total bias should not be passed through the same mechanism?  The problem here is that we would have to reserve a special name that was used for the total bias that the user would then not be allowed to call their actions.  What I mean is that you would get the total bias by doing something like:

plumed.cmd("setMemoryForData totalbias",bias) 

plumed would then understand that totalbias is the total bias.  Personally, I am happy enough the way the code is and I think this might just be adding a complexity that is not required.

(3)  Dealing with all python types:  At the moment I am not sure how well set up we are to handle every python type.  I think the code will crash if the python coder does something that is not supported.  This is perhaps annoying though.  I would like to talk to a python expert here to get some advice.

(4)  Dealing with plumed exceptions in python:  It would be nice if you could raise the plumed errors in python so that the person coding in python does not have to go through the output of the code.  I am not sure how feasible that is?  There are some paragraphs of text on the Doxygen page for the PLMD::Exception class that I have read that suggest you can pass the exception to an external code written in C++.  It may also be possible to link this all with cython but I am not sure.  Do you have any thoughts?

(5)  Regtests:  Can I call a python script using the PLUMED regtest machinery as it is at the moment?

Thanks 
Gareth


GiovanniBussi:
Hi @gtribello ! Sorry for the long delay, I was practically disconnected for a while. I had a quick look and I have a few comments:
- `PlumedMain::readInputLine` does the same as `createAction` right?
- I did not look at the code carefully, but you should make sure that plumed assumes the pointer to the data is a double/float depending on what is set in core/Atoms. Check how the conversion is made for other data.

About your questions

1. This should work automatically if you install plumed, though with a small change. Replace `#include "Plumed.h"` with `#include <plumed/wrapper/Plumed.h>` etc. If you want to work from a not installed plumed, it can be done adding the proper CPATH and INCLUDE variable (I used both in the modulefile, see lib/modulefile.in, since some compiler use one some use the other). They should be set to `plumed2/include`. Can you double check if this solves your issue?

2. In principle bias should be grabbed in the same way. However, the difference is that bias is always computed. The key point in the way you implemented it (if I got it right) is that when you call `setMemoryForData` you also activate the action. So, I think it is fine if the two things are treated differently, as it is now.

3. No idea here. I think to keep the API simple we should only allow transfer of int and double/float arrays.

4. In v2.4 by default PLUMED raises exceptions. @carlocamilloni told me this propagates correctly to gromacs. I think you should check python documentation to know if you can intercept the exception from python.

5. I guess you managed already, right?

Let me know if you need further inputs!

Giovanni

gtribello:
Hello @GiovanniBussi and @ceriottm 

Thanks, @GiovanniBussi for the response first of all and the feedback.  The situation has evolved since the last message so here is an update.  So first of all the code in the python-interface branch on the COSMO fork has changed since I made this pull request.  Does that mean I would need to delete this pull request and create or a new one or are the changes that I made to the branch also incorporated automatically into this pull request?

Anyway, in response to easy questions first:

-> Yes readInputLine is the same as createAction.  I have thus deleted createAction and am using readInputLine instead.  This works fine.  Thanks for pointing it out.

The more difficult question.

-> I set the real precision in the python constructor to 8.  I think that means I am working with doubles?  The user can, in theory, change the precision if they want but I was thinking of hard coding the precision in order to make my life simpler.  I don't really understand all this precision stuff and I have never found a satisfactory explanation.  

In terms of my questions:

1.  I haven't tried what you suggested as I really am completely lost.  At the moment I copied the standard module makefile to the python directory.  This is bad as the complilation crashes in that directory as cython generates a cpp file, which plumed can't compile as it doesn't include the python headers.  Basically I am totally lost on this front and if you had time for a Skype chat on what to do here that would be really helpful.

3.  I wanted to try to talk to Giovanni Pizzi (who is here in Lausanne) about this.  I think he knows the stuff on the python side better.  I am in agreement with you on what we should pass.  I would possibly even go further and only allow double arrays (i.e. no floats).  What I want is help making sure that I catch everything that someone coding in python might do wrong.

4.  I have made some progress on the exceptions there are still some problems, however.  As an example suppose I have a python script containing the following code:

import plumed
p = plumed.Plumed()
p.cmd("readInputLine","t1: TORSION")

The problem here is that no atoms are specified in the input to TORSION. 

Now with the interface as it is currently this would throw the following error:

libc++abi.dylib: terminating with uncaught exception of type PLMD::Exception: 
+++ Internal PLUMED error
+++ file Action.cpp, line 239
+++ message: ERROR in input to action TORSION with label t32 : VECTOR1, VECTOR2 and AXIS should specify 2 atoms each
FAILURE
FILE logfile does not exist

which is fine.  What is annoying though is that a proper python error would tell you the line number in the python code where the error occurred.  Now I can also change the interface so that it throws an error like this:

Checking for python
Traceback (most recent call last):
  File "./python-script.py", line 124, in <module>
    t32 = create_plumed_var( p, "t32", "TORSION " )
  File "./python-script.py", line 63, in create_plumed_var
    p.cmd("readInputLine", name + ": " + command )
  File "plumed.pyx", line 85, in plumed.Plumed.cmd (plumed.cpp:3247)
    self.c_plumed.cmd( ckey, <void*>cval )
RuntimeError

which is less useful as the plumed error is not returned but in some ways better as you at least are told what line of python the code went wrong on.
 
The problem is that I cannot do both things at the same time - i.e. I cannot say what line number the code failed on and what the PLUMED error message is.  

I have tried everything that I can think of here and I am totally stuck.  I have written a c++ helper class that contains a catch and except statement and tried to call it from the cython rather than calling cmd directly.  This does not work, however, as if it is called from cython the exception is never caught.  I have this code on a different branch if you are interested in looking?  I guess the question here is do you know anyone who can help as I have read a lot of stuff online and I am wiser but still very much stuck.

5.  Yup I did the regtests.

Lastly, for giggles, I found the attached image to illustrate plumed and python working together.

unknown

ceriottm:
Another thing I might add - we had this problem with restarts, where it
appears that PLUMED will not deploy a hill on the first call to it when it
is restarted, even if it is time to do so according to the stepper.  I
think this behavior is wrong in general, and is easy to fix, but I wanted
to hear from the other devs if there was some rationale behind this
behavior.

GiovanniBussi:
Hi @gtribello 

About precision
----------------

I think we should decouple the two problems:
- Allow MD codes to retrieve values of CVs
- Make a python interface

Concerning the first issue, notice that the current `cmd` interface is designed as follows:
- first argument is always a string with a command
- second argument is one of the following:
  - a NULL pointer
  - a pointer to a int or long int (this is an exceptional case for `setStepLong`)
  - a pointer to a real (double or float)
- to decide whether real pointers are double or float (by default: double) one should call cmd `setRealPrecision`.

This precision stuff is needed for gromacs (and possibly other codes) to work correctly. So the point is, if someone wants to retrieve CVs from gromacs he should for consistency find them as single or double precision consistently with gromacs.

The conversion of large arrays (positions) is dealt with in MDAtomsTyped<T>, which can be instantiated in two variants depending on the requested precision. Perhaps the pointers to CVs in the MD code should be saved here as well? The nice thing of this structure is that it is implemented as a template, so that you can directly code ignoring precision. The bad thing is that the interface of any function that you add to `MDAtomsTyped<T>` should be added to `MDAtomsBase` as well.

Pointers to individual data (such as bias, timestep, etc) are converted with Atoms::MD2double and Atoms::double2MD.

Given this, you can obviously decide that python interface is limited to double, though I don't think it is necessary. Still, the interface with float should work for MD codes to work as expected when retrieving CV values.

About exceptions
-----------------

I don't know how cython works, but I think you should be able to extract the message ones the C++ exceptions has been caught. If you do
````
try{
 cmd ...
}catch(PlumedException * e){
  std::cerr<<e.what()<<"\n";
}
````
Does it make sense?

You should be able to write the message on stderr. The right thing would be to append it to the python message, so that the exception can be in turn caught by python.

Giovanni

**Comment by GiovanniBussi** [comment]
@gtribello about incorporating new changes, if you push to the branch you are merging from (`cosmo-epfl:python-interface`) the new commits will be included in the pull request, no need to open a new thread.

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

Thanks for the replies.  On the first item with the passing of data I think I have tried to decouple the problems.  I might not have done it correctly though.   If you look at the branch I have a class called DataFetchingObject, which is a template class like Atoms.  If you then look at PlumedMain at line 307-309 you see that this is supposed to be created here with the appropriate precision (i.e. when you call setRealPrecision).  This may not be done correctly at the moment but it should be easy to fix its just a matter of setting the correct precision value rather than setting the value to be a sizeof(double).

On the exceptions, we have made progress.  I was always doing the call to Plumed inside a catch and except look as in the example you sent.  The problem is that on my computer the exception is never called for some reason.  In other words, if there is a mistake in the input the exception is not triggered.  On Michele's computer, however, this works perfectly.  What we thus need are some other computers to test this on.  We would like to know if this is a mac-specific problem or just a Gareth's-mac-specific problem.

**Comment by GiovanniBussi** [comment]
About precision: great, so it is basically done.

About exceptions: check if in your executable exceptions are enabled (`plumed config show | grep exc` should be `on`). Or it might be some python problem...

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

I had a go at writing the Makefile stuff for the python interface and I have pushed the changes on the branch for this pull request.  I would appreciate it if you would take a look and respond to the following questions:

(1)  In terms of getting the regtest for the python interface running, I have made some changes to regtests/scrtipts/test.make.  Can you look at those and let me know if that makes sense.  Essentially it is meant to check if python is installed, it then checks if the plumed module is installed in python and only if these two conditions are true does it run the test.

(2)  I build the python interface after the vim wrappers are built.  As such I can do everything with only minimal changes to the main plumed makefiles.  I guess this is OK?

(3)  In terms of the changes to autoconf I have that the python interface is compiled by default.  I then  check that python is installed and cython is installed.  If both these codes are there I go ahead and tell the makefile to build the plumed python interface.  Do you think there is anything else I should be doing?

(4)  To trigger the building of the python interface I add the CPP flag -D__PLUMED_HAS_PYTHON.  In the makefile in python I then check if that flag is in the CPPFLAGS.  I did it this way because (a) it minimised the amount of automake and make I had to understand ;-) and (b) I thought this made sense as if someone does something in the future that links python as a library through c++ they might want to use -D__PLUMED_HAS_PYTHON as a preprocessor directive.  Yes argument (b) is probably bollocks and really it was all argument (a) so if you can think of a better way of doing it then let's discuss it!  

(5) Make install works.  It turns out it is a matter of writing a short script and then the following line in the makefile:

python buildPythonInterface.py install

Pretty cool huh!  Can you do that with awk? ;-)

**Comment by GiovanniBussi** [comment]
Hi @gtribello 

I had a look at the files and I have a few comments.

python/PythonWrapper.cpp, line 33: is it necessary to re-throw as a std::runtime_error?
Notice that PlumedException inherits from std::exception. I would be surprised if python requires the error to be of type std::runtime_error. Are you sure it is necessary?

sourceme.sh: is PYTHONPATH variable needed? I don't see it used anywhere.

version: do you confirm that this goes to master (and not v2.4)? this also means more time to polish the configuration options (see below).

regtest: I think you changed `regtest/scripts/run`. It looks ok to me. In perspective, perhaps we should use an env var instead of the name python, or perhaps configure is at `./configure` time. This would be required e.g. for installing plumed with MacPorts, where we could have variants with different python versions and the name of `python` (and `cython`) should contain the version number. Given this, consider that regtests can also be executed out-of-source (if you install plumed, then download somewhere **only** the `regtest/` directory, they should work). So, the way to access from regtest to configuration options is through the `plumed config` command (see also below)

build: I see you are not building anymore a "plumed module". That's ok I think, if it works. It looks simpler.

autoconf: looks ok, except perhaps we should allow user to customize the name of the python executable. Probably the best way is to use an env var at configure time, store the name of python and cython in Makefile.conf, and retrieve them later while running regtests. Don't worry about this now, I can do it later when everything work. But I want to do it to better integrate with MacPorts (see above).

CPPFLAGS: what you did is consistent with the rest of the code, so I think it's correct! Notice that an user will be able to type `plumed config has python` to know if it is available.

install: I think you did not change `src/lib/Makefile` (not yet). I think you should change it in order to install the python wrapper, right? In case so, I will double check that. It is important that we document all the things that an user should do to uninstall plumed (in case files are copied also out of the plumed main directory). Do you know where `python buildPythonInterface.py install` is adding files?

awk: ops... no, I think with awk there is no way to spend days to write a wrapper to plumed. So maybe awk is better than python? :-P 

Giovanni

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

Thanks for the comments here are my replies:

On the extra throwing on the runtime error I am not sure if it is necessary.  I think if I throw std::runtime_error python can automatically catch the c++ error and translate it into a python runtime error.  That is to say I don't have to tell python to throw a runtime error I can just tell it to look for a c++ exception and it will work out what to do.  Having said that I do tell python to throw a runtime error so maybe it is unnecessary.  Add to this the fact that none of this works on my laptop anyway.  Anyway, what would you put in the catch section to replace the runtime_error? 

I think that the PYTHONPATH in sourceme is necessary if someone tries to run python+plumed and doesn't do make install.  In other words, someone who installs plumed using 

./configure
make
source sourceme.sh

Can use python+plumed because the PYTHONPATH is updated in sourceme.sh.  I think this should stay there as lots of people use the sourceme.sh thing and do not do make install.

This is absolutely for master and definitely not for version 2.4.

Thanks for looking after the stuff with autoconf.  There are possibly better ways to check for python as I think there are custom macros that people have written.  I couldn't get them to work though.  If you could fix the configuration options here that would be great as I think you have a better idea of what you want.

On install I changed plumed2/Makefile as follows:

install:
        $(MAKE) -C src install
        $(MAKE) -C python install

This does the install on my computer.  I don't think I need to change src/lib/Makefile if I do this?

**Comment by GiovanniBussi** [comment]
@gtribello 

About exception. You could re-throw a std::exception, but I think it is even better to remove the whole try/catch statements (i.e.: perhaps no wrapper is needed!) if it works. In case of an error, plumed will throw a PlumedException that inherits from std::exception. So it should work correctly at least if python can deal with any exception class that inherits from std::exception. In case this does not work, in the wrapper I would throw a std::exception and not a std::runtime_exception, since the latter according to C++ manual looks something more specific.

PYTHONPATH. I see, so PYTHONPATH is a standard python variable? Then I agree with what you did. You should also change the `src/lib/modulefile.in` so that PYTHONPATH is set also there. This is needed if someone installs multiple versions of plumed using `module load` commands.

As a related note: your `make install` (from `python`) does not take into account DESTDIR and PREFIX. I think this could be problematic for people using multiple plumed versions. Is there a way to avoid this?

Install: Mm, maybe it is not a good idea, but the way it works now is that `make install` from any directory calls in `make install` from `src/lib`. I think in some sense it is better since it keeps everything in a single place. So it might be better to call `make -C ../../python install` from `src/lib/Makefile`.

Now the point is that `make install` (in `src/lib`) writes a lot of information in the output about which files have been installed where and which files should be removed. It would be nice if you could add there also the information about the additional files created by `make install` (in `python`).

The bottomline is that perhaps the only way to make something consistent with `--prefix` installations is to not use the python install commands and tell the user to set PYTHONPATH at runtime also when plumed is installed...

Does this whole message make sense??

Giovanni

**Comment by gtribello** [comment]
@GiovanniBussi 

I made a few changes to the branch in response to your message.  In particular:

I moved the install command to src/lib/Makefile and removed it from plumed2/Makefile.  I added the pythonpath to src/lib/modulefile.in.  I haven't done anything on the exceptions yet as I need another computer to check that stuff out but I will look into it presently.

The one remaining problem as I see it is this business of the DESTDIR and PREFIX.  I don't really want to manually copy the python module stuff to the destination directory manually.  I would much rather use the install command from distutils as it makes life simpler.  I thus think DESTDIR and PREFIX should be ignored as you want to install python+plumed in the appropriate directories for the python version you specified at compile time.  My thinking is that you would deal with users having different versions by having them change the name of the executable. Does that make sense?

So lets say we compiled two version of plumed aplumed and bplumed.  I thought if I changed got the name from makefile.conf and set the python module name to the name of plumed I would be able to deal with.  The user would then be able to do:

import aplumed 

or 

import bplumed 

and switch between the different versions.  That doesn't seem to work though.  I think it struggles because the library is only given the name you specify in configure until you do the make install.

I didn't understand the recent email at all sorry.

**Comment by GiovanniBussi** [comment]
Currently there are [two different ways](https://plumed.github.io/doc-v2.3/user-doc/html/_installation.html#InstallingPlumed) to install multiple versions.

#### Same prefix different progname

In this case, you pick a different progname (e.g. `aplumed` as `bplumed` as you said). I am not sure this is really useful, but I think we should support it. In this case, I guess you would be able to use the install command as you did.

Just to confirm: if you do this, can you decide to import either `aplumed` or `bplumed`, right?

#### Same progname different prefix (with tcl modules)

This is the way I would expect most people use in clusters, and is based on the modulefile. Here, plumed is always called `plumed` (though the name could change in principle) but is installed using different `--prefix` options in `./configure`. On my machine, e.g., prefix is `$HOME/opt/install/plumed/$version` so that the executable of plumed 2.2.2 is in `/Users/bussi/opt/install/plumed/2.2.2/bin/plumed`. To select one version or another you use `module load` commands in order to add `/Users/bussi/opt/install/plumed/2.2.2/bin` to the path.

The point is the following. If you do
````
./configure --prefix $HOME/test-install
make
make install
````
then the python module should be installed somewhere inside `$HOME/test-install`.

Notice that this is a bit simplified. There are further dirs that can be specified (I added all those indicated in the autoconf documentation that could make sense with plumed, e.g. doc dir etc).

#### DESTDIR
This is a different story, only needed to temporarily stage the whole package somewhere (used by package managers such as MacPorts or Conda). This should also be supported in order for plumed to be installable. It should be prepended to all paths during install.

#### How to install
I think you have two options:
- In case python allows it, you can learn how to instruct the python install command to place files in the standard place. You should also define the correct place based on the autoconf directories (perhaps `$(DESTDIR)$(libdir)/pythonX.Y/site-packages/$(program_name)`?). I guess that if you use the default prefix the module will be immediately visible. Finally, you should write in the doc (and in the output of `src/lib/Makefile`) which files/directories should be removed in order to uninstall plumed. 
- Instead of using `python install`, you copy the relevant files to `$(DESTDIR)$(libdir)/$(program_name)/python`. In this case, you should write in the documentation (and in the output of `src/lib/Makefile`) which dir should be prepended to `PYTHONPATH`.

In both cases, for people using modules you should set the`PYTHONPATH` variable in the modulefile.

I would say the second option is simpler (treats the installed plumed in the same way as the non-installed one) to implement. The advantage of the first option is that users installing plumed in the default place (or with a package manager such as MacPorts or Conda) will probably have the python module available without the need to set `PYTHONPATH`.

Hopefully this is clearer now...

Giovanni

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

I think I understand now and I have tried fixing it so that it works in the way you suggested.  From my understanding of these messages there are two outstanding issues:

(1)  Exceptions (again)

I have checked this now and the PLMD::Exception is caught by python (but not on my laptop) if you remove the PythonWrapper.cpp file.  I have thus removed PythonWrapper and am calling plumed_cmd directly.  In short, I think this is sorted.

(2)  Make install business

I have reworked all this and am no longer using distutils to do the install.  Instead, I copy everything for python to the directory:

$(DESTDIR)$(libdir)/$(program_name)/

as you suggested.  I think I am perhaps overdoing this as I think all I really need to copy is the library that is generated at the end of the cython build process.  I copy everything though because I have found that this library has different names on different machines.  I also don't know how to control the name of the library.  

I also added in some code to the makefile to write out suitable instructions in the makefile about how to use PLUMED + python if it is installed.

I know this is not quite perfect as yet but we would really like to merge all this stuff into master ASAP as I only here in Lausanne for one more week and there are a few other branches we would like to get this onto so that we can try some other things whilst I am here.  As such are you happy for me to merge this stuff once I fix the conflicts.  

Thanks

**Comment by GiovanniBussi** [comment]
@gtribello great! For me you can merge.

I don't know if you tried Travis-CI already. If I recall correctly, Travis-CI does install plumed and executes the tests with the installed plumed (it does not for instance `source sourceme.sh`). So probably in order for the test to work you need to set PYTHONPATH in .travis.yml

Giovanni

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

I am trying to get the travis.ci to work for this but I have a problem with cppcheck and I am not sure what to do.  In essence, it complains because there is nothing within the c++ code that is within a 
-D__PLUMED_HAS_PYTHON.   Do you think I should just stick something like:

#ifdef __PLUMED_HAS_PYTHON
#endif 

Somewhere inside the code to silence the error or is there something more intelligent that can be done?

Thanks
Gareth

**Comment by GiovanniBussi** [comment]
@gtribello the check is there to avoid spelling errors. I think the elegant way would be to have some way to allow exceptions in the plumedcheck script (similar to cppcheck script), but clearly adding the macro somewhere in the code is way easier... Just put a comment close to it so that we remember why it is there

**Comment by carlocamilloni** [comment]
Hi @gtribello, does this work with both python2 and python3? (simple curiosity)

**Comment by gtribello** [comment]
@carlocamilloni I think it should work with both yes.

**Comment by gtribello** [comment]
Hello @GiovanniBussi 

I am having a bit of difficulty with getting the python interface to pass the tests on travis-ci and I wondered if you might be able to make some suggestions.  The issue is that you cannot seem to import the plumed module in python.  I am not sure if this is because I made a mistake with the setup of the environment variables or if it is a problem in the compilation on travis.  When I compile on Travis I get the following error message at the point after it builds the python interface:

ccache -pthread -shared -L/opt/python/2.7.13/lib -Wl,-rpath=/opt/python/2.7.13/lib build/temp.linux-x86_64-2.7/plumed.o -L../src/lib/ -L/opt/python/2.7.13/lib -lplumed -lpython2.7 -o /home/travis/build/plumed/plumed2/python/plumed.so
ccache: invalid option -- 'p'
Usage:
    ccache [options]
    ccache compiler [compiler options]
    compiler [compiler options]          (via symbolic link)
Options:
    -c, --cleanup         delete old files and recalculate size counters
                          (normally not needed as this is done automatically)
    -C, --clear           clear the cache completely
    -F, --max-files=N     set maximum number of files in cache to N (use 0 for
                          no limit)
    -M, --max-size=SIZE   set maximum size of cache to SIZE (use 0 for no
                          limit; available suffixes: G, M and K; default
                          suffix: G)
    -s, --show-stats      show statistics summary
    -z, --zero-stats      zero statistics counters
    -h, --help            print this help text
    -V, --version         print version and copyright information
See also <http://ccache.samba.org>.
error: command 'ccache' failed with exit status 1

I don't seem to do anything with ccache when I compile on my laptop though.  Is this relevant do you think?

Thanks
Gareth

**Comment by GiovanniBussi** [comment]
@gtribello if I understand correctly `python/Makefile` the only command that you execute is `buildPythonInterface.py build_ext -i`, which in turn should learn the name of the C++ compiler and linker from somewhere. Do you know how?

It looks like the compiler is `/usr/bin/mpicc` and the name of the linker is `ccache`, though it should be something like `ccache /usr/bin/mpic++` (see line 3608 of [your logfile](https://travis-ci.org/plumed/plumed2/jobs/272372068))

**Comment by gtribello** [comment]
@GiovanniBussi Thanks for that.  I think this is the problem.  Also I think I have worked out how to set the environment variables.  The only thing is that I haven't worked out how to set the right variable for the linking stage.  I set CC and CXX but I still get the same error.  Do you have any idea of what environment variable I need to set?  I guess the question is what would it be using to link.  Thanks.

**Comment by GiovanniBussi** [comment]
@gtribello no idea.... However, it looks python build is deciding in someway to use ccache as the linker. You can try the following:
1. Add a `env` command in travis-ci just before `make python` so as to dump the environment
2. Check the log after the run to see which env vars contain the word `ccache`

Notice that plumed uses env vars such as `PLUMED_CXX`, but travis-ci might set other variables as well.

Finally, notice that I now set the option `autocancel` on pull-request builds. This means that if you push again you will kill the previous build (only for pull requests). This is because the queue is getting very crowded and some jobs take one day to complete.

**Comment by carlocamilloni** [comment]
@giovannibussi I would set the autocancel also for all development branches (not master, v.X.Y)
Do you think it is possible?

**Comment by GiovanniBussi** [comment]
@carlocamilloni looks like [it's not possible](https://travis-ci.org/plumed/plumed2/settings).

Notice that you can overcome the limit of 5 jobs pushing to a repository owned by a different user (e.g. your fork). Maybe we should do development on separate repositories?

If it's just for using travis-ci, I use a [test repo](https://github.com/GiovanniBussi/test-travis-ci) where I force push things that I want to test on travis-ci

File: python/buildPythonInterface.py [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+  name='plumed',
+  version=plumedversion,
+  description='Python interface to PLUMED',
+  author='Gareth A. Tribello',
+  author_email='plumed-users@googlegroups.com',
+  url='http://www.plumed.org',
+  ext_modules = cythonize([
+                  Extension( name="plumed", 
+                             sources=["plumed.pyx"],
+                             library_dirs=["../src/lib/"],
```
GiovanniBussi: @gtribello I think the problem on osx is that you should set library_dirs to `${libdir}` variable (set in `Makefile.con`). On osx typically you have to correctly hardcode the path to the library in order to be able to find it later.

A possible solution would be to create two modules, one for usage "without install" and another for usage "after install", and then install the latter. Or maybe to say that python module can only be used if you full install plumed.

Then, there will be another subtle issue: it is now possible to change the prefix when you do `make install` (e.g. `./configure --prefix=/a && make && make prefix=/b install`). So in principle you should detect if libdir has changed and recompile the library when you install.

Perhaps it is simpler if you just rebuild python.so when using `make install` directly in place (with correct include_dirs and library_dirs)

gtribello:
Hi @GiovanniBussi 

Thanks for the tips.  I had another go at this thing.  I have my fingers crossed for it to work.  It at least works on my mac but it was working before on my mac so who knows if that is any indication.

What I have tried to do is this:

(1)  The python interface is rebuilt when the user invokes make install so as to deal with the problem of the path not being specified correctly in the dynamic library.

(2)  The script that builds the python interface now checks if there is a version of plumed in the users path when it builds.  If it finds one it uses the info from that build on the root and include directories (which it gets by calling plumed info) when cythonizing.  If it doesn't find a plumed then ../src/lib/plumed is used.

A problem I can see with this solution is if a user does:

make
source sourceme.sh
make install

Then the plumed in src/lib might be used when really the installed version in the installation directory should be used.  Now that I have written that down I am not sure it is true.  I am so confused.  Would you mind taking another look and checking what I have done.  I would be very grateful as this is making my brain hurt.

GiovanniBussi:
@gtribello yes this is painful... Anyway, the way you did I can see these problems which are similar to those I found with the normal `src/lib/Makefile` install procedure.

About looking in the path, I am not sure it is a good idea for the reason you explained. It might be ok, but then the instructions should be totally different. `make install` should **not** install python. Users should first install plumed, then go to `plumed2/python` dir and type `make install`.

Notice that there is another possible issue. At `make install` you rebuild the `plumed.so` in place. This means that if you do `make; make install` then set `PYTHONPATH` to `plumed2/python` then python is going to use the installed version instead of the development one.

The solution I used for normal install is that a separate version of the libraries and executables are built in the `src/lib/install` subdir when you do `make`. Those are the ones that hardcode the paths to the install place, and those that are copied later when you do `make install`. You could do something similar.

There is yet another possibility, since you are using only the `cmd` interface. Instead of linking `-lplumed`, you can directly link the `src/wrapper/Plumed.o` object, which is self-contained and has no problem with libraries. At runtime, by choosing PLUMED_KERNEL env var, you can pick the plumed version. In order for this to work without setting PLUMED_KERNEL it is possible to hardcode the location of the installed library in `src/wrapper/Plumed.o` compiling with `-D__PLUMED_DEFAULT_KERNEL=/path/to/libplumedKernel.so`. This is the solution I used to install plumed on MacPorts only, but we could make it the default. The consequence would be that people patching an MD code with `--runtime` option will be able to set PLUMED_KERNEL env var and, if this is undefined, will use a fallback solution with the installed plumed.

Would this make things easier for the python interface as well?

GiovanniBussi:
@gtribello I think I know why it is failing with MacPorts. The problem is the test:
````
(python)
  if command -v python > /dev/null 2>&1; then
     python -c "import plumed"
````
You are only checking if python is available, but not if it was available when plumed was compiled. You should do something like ````plumed config has python````.

In MacPorts typically you do not link libraries or use binaries that are outside of the MacPorts tree (in order to avoid irreproducible builds on different machines), so you should install MacPorts python/cython first if you want your test to work.

I will fix the MacPorts Portfile later (plus I will have to make the name of the python tool a variable, since by default it is called pythonX.Y). So don't worry about this, I think it is ok if you just change the test in the run script so that temporarily it does not run on MacPorts.

However, the problems that you mentioned before with paths will still remain, so maybe better to address them.

gtribello:
Hello @GiovanniBussi 

I am a bit confused by the last few emails so I just want to double check that we are on the same page.  This is a summary of what I think the salient facts are regarding this python business.

(1) Plumed + python now works for both the linux and osx installs on travis.  Hurrah!  It doesn't work on Macports.  Boo!  However, for Macports I should be able to resolve the problem by replacing my test (python -c "import plumed") with the test (plumed config has python).  Stupid question but where is the macports executable built?  As in where is ./configure run and where is the making done?  It is not on Travis correct?

(2) There is a problem in the way I have implemented make install.  In particular, the fact that I build the python interface in place is problematic.  What I should do instead is copy the python directory to the staging area (src/lib/install) when I do make install and rebuild the interface in the staging area.  In terms of the options you have given I would rather do it this way as it seems a relatively straightforward change to make.

Thanks

GiovanniBussi:
(1) It's built inside MacPorts using the Portfile in `plumed2/macports/science/plumed` (generated with `make` in `plumed2/macports` from template `Portfile.in`). This really simulates what happens when someone types `port install plumed`, except for the fact that it does not download the source from a remote repository but from the path to the present repository. It thus runs under some special macports user and does not see your environment variables. In order for it to see python, one should make `plumed` package depend on `python` on macports, perhaps with a variant to control python version (e.g. `port install plumed +py27`).

(2) This should work and solve the problem

gtribello:
OK @GiovanniBussi that is great.  I have made those changes.  Fingers crossed that this thing will finally pass all the tests on travis.  Thanks for all the help.

GiovanniBussi:
@gtribello I am checking how to customize name of python interpreter (see my attempts at [this branch](https://github.com/GiovanniBussi/plumed2/tree/python), not done yet). I thought it is better not to check for `cython` executable, since you are not using it, but rather to see if `import cython` works. In addition, I noticed you use a number of other python modules (`numpy` etc). I would check for all of them (on my machine with macports python3.3 for instance has no `numpy` installed right now). These things should be easy to fix. I will try to import all the modules you use in the build script.

I also have some doubt:

- Here:
https://github.com/plumed/plumed2/blob/15b68ba8ed7e6ae96635d32077a5cbe3f9efee18/python/buildPythonInterface.py#L38
you are looking for `plumed` in the executable path. This is causing problem on my machine, for instance since I have plumed2.3 installed and I am trying to build plumed2.4. Is this expected? I think while you build you should always look for `src/lib/plumed`, otherwise you will use a possible previously installed `plumed`.

- Here: https://github.com/plumed/plumed2/blob/15b68ba8ed7e6ae96635d32077a5cbe3f9efee18/regtest/python/rt-protein/config#L1
I think you are missing a `type=python` command.

PR #265, opened 2017-08-17
Close structure for property map/path CV

jpazurikova:
Hi,
I would like to propose a close structure method for acceleration of path/property map CV using (R)MSD. As the most demanding part of this distance computation is the matrix diagonalization (which results in rotation matrix), we reuse rotation matrices between the current structure and reference structures computed in previous steps. So we introduce the close structure, which is basically the simulated system at the certain point of time. At this time point, we save calculated rotation matrices to all reference structures. Then, in the next steps (until the close structure moves beyond the distance threshold from the current structure) we reuse them to approximate the rotation matrices between the current structure and reference structures. This way, we can speed up the calculation. Details about the method are available here http://aip.scitation.org/doi/abs/10.1063/1.4978296

The code is rather limited, it affects only classes colvar/PathMSDBase and tools/RMSD. Default remains unchanged, the method turns on when the distance threshold EPSILON is set as option of PathMSDBase. Moreover, additional logging and debug mode can be turned on with options LOG-CLOSE and DEBUG-CLOSE. In addition to the method implementation, we performed some basic vectorization on the code. This is effective only with Intel compiler, gcc compilation works (it gives warnings about unknown pragmas), but it does not accelerate as much.

- [x] I added a new regtest or modified an existing regtest to validate my changes - directory basic/rt-close-structure
- [x] I verified that all regtests are passed succesfully on Travis - done, see https://travis-ci.org/jpazurikova/plumed2

Of course, I would gladly make any changes regarding the code if you consider them necessary. 
Looking forward to hearing from you,
Jana Pazurikova


GiovanniBussi:
Dear @jpazurikova thanks for your contribution.

I had a look at the code and I think there are a few things that should be clarified in order to merge it.

1. You are asking to merge on `master`. If you want this to end up in v2.4 (due end of this year) please ask to merge it on `v2.4`.
2. In order to facilitate the merge, please rebase your commit on the proper branch (see previous point) and split it as a series of meaningful commits, each addressing a single point. A possible series could be:
    1. Changes to `tools/RMSD.cpp` and `tool/RMSD.h`  necessary to implement your stuff.
    2. Changes to `colvar/PathMSDBase.cpp` and `colvar/PathMSDBase.h` using the modified RMSD class.
    3. SIMD pragmas
    4. Regtests
3. There are many changes that appear only reformatting. They should be removed.
4. There are changes to `tools/Tensor.h` that I do not understand
5. Why did you add constructor/destructor to RMSD class?
6. About SIMD instructions, I have some question:
    1. Are they portable to other loops in the code? In case I would be happy to do it myself.
    2. If I understand correctly these are for `icpc` only. Is there a way to have them compatible with `gcc` as well? What about using openMP 4.0 statements?
    3. In my experience, `icpc` compiler does a good job in finding automatically the loops that can be vectorized. Did you check if these ones are found also automatically?
    4. Which is the performance impact of these pragmas?
6. Perhaps I will be able to understand it from the code (once it is reformatted). How do you compute derivatives? Normally when align==displace we use Hellman-Feynman trick for forces. It this still valid? (I mean: forces dues to reference frame rotation are zero).
7. Finally, which is the performance impact of the overall method? Perhaps I can find it in the original paper. It is a bit surprising to me that the bottleneck is in the diagonalization of the matrix.

Thanks!

Giovanni

jpazurikova:
Dear Giovanni,
thank you for your prompt response. I answer your comments below.

1., 2., 3. I will rebase the commits according to your requirements and make a pull request to v2.4 during this or next week. However, some formatting changes were due to astyle errors reported by Travis CI, I will try to keep these to the minimal level.

4. We have added an "inline" keyword to some functions. These can improve the performance in the case of often called functions, such as matrix multiplication, as the compiler will inline the function's code everywhere where the function is called.

5. Oh, my bad. I observed a bug as reference_center and positions_center were not zeroed out in the method clear(), which I fixed in the commit 5b29389 "Add copy constructor and clear centers in RMSD". I added a copy constructor during debugging and forgot to get rid of it later. 

6.
- i. Yes, these pragmas are portable to other loops. Further info here, https://software.intel.com/sites/default/files/m/4/8/8/2/a/31848-CompilerAutovectorizationGuide.pdf , especially parts 3 and 4. generally, you want to vectorize the loops over a larger number of iterations (e.g. over number of atoms). Profiling of the code with tools like Intel VTune can reveal the loops that would benefit the most.

- ii. Unfortunately, pragmas are compiler-dependent in general. However, some things can be done. #pragma ivdep that tells the compiler that the data structures in the loop are not dependent on each other and therefore the loop can be vectorized. Its version #pragma GCC ivdep should be available in gcc from version 4.9. So you should be able to switch between them with C macros, depending on the given compiler. Generally, gcc does not have such large support for vectorization as Intel compilers. OpenMP 4.0 is a good idea and #pragma omp simd should be more portable than intel pragmas we have used, at least with the latest versions of other compilers. I would be happy to rewrite and test it.
 

- iii.  We have analyzed what loops icpc automatically vectorizes via logs from compilation with flag  -qopt‐report    -­qopt‐report‐phase=4. We have added pragmas only to the loops that were not automatically vectorized, usually due to the assumption of dependencies among data structures (this compiler's assumption is negated with #pragma ivdep).

- iv. The performance effect of these pragmas using the original method for distance computation is negligible, as it spends the vast majority of time in BLAS function DSYEVR computing the diagonalization of the matrices. However, when the number of calls to DSYEVR is reduced using the close structure method, the loops calculating the derivatives become prominent. We reduced their cost with simple vectorization. Without it, the benefit of the close structure method does not reach its potential.

7. We have largely reused the implementation of derivatives calculation available in Plumed. The modified equations are in the paper, basically it bottles down to different rotation matrices used to calculate ddist_drr01:
ddist_drr01+=ddist_drxy[i][j]* drotationPosCloseDrr01[i*3+j]; in our method
instead of
ddist_drr01+=ddist_drotation[i][j]*drotation_drr01[i][j] in the original method.
ddist_drr01 is then used to calculate the derivates, we have not changed the method RMSDCoreData::getDDistanceDPositions() that calculates them.

8. The paper from Journal of Chemical Physics does not focus on performance, we had just a prototype implementation then. But the information from profiling software (Intel VTune) showing that the diagonalization takes really a lot of time is in the supporting information, see http://aip.scitation.org/doi/suppl/10.1063/1.4978296. We are currently submitting the paper regarding the performance of the method. In our experiments, we achieved the speed-up of 14 for the simulation of cyclooctane in the vacuum with 521 reference structures and no neighbourlist. In the simulations with Trp-cage, we achieved the speed-up of 1.5 with 50-item neighbourlist updated every 50 steps. I can send you the draft of the paper via e-mail if you are interested in details.

Regards,
Jana

GiovanniBussi:
Hi @jpazurikova 

thanks for the clarifications.

Format: what you report is a bit strange. In principle the code 'as is' should be already compliant with the astyle check on Travis-CI. Notice that to apply astyle you should use `make astyle` from the plumed root directory, in order to use the same `astyle` version and options that are used on Travis-CI.

Inline: I though all non-fully-specialized templates are automatically inlined. Did you see any improvement with those specific `inline` commands? If so, I would like to add the `inline` statement to all the functions in Tensor and Vector.

About OpenMP 4.0: it is not necessary to do it now, in case you don't have time for me it is ok if you leave the code as is (intel pragmas). I will then open a new issue so that we don't forget to check that.

Performances: when you say speedup is 40 (or 1.5) do you mean the final gromacs+plumed speed?

Derivatives: currently the code has two different ways to do forces, depending on the value of (align==displace). Does your code work in both cases? 

Thanks!

Giovanni

jpazurikova:
Hi Giovanni,

Format: Will do.
Inline: We have not tested the effect of inlining itself. I will run the tests and report back.
OpenMP 4.0: Sounds good to me.
Performance: Yes, we sped up the whole simulation (Gromacs+Plumed) 14 or 1.5 times. 
Derivatives: Yes, it does.

Thanks,
Jana

GiovanniBussi:
Dear @jpazurikova 

this is just to tell you that we plan to release v2.4b at the end of September. In order to include your changes in v2.4 we should thus be able to merge them before that date. There is still time, but I wanted to alert you in advance.

Thanks!

Giovanni

jpazurikova:
Thank you!
I apologize for my tardiness, I am in the final stages of submitting my thesis and last-minute changes took longer than expected.
I will request for pull in branch v2.4 next week so that we have time to deal with any issues.

jpazurikova:
I have requested a new pull onto branch v2.4 in #285 , I will continue the discussion there.

PR #269, opened 2017-08-28
V2.4 c++11

GiovanniBussi:
Dear @plumed/all 

I open a pull request to collect comments on this branch that I plan to merge. It is based on v2.4 but I would rather merge it on master directly (in order not to introduce unnecessary changes in v2.4 close to release).

I think we should try to make plumed exception safe at some point, that is make sure that when an error is raised data are deallocated correctly. An easy way to make at least memory managed correctly is to stop using (or minimize the use of) new/delete statements.

I tried to remove almost all of them (there are a few still) in `tools` and `core` in this branch. In particular, I replaced most pointers with c++11 `std::unique_ptr`, that get deallocated automatically when the pointer goes out of scope. In `tools/Matrix.h` instead I just used many `std::vector`. I also added some documentation on smart pointers (look for `cxx11features` in [this file](https://github.com/plumed/plumed2/blob/v2.4-c%2B%2B11/developer-doc/plmdIntro.md))

Any comment on this? In case I don't receive feedbacks I will merge it to master next week. Then I think we could slowly try to update also the rest of the code. It is not difficult but I find it risky to change myself parts of the code that I do not know well, so I might ask some help.

Giovanni


gtribello:
On the subject of pointers (but my question is perhaps tangential to this change).  Why inside ActionWithValue do we have a std::vector<Value*>?  Can this not just be a std::vector<Value>?  The point is that when you reuse the data in the values inside ActionWithArguments you could pass references rather than pointers.   As these are in a std::vector you have no problems with the fact that the references are not allocated with the ActionWithArgument is constructed.  You would only have a problem if you were trying to build a class that had Value& object within it.  We wouldn't need that though as the arguments object in ActionWithArguments would be a std::vector<Value&>.

GiovanniBussi:
@gtribello I am not sure I understand, but notice that you cannot declare a `std::vector<Value&>`.

I think originally this was since in order to be storable in STL containers objects should be copyable. Anyway, with C++11 you can store also non copyable objects. So perhaps we could replace it with a `std::vector<Value>` now, even without changing the interface of the ActionWithValue class.

carlocamilloni:
I think it is a very good idea. I am happy to have this merged in master.

zwang123:
Dear @GiovanniBussi 

I believe that using the smart pointer introduced in c++11 provides much more memory safety. However, mixing the usage of smart pointer (including std::unique_ptr) with raw pointer is dangerous. For example, after the unique_ptr is destructed, the raw pointer obtained from get() function (and possibly the following dynamic_cast) is no longer valid. Besides, it is sometimes difficult to check if we need to delete the raw pointer without knowing the status of the original unique_ptr.

I suggest using std::shared_ptr and std::dynamic_pointer_cast instead, despite some (maybe negligible) runtime overhead since the deleter is bound at runtime for a std::shared_ptr. If you do not want to mess up the internal pointer counter, std::weak_ptr is another choice. What is your thought on this?

Thank you,
Zhi

GiovanniBussi:
Hi,

the transition to smart pointers implemented so far is:

1. Owning pointers become unique_ptr.
2. Non-owning pointers remain raw pointers.

Notice that I did not change the whole code yet, there are parts that are a bit more difficult to modify.

I think shared_ptr make sense if you really want the pointed object to survive as long as there is at least one pointer still active. Do you think there are places where this would be desirable? To me it looks that this would add unnecessary complexity. E.g., do we want PlumedMain object to stay alive just because someone forgot to close a File object?

About std::weak_ptr, I thought that only makes sense in conjunction with shared_ptr, is this correct?

Given this, I am not an expert at all, so I am totally open to inputs.

Thanks!

Giovanni

GiovanniBussi:
About this point: "Besides, it is sometimes difficult to check if we need to delete the raw pointer without knowing the status of the original unique_ptr". I would say in the long run I want to remove all delete statements. In other words, all the raw pointers will be not-owning. So I think this should not be a problem, correct?

Thanks again

Giovanni

zwang123:
Yes, I agree that your design is reasonable and we should not unintendedly increase the lifetime of the pointed-to object by having one more shared_ptr.

However, the developer has to make sure the owning pointer is designed to be the longest-living one, and the user generally has to know the validity of the non-owning pointers before using them. I do not know if this might be an issue. If yes, we might need something like "weak_ptr" for unique_ptr. There are some discussions in stackoverflow:
https://stackoverflow.com/questions/17536731/shared-ptr-is-to-weak-ptr-as-unique-ptr-is-to-what

A possible solution is to encapsulate the shared_ptr in a class and delete the copy constructor and copy-assignment operator. But I do not know how much runtime overhead does weak_ptr have, as compared to a raw pointer.

Anyway, it is best if there is an unambiguous owning pointer.

Thank you,
Zhi

PR #273, opened 2017-08-30
Added cross refs in EDS and MAXENT

GiovanniBussi:
As discussed with @hockyg, it makes sense to cross link these two implementations of similar methods.

@hockyg or @whitead: can you confirm you are ok with this change of the EDS documentation?

I also cc @a-cesari who originally wrote the MAXENT keyword.

Thanks!


hockyg:
Looks good to me

PR #278, opened 2017-09-08
unique_ptr: analysis

GiovanniBussi:
Hi @gtribello 

I am slowly replacing delete statements with std::unique_ptr. In this commit I did it for the `analysis` module, which however I don't feel I understand well enough. Could you have a look and tell me if this can be merged?

Thanks!

Giovanni

PS Tests on travis should be ok (though osx is frozen now), but I will anyway wait for all the tests to complete.


gtribello:
Hello @GiovanniBussi 

I had a look at this and I think it looks fine so please go ahead and merge it.

Gareth

PR #279, opened 2017-09-09
Add more regtest for DRR module

HanatoK:
##### Description

Add more regtest for DRR module

##### Target release

##### Type of contribution

##### Copyright

##### Tests

Travis-CI:
https://travis-ci.org/HanatoK/plumed2


HanatoK:
@GiovanniBussi 
The travis-ci test in [https://travis-ci.org/HanatoK/plumed2/builds/273522692](url) failed on Xcode.
The rt-merge regtest seems broken with libc++, but I can't reproduce this in linux.
I had recompiled the whole boost with libc++ and clang, and then compiled plumed with libc++ and clang, but every test in regtest/drr is OK.
There is a Mac computer in our lab but I don't know how to setup the whole compilation environment.
Do you know something about the regtest failure?
Thanks!

GiovanniBussi:
Hi

Looking at the tests it looks like the problem is either with `libc++` (as you suggest) or with the `boost_serialization` library installed by MacPorts. Notice that the  [build with homebrew](https://travis-ci.org/HanatoK/plumed2/jobs/273743322) does not fail, though this links `libstdc++`. 

A few things that you can try:

- Add `--disable-static-patch` to `./configure` and check if travis jobs 7 and 8 (osx with homebrew) are still successfull. This should avoid linking `libstdc++` and (I guess) link the default `libc++`. If these fails, `libc++` on osx should be blamed I think.
- On the mac you have available, you can try to install plumed with macports. Should not be difficult. You can also read [this](https://github.com/plumed/ports), which however installs a specific plumed version. In order to install your own plumed version you should do the following:
  - Make sure you have macports installed (possibly it is already installed).
  - Go to `plumed2/macports` and type `make ; portindex`.
  - Edit file `/opt/local/etc/macports/sources.conf` and add somewhere at the beginning `file:///path/to/plumed2/macports`
  - Check if plumed is avalable with `port info plumed`. The version should be the hash of your plumed repository. In case the version is something like 2.3.2 then you are using default `plumed` from macports and you should check the previous steps.
  - Type `sudo port install plumed +allmodules` to install plumed with optional modules (including `drr`) enabled.
  - Try the regtests with this plumed version.

In case you cannot fix the problem, probably you have to leave the failing test out of the regtest suite. Rebase on `v2.4` (I just added a new small thing) and add the following lines to the regtest `config` file:
````
function plumed_custom_skip(){
  if [ "$(uname)" == "Darwin" ] ; then
    return 0
  fi
}
````
Plus you should probably write in the documentation of your module that some features might not be available when compiling on osx. In this way, the test will still run on Linux and contribute to the coverage.

HanatoK:
@GiovanniBussi 
Is it right to modify the PLUMED_CXXFLAGS under os: osx to `PLUMED_CXXFLAGS="-O3 --disable-static-patch"`?
The mac computer is mainly used to show ppt or pdf slides. There are no Xcode or macports on it. Currently I am trying to install macports stuff but I am totally unfamiliar with mac. It may be a hard time for me to operate on it.

GiovanniBussi:
No, you should change e.g. these lines:
````
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && test -z "$PLUMED_MACPORTS" ; then
      brew update > /dev/null ;
      brew install ccache ;
    fi
````
to
````
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]] && test -z "$PLUMED_MACPORTS" ; then
      CONFIG_FLAGS="$CONFIG_FLAGS --disable-static-patch"
      brew update > /dev/null ;
      brew install ccache ;
    fi
```` 

Anyway, this is just to know if osx+homebrew+libc++ will fail as well. It will not solve your problem.

HanatoK:
I installed plumed using macports successfully. However, all drr regtest are passed. The mac version is 10.10 and Xcode is 7.2.1.
Should I upgrade the system and test again?

HanatoK:
@GiovanniBussi 
I installed plumed in macOS Sierra and ran regtest. All drr regtest are passed.

I don't know how to debug it on travis-ci.
I may just disable the regtest on travis-ci's macports. Do you have any better suggestions?

HanatoK:
@GiovanniBussi 
I have added plumed_custom_skip(), but on Linux it also shows NOT_APPLIABLE (plumed_custom_skip).

GiovanniBussi:
@HanatoK sorry I did a mistake copying the plumed_custom_skip function.

Tests look ok now. Just let me know if I should merge. Thanks!

Giovanni

HanatoK:
@GiovanniBussi 
It's OK to merge.

GiovanniBussi:
@HanatoK this is just to let you know that as of 75645e85104ea30ffa7ad4eb2c673e02e9357551 we are skipping another regtest on OSX.

Due to another issue (#306) we realized that plumed executable returning non zero value was not detected as an error. After fixing this problem (dae47d04d01e6f2d20f992842af416f3fc7ad688), one your regtests crashed on travis-ci (see [here](https://travis-ci.org/GiovanniBussi/test-travis-ci/jobs/288790295)).

My guess is that there are some incompatibilities between homebrew and macports that create troubles on travis-ci using boost. Anyway, no need to solve them now.

PR #281, opened 2017-09-13
unique_ptr: PBMETAD

GiovanniBussi:
@carlocamilloni @maxbonomi 

I replaced pointers with unique_ptr's in PBMetaD. Could you please check if what I did is ok? In case so I will merge it to master. Thanks!

Giovanni

### DIFF
--- a/src/bias/PBMetaD.cpp
+++ b/src/bias/PBMetaD.cpp
@@ -35,6 +35,7 @@
 #include <iostream>
 #include <limits>
 #include <ctime>
+#include <memory>
 
 #define DP2CUTOFF 6.25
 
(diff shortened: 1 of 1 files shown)


carlocamilloni:
It looks fine, but I am not very confident about my understanding of unique_ptr yet.

C

carlocamilloni:
@giovannibussi I don't understand why in the destructor we don't need to close the files anymore

GiovanniBussi:
@carlocamilloni I checked the code for [FileBase::~FileBase()](https://github.com/plumed/plumed2/blob/566d287a0ca50f9ed711f0a35a353ffb2c675619/src/tools/FileBase.cpp#L121) and [FileBase::close()](https://github.com/plumed/plumed2/blob/566d287a0ca50f9ed711f0a35a353ffb2c675619/src/tools/FileBase.cpp#L95) and it looks like the destructor is also closing the file. The only difference is that close() also asserts that file has been explicitly opened (not cloned from another file) and resets the local pointers to NULL (in order for the file object to be reusable).

So, in short, closing the file should not be necessary if you are going to destroy the file object.

carlocamilloni:
ok then is good to me!

PR #283, opened 2017-09-14
More versatile Dimer input

marckn:
##### Description

The Dimer CV now allows to select each dimer singularly so that it can be used 
for systems where only a part is being dimerized. 
Input interface has been renewed and documentation changed accordingly.

##### Copyright

- [ X] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

- [X ] the module I added or modified contains a `COPYRIGHT` file with the correct license information. I also used the command `cd src && ./headers.sh mymodulename` in order to make sure the headers of the module are correct.

##### Tests

- [ X] I added a new regtest or modified an existing regtest to validate my changes.
- [X ] I verified that all regtests are passed successfully on Travis-CI.


GiovanniBussi:
@marckn please check the failure on travis (looks like astyle problem).

Also notice that with this request your changes will end up in v2.5 (2018). In case you want them to be included in v2.4 please rebase them on branch v2.4.

marckn:
Yeah it's an astyle problem but the message is a bit cryptic. Does that mean that the code formatting is not compliant with the standard adopted by plumed?
As for the version I didn't notice: if I include it in 2.4 will it be automatically available in 2.5 or would you like me to make a pull request on both branches?

GiovanniBussi:
Astyle: You should apply the `astyle` formatting to your code. Should be as easy as `make astyle ; git commit` from plumed2 dir.

Branch: v2.4 is frequently merged to master, so whatever we put in branch v2.4 will also be present in v2.5 (unless we explicitly remove it there).

marckn:
wow cool ! I've never heard of that ! I'll do that right away and also rebase on v2.4. Now it should pass the astyle test...

marckn:
ok, compilation is in queue on travis for both master and 2.4 branches. How do we proceed? Shall I open another pull request on the 2.4 branch ?

GiovanniBussi:
I think you should open another pull request (unless it is possible for you to edit the base branch).

marckn:
o.k. I'll open another pull request as soon as I see that travis compiles successfully

PR #289, opened 2017-09-18
Added support for PYTHON_BIN configure variable.

GiovanniBussi:
@gtribello this should work to set python executable even when its name is different from `python`.

Notice that regtests are still running with `python` (I will fix it later). Meanwhile you might check this and if it's ok merge it to master branch (after having seen the results from Travis, that are in the queue now).


GiovanniBussi:
Hi @gtribello I think this last push should solve all the issues. You can see the commits on the branch (notice that I rebased it, the one I uploaded a couple of days ago is outdated now). I commented every change, it should be easy to understand.

Some additional comments:
- Fixing stuff on mac allowed me to find some issue in the install names of the dylib (plus some crazy behavior of El Capitan), that is solved now.
- I made it such that it is not necessary to set PYTHONPATH to run the test. Basically, since the interface is installed *inside* plumed (see below however), plumed knows where to find it.
- Variables in the build script are inherited from the Makefile, and not obtained parsing `Makefile.conf`.

Remaining things to be fixed (perhaps after merge if you prefer):
- Could you add a destructor to the python class calling `plumed_finalize()`? Otherwise, we risk that some files are not closed and we loose data in the buffers. Plus, other bad things could happen if you have multiple plumed objects.
- There are still some "plumed version dependent" things in `plumed.pyx`. In particular, cmd strings `getBias` and `getDataRank` are treated specially, in the sense that a special warning is printed if the user tries to pass a non-array data. Even if you remove those checks, the user would receive an error anyway if he tried to pass non-array data, right? Could we then remove the check? I am saying this since I would like to decouple as much as possible the python interface so that it works with any plumed version. In case you need special information (e.g.: list of cmd strings with some special behavior) we might setup a new cmd string in order to retrieve the list.

Finally, I am still not completely satisfied about the way the plumed wrappers are installed. In particular, since typically one has multiple python versions installed and wrappers are not compatible, it is inconvenient to install the wrapper inside plumed. Imagine a cluster with 3 different python versions. One would have to install plumed 3 times, and to redo it whenever plumed is updated.

I think the only solution is to install the wrapper separately from plumed (they could even live on a different repository if we want! but it's not necessary). I don't have time to go more into this now, so for me the present implementation is fine since at least it allows running the tests on travis and have a working environment with a single python version. If you are ok with this, I would open an issue so that we don't forget it but merge this pull request as is.

gtribello:
Hello @GiovanniBussi 

thanks for doing all this stuff first of all.  I have looked and I am for you to merge these changes.  In terms of your questions:

(1)  Am perfectly happy to add the wrapper to plumed_finalize.  Do you want me to do that on this pull request?   If so how do I download the code for the pull request or can I do it github in some way.  If I can't do it this way I can add it on master.  It shouldn't take too long.

(2)  The plumed dependent thing in plumed.pyx are there for python reasons and I am not sure I can make them plumed independent.  The issue is that python cannot change the values of function arguments if those function arguments are ints or floats.  If I have the following python code:

def myfunc( myint, myfloat ):
       myint = 1
       myfloat=3.5

myint=0
myfloat=2.1
myfunc( myint, myfloat )
print( myint, myfloat )

Then the output would be 
0, 2.1
as the function cannot change the values of the arguments of the function as they are not (in python words) mutables.  The reason those variables are as they are is that a user could write 

p.cmd("getBias",bias)

and have bias be a float and legitimately expect bias to come back as the value of the bias.  It won't though because of a weird python thing.  (I know super annoying)  Anyway, given all this what would you suggest I do?  We could, for instance, insist that in future the keywords that have the word get in them must get data from plumed and thus cannot be passed a non-mutable object?

In terms of the whole different versions business, I see the problem but I don't know what the solution is.  The issue is that all of this distutils stuff is designed to work in the python way and this kind of avoids all this business of making and compiling.  Maybe what we have is a not bad compromise.  It is installed automatically with the default version of plumed on compile.  This makes it easy for everyone.  If a more advanced user wants to install it with a different version of python they can go to the python directory again and run the install scripts again with a different version of python.  We can even document how to do all this online in the manual?  

Incidentally, on the subject of the manual I was thinking about changing the installation section as at the moment I think there is a lot of information there and it not easy for people to find exactly what they need.  Also as I was reading countless crappily-written installation pages for various codes I had some thoughts about how to present all this information better.  At some stage, I hope to have some time to have a mess and work on my idea.

GiovanniBussi:
Great!

About `plumed_finalize`, I guess you can do it on master (not sure you can write of this branch on my repository).

About the warnings from pyx file, I am not sure I understand. Except for the conceptual mistake that cmd "getBias" should be called after cmd "calculate", it is like all the other cmds. I mean, there are other cmds that expect to modify the variable that you are passing.  The ones that I can find now are:
- getApiVersion
- setStopFlag
- isEnergyNeeded
- checkAction
- getBias

Shouldn't you add a check for all of them?

I think there are two solutions:
- You don't allow people using `p.cmd("whatever",whatever_is_not_an_array)`. To get the bias (or the Api version) they will have to pass an array of size 1.
- We make a new cmd that asks to plumed the list of all cmds that apparently accept a scalar but actually accept a pointer and you use that list to construct the python wrapper.

Does this make sense or is there something that I am missing?

About instruction for installation, I am happy if you want to change them. I agree that they are not well organized, however most of the information present I think should be there, plus there is some missing (e.g. on installation directories). Let's iterate on them.

Thanks!

Giovanni

PR #290, opened 2017-09-18
refactor EDS documentation to fit into Camilloni submodule format, an…

hockyg:
Refactor EDS documentation to fit into Camilloni (@carlocamilloni) submodule format, and added reference to new paper describing some features of the module.

##### Description

Updated EDS module documentation so that it now appears under Additional Modules in docs. Also edited bibliography file to include a new paper referred to in the revised docs. 

##### Type of contribution

##### Copyright

### DIFF
--- a/src/eds/EDS.cpp
+++ b/src/eds/EDS.cpp
@@ -36,7 +36,7 @@ using namespace bias;
 namespace PLMD {
 namespace eds {
 
-//+PLUMEDOC BIAS EDS
+//+PLUMEDOC EDS_BIAS EDS_EDS
 /*
 Add a linear bias on a set of observables.
 
@@ -124,7 +124,6 @@ eds: EDS ARG=dist,dist2 CENTER=2.0,1.0 PERIOD=50000 TEMP=1.0 IN_RESTART=restart.
 //+ENDPLUMEDOC
 
 class EDS : public Bias {
-
 
 private:
   /*We will get this and store it once, since on-the-fly changing number of CVs will be fatal*/
--- a/user-doc/EDS.md
+++ b/user-doc/EDS.md
@@ -1,0 +1,30 @@
+\page EDS Experiment Directed Simulation
+
+<!-- 
+description: Methods for incorporating additional information about CVs into MD simulations by adaptively determined linear bias parameters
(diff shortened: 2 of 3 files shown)


carlocamilloni:
Hi @hockyg, a couple of comments:

1) I think it is better if you push your request to be merged in v2.4 (we are using the new format from v2.4) as usual we will then immediately merged it to the master branch
2) there is no need to name your action as EDS_EDS, EDS is just fine (because it is only one action and would sound quite redundant)

This should also fix the travis-ci cppcheck error I think

hockyg:
Hi Carlo,

For 1, I'll try to change it

For 2, there is an error about recursive names in the documentation if the
function is called EDS, it doesn't seem to know how to build the table of
contents correctly. If you think that's not supposed to happen, you can
clone mine and test? Probably you can figure out what is happening.

Finally, since this is in the github thread, maybe you know the answer to
this other question? Perhaps it is related to the cppcheck error:

The docs I compiled are missing the usual list of components, etc
>


> Any idea if I need to include a new flag, or if I'm just compiling the
> docs wrong?
>


> Here is the new docs that I made:
>
> https://users.rcc.uchicago.edu/~glen842/plumed-eds-contrib-docs/user-doc/html/_e_d_s__e_d_s.html
>


> And here is the old one:
> https://plumed.github.io/doc-v2.4/user-doc/html/_e_d_s.html

carlocamilloni:
Hi Glen

I will check everything in the next days

Carlo

Sent from my iPhone

hockyg:
Thanks Carlo!

carlocamilloni:
Hi @hockyg 
The issue is that there is a circular definition between the EDS action and the EDS module. A solution that works is:
in EDS.cpp modify the doxygen to:  PLUMEDOC EDSMOD_BIAS EDS
then rename EDS.md to EDSMOD.md and change its first line to 
\page EDSMOD Experiment Directed Simulation
this also fix the issue with the components. Finally it is better to make the pull request to the v2.4 branch

Carlo

hockyg:
Great! I'll do it today. I also found a small bug that I'll try to get
committed, but the PRs shouldn't interfere with each other

PR #299, opened 2017-10-02
unique_ptr: fixes to ves

GiovanniBussi:
@valsson can you check if this is correct? Thanks!

Giovanni


valsson:
This looks right to me. These are basically the same changes that I had done. 

Thanks,
Omar

PR #311, opened 2017-10-30
Patch for quantum espresso 6.2

ralf-meyer:
##### Description

Adaptation of the quantum espresso 5.2.0 patch to the new file structure of quantum espresso 6.*

##### Target release

I would like my code to appear in release v2.4?

##### Type of contribution

- [x] changes to code or doc authored by PLUMED developers
- [ ] changes to a module not authored by you
- [ ] new module contribution or edit of a module authored by you

##### Copyright

- [x] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

- [ ] the module I added or modified contains a `COPYRIGHT` file with the correct license information. I also used the command `cd src && ./headers.sh mymodulename` in order to make sure the headers of the module are correct.

##### Tests

- [ ] I added a new regtest or modified an existing regtest to validate my changes.
- [ ] I verified that all regtests are passed successfully on Travis-CI.

PR #312, opened 2017-10-30
Added patch for QE 6.2

GiovanniBussi:
Replace #311

@ralf-meyer Could you please check if this is ok? Thanks

Giovanni


ralf-meyer:
Perfect, thank you!

Ralf

PR #315, opened 2017-11-06
Fixed warning

GiovanniBussi:
@gtribello can you double check this fix before I merge? Thanks


gtribello:
@GiovanniBussi that looks fine.  Go ahead and merge.

PR #319, opened 2017-11-22
Exceptions

GiovanniBussi:
@plumed/all can you have a look at this?

This request implements the following stuff:
- remove flag `--disable-cxx-exceptions` so that exceptions are always enabled
- add a try/catch block in PlumedMain so that errors are reported to plumed log

The important commit is this one: 307c4ef458fe0245a3192f1c6f85c4d40cb5fcef

The rationale is that any error arising within a `cmd` should be reported both on stderr and on the plumed.log. I noticed that in many parts of the code we just duplicate the code, preparing a message that is written both in `plumed_merror(msg)` and in `log<<msg`. This change would allow to remove the duplicate code. More importantly, it would make all errors raised with `plumed_merror` (also those from internal plumed tools that do not have access to log) reported on the log file.

What do you think?

### DIFF
--- a/configure
+++ b/configure
@@ -710,7 +710,6 @@ enable_debug_glibcxx
 enable_shared
 enable_dependency_tracking
 enable_rpath
-enable_cxx_exceptions
 enable_ld_r
 enable_ar_cr
 enable_mpi
@@ -1379,7 +1378,6 @@ Optional Features:
   --enable-dependency-tracking
                           enable dependency tracking, default: yes
   --enable-rpath          enable store rpath, default: no
-  --enable-cxx-exceptions enable c++ exceptions, default: yes
   --enable-ld-r           enable group object files, default: yes
   --enable-ar-cr          enable use ar to build libplumedWrapper.a, default:
                           yes
@@ -2665,24 +2663,6 @@ else
   case "no" in
              (yes) rpath=true ;;
              (no)  rpath=false ;;
-  esac
-
-fi
-
-
-
-cxx_exceptions=
-# Check whether --enable-cxx-exceptions was given.
-if test "${enable_cxx_exceptions+set}" = set; then :
-  enableval=$enable_cxx_exceptions; case "${enableval}" in
-             (yes) cxx_exceptions=true ;;
-             (no)  cxx_exceptions=false ;;
-             (*)   as_fn_error $? "wrong argument to --enable-cxx-exceptions" "$LINENO" 5 ;;
-  esac
-else
-  case "yes" in
-             (yes) cxx_exceptions=true ;;
-             (no)  cxx_exceptions=false ;;
   esac
 
 fi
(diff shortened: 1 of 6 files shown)


carlocamilloni:
It looks good to me!

Sent from my iPhone

maxbonomi:
It is ok to me.
I was wondering, if not already possible, if we can simply the error message.
Now, when encountering an error you get at the bottom a long list of things, like:

[chaos:43326] *** Process received signal ***
[chaos:43326] Signal: Abort trap: 6 (6)
[chaos:43326] Signal code:  (0)
[chaos:43326] [ 0] 0   libsystem_platform.dylib            0x00007fff73815f5a _sigtramp + 26
[chaos:43326] [ 1] 0   ???                                 0x00007ffe00020008 0x0 + 140728898551816
[chaos:43326] [ 2] 0   libsystem_c.dylib                   0x00007fff7364130a abort + 127
[chaos:43326] [ 3] 0   libc++abi.dylib                     0x00007fff71623f8f __cxa_bad_cast + 0
[chaos:43326] [ 4] 0   libc++abi.dylib                     0x00007fff71624113 _ZL25default_terminate_handlerv + 241
[chaos:43326] [ 5] 0   libobjc.A.dylib                     0x00007fff729a8eab _ZL15_objc_terminatev + 105
…

Could we avoid providing all these information by default and having a flag in configure to activate
this more verbose error log?

Max

carlocamilloni:
Yes I agree with Max, can we have this only if we compile in development mode?

C

Sent from my iPhone

> Il giorno 23 nov 2017, alle ore 09:14, Massimiliano Bonomi <notifications@github.com> ha scritto:
> 
> It is ok to me.
> I was wondering, if not already possible, if we can simply the error message.
> Now, when encountering an error you get at the bottom a long list of things, like:
> 
> [chaos:43326] *** Process received signal ***
> [chaos:43326] Signal: Abort trap: 6 (6)
> [chaos:43326] Signal code: (0)
> [chaos:43326] [ 0] 0 libsystem_platform.dylib 0x00007fff73815f5a _sigtramp + 26
> [chaos:43326] [ 1] 0 ??? 0x00007ffe00020008 0x0 + 140728898551816
> [chaos:43326] [ 2] 0 libsystem_c.dylib 0x00007fff7364130a abort + 127
> [chaos:43326] [ 3] 0 libc++abi.dylib 0x00007fff71623f8f __cxa_bad_cast + 0
> [chaos:43326] [ 4] 0 libc++abi.dylib 0x00007fff71624113 _ZL25default_terminate_handlerv + 241
> [chaos:43326] [ 5] 0 libobjc.A.dylib 0x00007fff729a8eab _ZL15_objc_terminatev + 105
> …
> 
> Could we avoid providing all these information by default and having a flag in configure to activate
> this more verbose error log?
> 
> Max
> 
> 
>

GiovanniBussi:
Hi,

I agree. I would do it with an env var (and not a configure option), so that if a user has a problem we can always say "please rerun with this env var set and report the message". However, I don't know how to remove the things Max is reporting because I think they are written by the operating system.

Now it works in this way:
- When there is an error (either internal or in the input) an exception is thrown. I think this is the right thing to do since it allows handling errors easily from python or from a C++ MD code.
- The exception contains a message that can be extracted with `what()` (this is standard)
- Currently, the message is like this one
````
+++ Internal PLUMED error
+++ file MDAtoms.cpp, line 296
+++ message: cannot create an MD interface with sizeof(real)==127
````
Perhaps it is too verbose? This is easy to change. I would still keep file/line number there, unless we are sure that all the `plumed_merror()` statements report a different message.
- In addition to this, the message also contains the stack trace (all the previous calls), but only if env var `PLUMED_STACK_TRACE` env var is defined (it is defined on travis). Normal users should not see it. This is something for developers (you might want to set it in your `.bashrc`).
- In this pull request, the very same message is also written on plumed.log. If you want this can be simplified easily (e.g., plumed.log might contain only the message, still keeping the whole text in the `what()` string)
- If the error is not caught anywhere, the operating system writes the `what()` string. However, as Max reported, it might also write other information. I don't know how to remove those...

In summary, the only thing I know how to do is, in the log file, only write the message (e.g. `cannot create an MD interface with sizeof(real)==127`). We can also shorten a bit the message that goes to stderr (I would still keep the line number there however). Let me know if you think it's worth.

GiovanniBussi:
As a note, this request was actually merged.

PR #343, opened 2018-02-23
V2.4 fix simd

GiovanniBussi:
It looks like there is an important bug in RMSD (and RMSD based: PCA, FIT_TO_TEMPLATE, etc) when using intel compiler and -O3. I think there were many incorrect `pragma omp simd` statements. I tried to fix them.

@jpazurikova could you have a look before I merge? Thanks!

Giovanni


jpazurikova:
Hi,
well, this is weird. I specifically remember doing these kind of repairs myself before. There was definitely a point in development where I had these errors in code, but then I fixed them. Maybe I merged a wrong commit? I am a bit confused but glad you noticed this.
All in all, changes make sense.
Regards,
Jana

GiovanniBussi:
@jpazurikova great so I will merge. Thanks!

PR #362, opened 2018-05-28
Improved plumed loader

GiovanniBussi:
##### Description

Dear all,

I open a pull request for this in case someone has comments. I plan to merge it next week.

I made significant changes to the way PLUMED is loaded when using `--runtime` option resulting in the following improvements:

1. It is now not anymore necessary to link the MD code calling PLUMED with the `-rdynamic` (or `-Wl,--export-dynamic`) option. I think this is useful in term of portability of the `Plumed.c` file, especially when it is directly included in other MD codes (as in Amber or Espresso++).
2. It should be possible to load PLUMED also if the `Plumed.c` file has been in turn loaded with `dlopen` flag `RTLD_LOCAL` or `RTLD_LAZY`

The latter is related to a problem that was reported in linking PLUMED within Espresso++ by a collaborator of @valsson.

The main changes are the following ones:

- The loader (`wrapper/Plumed.c`) now uses `dlsym` in order to find symbols in the `libplumedKernel` library, instead of relying on automatic registration. This allows loading the library also when the symbols in `wrapper/Plumed.c` are not visible.
- The `libplumedKernel` library still uses self-registration in order to be compatible with MD codes patched with `--runtime` before this fix (that is, including an old `Plumed.c` file). However, it does so by using `dlsym` as well. Thus, `libplumedKernel` will be loadable also if symbols of `wrapper/Plumed.c` are not available.
- To allow loading old versions of the `libplumedKernel` library which are not selfcontained and thus cannot be opened if `wrapper/Plumed.c` symbols are not visible, in case of a failure loading `libplumedKernel` we remove the `Kernel` portion of the file name and try to directly load `libplumed`. This seems to work correctly.

Some env var can be used to control the behavior:

- `PLUMED_LOAD_DEBUG` can be used to debug all the loading procedure. Just do `export PLUMED_LOAD_DEBUG=1` before launching the MD code.
- `PLUMED_LOAD_NAMESPACE` controls in which namespace PLUMED is loaded. One can use `export PLUMED_LOAD_NAMESPACE=LOCAL` to load it in a local namespace (`RTLD_LOCAL`) and avoid clashes if multiple PLUMED versions are loaded. This however makes the `LOAD` action not functional (since the loaded dylib cannot see the PLUMED library anymore). The default is `PLUMED_LOAD_NAMESPACE=GLOBAL` and is identical to the previous loader.

All combinations are expected to work (previously patched MD codes with the new library, newly patched MD codes with the old library, etc). In case anyone finds problem (also after I will have merged this) let me know.

Giovanni

### DIFF
--- a/CHANGES/v2.5.md
+++ b/CHANGES/v2.5.md
@@ -60,3 +60,14 @@ Changes from version 2.4 which are relevant for developers:
 - When launching `plumed`, flags `--no-mpi` and `--mpi` can appear multiple times. The last appearence is the effective one.
 - Internal blas and lapack libraries updated to gromacs 2018.
 - Choosing `./configure --prefix=$PWD` does not lead anymore to deletion of all header files.
+- Runtime loader in `Plumed.c` now works also when linked without `-rdynamic` (that is, 
+  its names are not exported). Notice that all the combinations are expected to
+  work, that is: `Plumed.c` from <=2.4 or >=2.5 combined with libplumedKernel
+  from <=2.4 or >=2.5. In order to achieve this the following changes are implemented:
+  - libplumedKernel does not depend anymore on `Plumed.c`. This allows loading it even
+    in cases where names in the loader are not visible. The relevant function needed
+    to be compatible with `Plumed.c` <=2.4 are found using `dlsym`.
+  - `Plumed.c` does not need anymore libplumedKernel to register itself, but rather
+    searches the relevant functions using `dlsym`. In addition, if it is not able to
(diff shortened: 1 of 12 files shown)


songbin6280:
Hello Giovanni,

Thank you very much for modifying the code. I have been successful in packaging Plumed.c and Plumed.h with ESPResSo++.

Bin

GiovanniBussi:
Great.

Please notice that I might change it again in the near future. The version that is now in the repository is expected to continue to work anyway.

Giovanni

PR #369, opened 2018-06-15
PIV-PathCV colvar added

SilvioPipolo:
##### Description

We added the PIV colvar that computes a "distance" between frames, based on the PIV definition, compatible with PathCV. 
The PIV.cpp file is in a moldule folder named "piv" with the COPYRIGHT .
We also added functions in the following files in the plumed master:
/tools/NeighborList.h
/tools/SwitchingFunction.h
/tools/PDB.h
/tools/PDB.cpp
/tools/NeighborList.cpp
/tools/SwitchingFunction.cpp
and modified the following file to allow the coupling with PathCV
/function/FuncPathMSD.cpp
All changes are highlighted by the Pipolo_Pietrucci keyword.

##### Target release

##### Type of contribution

##### Copyright

- [X] the module I added or modified contains a `COPYRIGHT` file with the correct license information. I also used the command `cd src && ./headers.sh mymodulename` in order to make sure the headers of the module are correct.

I couldn't find the headers.sh script

##### Tests


carlocamilloni:
Hi Silvio, thanks for sharing your PIV implementation 
I have just quickly look at the code, but at first sight there are two important things missing:
- a regtest to cover the new syntax, this could also be also done by extending the scope of an already existing regtest (for example regtest/basic/rt37)
- add an explanation about PIV in the documentation of FuncPathMSD as well as an example, you can also add references into user-doc/bibliography.bib

carlocamilloni:
Sorry, even more important probably, I cannot see the piv.cpp file in the commit

GiovanniBussi:
@SilvioPipolo thanks for your contribution!

There are a few other things missing:
- run `cd src && ./headers.sh piv`
- please use `git rebase` to split the contribution in multiple commits:
  - one commit with changes to the common part of the code
  - one commit with the piv module
  - one commit with regression test
- I think it would be much better if you split the contribution to the common part of the code in individual commits with proper comments.
- remove the `Pipolo_Pietrucci` tags from the common part of the code
- use `astyle` to fix formatting

I also have some comments about functions that you added.

In `SwitchingFunction.h`:
````
  /// Returns nn
    int get_nn() const;
  /// Returns mm
    int get_mm() const;
````
Notice that only rational switching functions define `nn` and `mm`. I think you should also add a function to check if the function is of rational type. In general, I suspect logging the switching functions as you do in PIV.cpp is not a good idea. Why not using `SwitchingFunction::description()`?

In `PDB.cpp`. Why did you add these members? Aren't they redundant?
````
  std::vector<AtomNumber> residueAN;
  std::vector<unsigned> index;
````
As a note for us: after carefully checking the columns read in record `CRYST1` are correct, we should document them the manual.

In `NeighborList.h`:
````
  std::pair<unsigned,unsigned> getOClosePair(unsigned i) const;
  std::pair<unsigned,unsigned> getOIndexPair(unsigned i) const;
````
Why are these returning unsigned? I think it is clearer if they return `std::pair<AtomNumber,AtomNumber>`. Notice that you can always extract the index later. Notice that we try to consistently use `AtomNumber` everywhere an int refers to the index of an atom, to make the code clear. There are exceptions here and there in the lowest level parts of the code (e.g. inside `MDAtoms.cpp`) but I think we should try to be consistent.

GiovanniBussi:
Some additional comment below.

### Header script

There is a typo in the instructions, sorry. Should be `cd src && ./header.sh`

### About changes to the core

I think function `getOClosePair` should be named `getClosePairAtomNumber` to make it clear that it is identical to `getClosePair` but returns an `AtomNumber` pair (after having it modified to return a pair of `AtomNumber`s as I suggested in my previous comment).

I also do not understand why you added function `getOIndexPair`. This function corresponds to `getIndexPair` but returns absolute atom numbers. However, notice that `getIndexPair` is private. I don't think you should need this function outside of the `NeighborList` class.

### About the PIV action

Notice that travis build is failing. Please check the end of [this log](https://travis-ci.org/plumed/plumed2/jobs/392876443) and fix the warnings.

Taking care of the following comments instead is optional, since potential problems will only affect the `piv` module.

`PIV::calculate()` contains a bunch of `static` variable. This is a bad idea in general, since they might lead to problems with multithread. In the specific case, if I understand the meaning of the variables that you are using, you want to save variables across iterations. Those variables should be included as members of the PIV class. Using `static` variables as you are doing will lead to problems if you use multiple `PIV` commands in the same plumed input
file.

You are using function `time` for timing, that I think measures seconds. I would recommend using `PLMD::Stopwatch` class.

SilvioPipolo:
Thank you GiovanniBussi for your comments. I'm ready for a new commit with the changes based on your and carlocamilloni's suggestions/comments. I'm now going to:
1) split the old commit as you suggest 
2) add new commits with the changes on top of the split old commit
There will be then two commits for each "topic" (common part topics, piv topic, regtest topic).
If you can confirm that this is ok with you, I'll do that tomorrow. Thanks

GiovanniBussi:
@SilvioPipolo ok thanks!

SilvioPipolo:
I'm afraid I messed the commits up trying to split the first commit with the common changes, would it be ok if I open a new pull request with the new changes (one commit per topic)? I'm sorry I'm not a git pro.

carlocamilloni:
Yes no problem!

Sent from my iPhone

PR #370, opened 2018-06-21
piv module

SilvioPipolo:
##### Description

Pull request to add the piv module by Silvio Pipolo and Fabio Pietrucci.

Chages to common plumed files
/tools/NeighborList.h
/tools/NeighborList.cpp
/tools/SwitchingFunction.h
/tools/SwitchingFunction.cpp
/tools/PDB.h
/tools/PDB.cpp
/function/FuncPathMSD.cpp
are separated into 4 commits.

piv module /src/piv is added as a separate commit.
/regtests/piv are added as a separate commit.

In the following the suggested changes of previous (closed) pull request are commented (comments in capital letters):

   There are a few other things missing:
    run cd src && ./headers.sh piv

    please use git rebase to split the contribution in multiple commits:
        one commit with changes to the common part of the code
        one commit with the piv module
        one commit with regression test
    I think it would be much better if you split the contribution to the common part of the code in individual commits with proper comments.
    remove the Pipolo_Pietrucci tags from the common part of the code
    use astyle to fix formatting

DONE

I also have some comments about functions that you added.

In SwitchingFunction.h:

  /// Returns nn
    int get_nn() const;
  /// Returns mm
    int get_mm() const;

Notice that only rational switching functions define nn and mm. I think you should also add a function to check if the function is of rational type. In general, I suspect logging the switching functions as you do in PIV.cpp is not a good idea. Why not using SwitchingFunction::description()?

THE FUNCTIONS WERE NEEDED TO SET THE SWITCHING FUNCTION AFTER HAVING CHANGED THE r_0 VALUE. NOW THE TWO FUNCTIONS HAVE BEEN REMOVED AND A "set_r0" FUNCTION HAS BEEN ADDED VALID FOR ALL SWITCHING FUNCTION TYPES.

In PDB.cpp. Why did you add these members? Aren't they redundant?

  std::vector<AtomNumber> residueAN;
  std::vector<unsigned> index;

YES SORRY DELETED.

As a note for us: after carefully checking the columns read in record CRYST1 are correct, we should document them the manual.

In NeighborList.h:

  std::pair<unsigned,unsigned> getOClosePair(unsigned i) const;
  std::pair<unsigned,unsigned> getOIndexPair(unsigned i) const;

Why are these returning unsigned? I think it is clearer if they return std::pair<AtomNumber,AtomNumber>. Notice that you can always extract the index later. Notice that we try to consistently use AtomNumber everywhere an int refers to the index of an atom, to make the code clear. There are exceptions here and there in the lowest level parts of the code (e.g. inside MDAtoms.cpp) but I think we should try to be consistent.

I think function getOClosePair should be named getClosePairAtomNumber to make it clear that it is identical to getClosePair but returns an AtomNumber pair (after having it modified to return a pair of AtomNumbers as I suggested in my previous comment).

I also do not understand why you added function getOIndexPair. This function corresponds to getIndexPair but returns absolute atom numbers. However, notice that getIndexPair is private. I don't think you should need this function outside of the NeighborList class.

THE DISCUSSED FUNCTIONS HAVE BEEN DELETED AND ALL IS BASED ON A NEW FUNCTION getClosePairAtomNumber AS SUGGESTED.

Taking care of the following comments instead is optional, since potential problems will only affect the piv module.

PIV::calculate() contains a bunch of static variable. This is a bad idea in general, since they might lead to problems with multithread. In the specific case, if I understand the meaning of the variables that you are using, you want to save variables across iterations. Those variables should be included as members of the PIV class. Using static variables as you are doing will lead to problems if you use multiple PIV commands in the same plumed input
file.

THANKS, THIS IS DONE ON PURPOSE IN ORDER NOT TO COMPUTE THE TRAJECTORY PIV VECTOR FOR ALL PIV COMMANDS IN THE INPUT FILE. THE USE OF STATIC VARIABLE IS LIMITED TO THIS PURPOSE.

You are using function time for timing, that I think measures seconds. I would recommend using PLMD::Stopwatch class.

THANKS NOW STOPWATCH IS USED

A

##### Target release

##### Type of contribution

##### Copyright

##### Tests


src/tools/SwitchingFunction.h
```diff
@@ -39,85 +40,88 @@ class Keywords;
...
-  double get_r0() const;
+    double get_r0() const;
 /// Return dmax
-  double get_dmax() const;
+    double get_dmax() const;
 /// Return dmax squared
-  double get_dmax2() const;
+    double get_dmax2() const;
+/// Set a r0 switching function.
+    void set_r0(double r_0);
```
GiovanniBussi: @SilvioPipolo notice that the switching function also stores `double invr0_2`. In addition, there are many other members that depend indirectly on `r0` (e.g. `stretch` and `shift`). I would recommend not allowing the parameters to be changed individually as you are doing.

Can you change the `PIV` class so that instead of setting `r0` you reset the whole switching function?

Giovanni
`

GiovanniBussi: @SilvioPipolo It also looks from github that you modified the whole file. Please can you add only the new change? You should use the correct version of astyle and avoid using some additional tool for reformatting.

src/tools/NeighborList.h
```diff
@@ -27,58 +27,61 @@
...
 /// Get the i-th pair of the neighbor list
-  std::pair<unsigned,unsigned> getClosePair(unsigned i) const;
+    std::pair<unsigned,unsigned> getClosePair(unsigned i) const;
 /// Get the list of neighbors of the i-th atom
-  std::vector<unsigned> getNeighbors(unsigned i);
-  ~NeighborList() {}
+    std::vector<unsigned> getNeighbors(unsigned i);
+    ~NeighborList() {};
+/// Get the i-th pair of AtomNumbers from the neighbor list
+    std::pair<AtomNumber,AtomNumber> getClosePairAtomNumber(unsigned i) const;
```
GiovanniBussi: @maxbonomi I think you implemented the first version of the neighbor list. What @SilvioPipolo is doing now is making a private member public. Is this correct?

src/tools/SwitchingFunction.h
SilvioPipolo: ok for the switching function. I'll try to do that. 
Concerning the files, I formatted them using astyle version
Artistic Style Version 3.1
on linux because of the errors I got from travis

GiovanniBussi:
@SilvioPipolo about static variables: I think this is a bad practice in general. Your variable will not work for instance if you have multiple threads acting on different trajectories. I don't understand why   you want to share information between separate instances of `PIV`. This would be very counterintuitive for the user. In case you want to keep the code like this, I would put a very big warning since this might interfere with the functionality of the code.

GiovanniBussi:
About failure on travis, please check the end of the log. I see these errors:
````
[tools/PDB.h:0] (error) :plmd_astyle: astyle not satisfied
[function/FuncPathMSD.cpp:0] (error) :plmd_astyle: astyle not satisfied
[piv/PIV.cpp:0] (error) :plmd_astyle: astyle not satisfied
[piv/PIV.cpp:1264] (error) :plmd_missing_namespace_module: missing piv namespace
[tools/NeighborList.cpp:0] (error) :plmd_astyle: astyle not satisfied
[tools/PDB.cpp:0] (error) :plmd_astyle: astyle not satisfied
[piv/PIV.cpp:247] (warning) :uninitMemberVar: Member variable 'PIV::docomp' is not initialized in the constructor.
[piv/PIV.cpp:247] (warning) :uninitMemberVar: Member variable 'PIV::dosum' is not initialized in the constructor.
[piv/PIV.cpp:247] (warning) :uninitMemberVar: Member variable 'PIV::docmdist' is not initialized in the constructor.
````
Most of them are related to astyle. However, you should also solve the warnings from cppcheck in the PIV module. Thanks!

**Comment by SilvioPipolo** [comment]
About static variables: thanks for the suggestions, the problem we're facing is that building the piv already slows down classical MD of 1/2 orders of magnitude (on a 3000 particle system), then building the trajectory piv for e.g. 2 piv instances results in basically doubling the computational cost. Any suggestion on how to build the trajectory piv once for all piv instances would be very much appreciated!

**Comment by GiovanniBussi** [comment]
Hi.

one way would be to compute two "PIVs" in the same action, although this
will require a more complex input. Thin about Z and S variables in PATHRMSD.

Another way would be to give the label of the first PIV instance to the
second one. The second one can then obtain the pointer from ActionSet, use
a dynamic_cast to convert it to a PIV pointer and then use some private
method of the PIV class to extract the already computed stuff.

Giovanni

**Comment by SilvioPipolo** [comment]
Got it, thanks I'll discuss with Fabio, how much time do we have before the release of the next beta version? Thanks

**Comment by SilvioPipolo** [comment]
Ok great, thanks a lot!
Silvio

**Comment by SilvioPipolo** [comment]
@GiovanniBussi Regarding your previous suggestion
"Another way would be to give the label of the first PIV instance to the second one. The second one can then obtain the pointer from ActionSet, use a dynamic_cast to convert it to a PIV pointer and then use some private method of the PIV class to extract the already computed stuff." 
could you tell me if there is a place in the code where something similar is done? Thanks a lot!

**Comment by GiovanniBussi** [comment]
@SilvioPipolo I think the simplest way to do it is this one.

Assume that:

- `label` is the label of the other PIV action (obtained with some `parse()` command).
- `Yourclass` is the name of your class (I think `piv::Piv`, though I think you forgot to use the `piv` namespace in your previous pull request).

This command will search for it and return a pointer:

````
auto p=plumed.getActionSet().selectWithLabel<Yourclass*>(label);
````

If the action does not exist or the class is wrong, p will be NULL:

````
if(!p) give_some_error();
````

Since `p` is of type `Yourclass*` you can then access its members as usual:

````
p->method();
````

You can probably find some example with `git grep selectWithLabel`

**Comment by GiovanniBussi** [comment]
@SilvioPipolo I think in the next days we are supposed to finalize the v2.5 beta release.

Did you manage to fix the module that you plan to contribute?

Thanks!

Giovanni

**Comment by SilvioPipolo** [comment]
@GiovanniBussi Sorry Giovanni, but unfortunately I've not yet fixed the static variable issue. Do you still think we can include this version with a warning on the static variables? I'll start working again on this problem on thursday.

File: src/tools/PDB.cpp [hunk 3]
```diff
@@ -515,6 +528,12 @@ bool PDB::readFromFilepointer(FILE *fp,bool naturalUnits,double scale) {
     string z=line.substr(46,8);
     string occ=line.substr(54,6);
     string bet=line.substr(60,6);
+    string BoxX=line.substr(7,8);
```
GiovanniBussi: @SilvioPipolo I think these are incorrect. Please check [PDB doc](http://www.wwpdb.org/documentation/file-format-content/format33/sect8.html#CRYST1). I think it should be `string BoxX=line.substr(6,9);` and `string BoxA=line.substr(33,7);` etc

GiovanniBussi:
@SilvioPipolo for me the priority for merging is to make sure the changes to the common part of the code are correct. If you want to leave the `static` variables I am fine with it, I will just add a warning about the fact that this optional function is not expected to work with multiple threads. Also, can you confirm that it does not work if you compute multiple PIV in the same input file?

src/tools/PDB.cpp
```diff
@@ -523,6 +542,33 @@ bool PDB::readFromFilepointer(FILE *fp,bool naturalUnits,double scale) {
...
+      Tools::convert(BoxB,BoxABG[1]);
+      Tools::convert(BoxG,BoxABG[2]);
+      BoxXYZ*=scale;
+      Vector Vcos;
+      double tmpr;
+      for (unsigned i=0;i<3;i++) {
+        Vcos[i]=cos(BoxABG[i]*3.1415926535/180.); 
+        Box[i][0]=0.; Box[i][1]=0.; Box[i][2]=0.;
+      }
+      BoxXYZ[1]=BoxXYZ[1]/BoxXYZ[0];
```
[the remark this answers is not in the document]
SilvioPipolo: This part of the code was taken from an old code, it works but I agree that the plumed/vmd expressions are simpler. With the new commit this has been updated to the plumed expressions

src/tools/PDB.cpp
```diff
@@ -523,6 +542,33 @@ bool PDB::readFromFilepointer(FILE *fp,bool naturalUnits,double scale) {
...
+      BoxXYZ[2]=BoxXYZ[2]/BoxXYZ[0];
+      tmpr=sqrt(1.-Vcos[2]*Vcos[2]);
```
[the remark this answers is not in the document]
SilvioPipolo: This part of the code was taken from an old code, it works but I agree that the plumed/vmd expressions are simpler. With the new commit this has been updated to the plumed expressions

SilvioPipolo:
@GiovanniBussi we never tried with multiple shared-memory threads, with pure MPI it works. The static variable containing the PIV associated to the configuration of the running trajectory is used in order not to compute the PIV twice (or multiple times) if more than one PIV sections is specified as input, so it works with multiple PIV definitions in the input file. If you agree i'll open a new pull request and start working on this, if we don't manage to include this in the first version we'll do it for sure in the following one. What do you think?

src/tools/NeighborList.h
```diff
@@ -78,7 +78,9 @@ class NeighborList
   std::pair<unsigned,unsigned> getClosePair(unsigned i) const;
 /// Get the list of neighbors of the i-th atom
   std::vector<unsigned> getNeighbors(unsigned i);
-  ~NeighborList() {}
+  ~NeighborList() {};
```
GiovanniBussi: @SilvioPipolo the `;` at the end of this line is redundant and should be removed I think (compilers give warnings from time to time on this stuff).

src/piv/PIV.cpp
```diff
@@ -0,0 +1,1234 @@
...
+
+#include <string>
+#include <cmath>
+#include <iostream>
+
+using namespace std;
+
+namespace PLMD
+{
+namespace colvar
```
GiovanniBussi: @SilvioPipolo this should be `namespace piv`

GiovanniBussi:
@SilvioPipolo I checked again carefully all the changes to the core part of the code and, with the exception of the two (mostly aesthetic) changes I just suggested, for me the code can be merged. There are [trivial conflicts to fix](https://github.com/plumed/plumed2/pull/370/conflicts).

Concerning the static variables, my doubt is what happens if you have multiple PIV variables e.g. based on different atoms. Isn't the second instance trying to reuse inappropriate data from the first one?

In any case, for me it is ok to merge and you might optionally try to fix these static variables later. Let me know after you have fixed the small changes. Actually, rather than merging I would do myself the rebasing in order to keep the changes to the core separate from the rest. If it's ok for you, you will then have to reset your branch to the new master.

GiovanniBussi:
Replaced with #384 (squashed commits and rebased)

Issue #195, opened 2016-07-14
More strings for MOLINFO


GiovanniBussi:
As of now we can recognize anything in the form:
XXX-YYY
where XXX is the name of a special group or the type of an atom, and YYY is the number of a residue (optionally preceded by the letter identifying the chain).

I would like to add other strings:
nucleic  : all atoms in a DNA or RNA residue
protein  : all atoms in a protein residue
water    : all atoms in a water residue
ions       : all atoms with names matching those of ions
hydrogens : all hydrogens of the system
nonhydrogens : all heavy atoms

Other ideas are welcome!

tonigi:
Perhaps atom specifications like :15.A@CA to indicate CA of residue 15 chain A.  Or any other choice of separator. The colon/dot/@ syntax follows Amber's and Chimera's http://www.cgl.ucsf.edu/chimera/1.1700/docs/UsersGuide/midas/atom_spec.html . While I guess vmd's would be too complicated to implement.

GiovanniBussi:
@tonigi this issue is related to special strings (e.g. phi/psi) not to atom selection. Atom selection was #32 and has been sort of closed.

Let me summarize the status of molinfo strings:
- Strings in the form `@XXX-YYY` are currently considered as special strings.
- `YYY` should be the number of a residue, optionally containing a chain number.
- `XXX` should be a special group name (`chi`, `psi`, etc) or an atom name (`CA`).

Thus your selection would be `@CA-15A`. The reason to start with `@` + letter is that `@` is a reserved prefix for unlabelled plumed actions which receive a default label equal to `@num` where num is their progressive number. This allows to avoid name clashes.

The limitations now are that:
- only proteins and nucleic acids are detected (no way to do `@OW-1000` to pick 1000th water oxygen)
- I would like to add extra special groups without the residue number (e.g. `@nucleic`)

I think both are relatively easy to fix

PR #372, opened 2018-07-11
Fix 195

GiovanniBussi:
This branch addresses #195 by adding the following shortcuts. I open a pull request so that if anyone has comments on the names of the shortcuts I can change them before merging.

Always available:
- `@all` all atoms
- `@physical` physical atoms
- `@virtual` virtual atoms

Available when using MOLINFO:
- `@nucleic`
- `@protein`
- `@water`
- `@ions`
- `@hydrogens`
- `@nonhydrogens`

In addition when using MOLINFO, individual atoms can be chosen also from water molecules (e.g. `@OW-100`).


carlocamilloni:
@GiovanniBussi Good job! I am not sure about the global names, in particular because "virtual" could refer also to the virtual atoms in the system (for example the additional atom in TIP4P or the additional virtual particles that can be used to fix the position of hydrogens) also "physical" is not really convincing. Actually, I think I would remove these keywords and keep only the one defined with MOLINFO.

GiovanniBussi:
Thanks @carlocamilloni !

At least `@physical` is very useful. I use a lot `WHOLEMOLECULES`/`WRAPAROUND` + `DUMPATOMS` and it is boring to remember how many atoms I have in the system and in which order they are when I dump water molecules as well. But I agree that the name is not appropriate.

Alternative names? Maybe:

- `@allatoms` for MD atoms + virtual atoms (now `@all`)
- `@mdatoms` for MD atoms (now `@physical`, bad name)
- no shortcut for virtuals to avoid the ambiguity that @carlocamilloni mentioned

Indeed I don't think it would be particularly useful to only dump virtual atoms. One can always use `GROUP ATOMS=@allatoms REMOVE=@mdatoms` to obtain them without explicitly listing them.

PR #373, opened 2018-07-16
FaMetaD

valsson:
##### Description

Adds the frequency adaptive metadynamics scheme described [here](https://doi.org/10.1063/1.5024679) to the metadynamics action. 

Also added a note in the manual regarding the calculation of the acceleration factor when restarting. Added also a warning if the previous value is not read-in when restarting. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@valsson thanks! Just one small issue: I guess with MPI-based multiple walkers the code will lock (if different walkers pick different strides). Is that right? Is there an easy solution? Otherwise it might be better to crash if an user enables both this feature and multiple walkers.

valsson:
One further thing is how the acceleration factor is calculated. 

Currently the whole sum of the exp(\beta*V) values is stored. Then when needed the average is obtained by dividing by the number of steps (see line 1675 in MetaD.cpp).

I think it would be simpler to calculate the average as a running average 

ave[n] = ave[n-1] + (value[n]-ave[n-1])/n

and just store the average. This would for example simplify the code needed for restarting, etc.

valsson:
@GiovanniBussi good question regarding the MPI mwalkers, I hadn't thought about it. 

Yes, they should get different strides which would lead to problems. However, the whole idea of this scheme is for getting kinetics like with infrequent metadynamics where it doesn't really make that much sense to use mwalkers. 

Therefore, I think the solution is too put a check and exit with an error if the user tries to do that. I will add that.

GiovanniBussi:
@valsson thanks for fixing the multiple walkers stuff.

About the acceleration factor, how would you like to solve it? If I understand correctly, what you propose would lead to an incompatible change. In particular, someone might restart with v2.5 a simulation initiated with v2.4 and get an incorrect result without noticing it. It would be ideal if we can have some warning. In any case it should be specified at the beginning of the changelog.

I actually never used the acceleration keyword, so maybe other people can comment on this.

valsson:
@GiovanniBussi I don't think the change will be incompatible. It should be an internal change in the code that should not change the results in any way. But I need to test to be sure about that. I will test in  an separate branch and let you know.

valsson:
The changes to the calculation of the acceleration factor that I am proposing are the ones [here](https://github.com/valsson/plumed2/commit/242bfb629bcf3cb5517c5173d9ca953274aa8759). 

There is no change in the results if the calculation is restarted correctly using the ACCELERATION_RFILE keyword. There is only a difference if continuing a run and not reading in the previous value of the acceleration factor as seen in the rt-acc-1 test (the md1 run). The old way leads to even wronger results as the value becomes smaller than 1.

valsson:
@GiovanniBussi just a reminder on this. 

The remaining question is if the calculation of the acceleration factor should be changed as I proposed. Honestly, it is not really needed, perhaps it is best to make minimal changes to the code. Are there any other things that need to be checked before merging?

GiovanniBussi:
Hi! No prob if nobody else does it before I will merge the branch before the end of the month. Thanks! Giovanni

PR #384, opened 2018-09-12
Piv merge

GiovanniBussi:
Squashed pull request #370
 
@silviopipolo please check if everything is as you expect.

Once tests on travis pass and you agree, I will merge it on master.

Thanks for you contribution!

Giovanni


SilvioPipolo:
@GiovanniBussi Thanks Giovanni, for fixing my imperfections! All is fine for me. Tests run fine. I see 1 failing by travis, that I don't understand.

GiovanniBussi:
When we validate pull requests we run more tests, including one on mac and MPI. Apparently, some of the tests do not work correctly there, I think it is because they use too many processors. No problem.

GiovanniBussi:
@SilvioPipolo it looks like your module is not tested: https://plumed.github.io/coverage-master/coverage/piv/PIV.cpp.gcov.html

Can you check why? Check if the makefiles in the regtest directory are correct

Issue #250, opened 2017-06-20
Simple algebra in input values


GiovanniBussi:
Currently, whenever we read a double, we check it first against a few special strings, namely those in the form Xpi (with X a double). Exploiting either libmatheval or lepton it would be easy to also parse the string to solve simple expressions, such as:

    RESTRAINT ARG=a AT=1+0.3

Instead of implementing it with libmatheval, I would rather wait for #244 so as to implement it directly using lepton. In this manner it would be always available.

An advantage of lepton (wrt libmatheval) is that it allows constants to be added to the pool. This would even allow defining some global parameters, e.g.

    SET_GLOBAL VAR=lambda VALUE=replicas@{0.0,0.2,0.4,0.6,0.8,1.0}
    RESTRAINT ARG=a AT=1+lambda*0.5

In this example `lambda` would be a constant similar to `pi`, but defined specifically on different replicas.

GiovanniBussi:
Simple algebra is possible as of 7c5b10470bf5806c7b431e71125b91310101fa64

I leave this issue open for the implementation of SET_GLOBAL action, which is non trivial and
will require some time.

PR #387, opened 2018-09-14
Atomic units

valsson:
##### Description

Add atomic units for energy (Hartree), length (Bohr), and time. Could be useful for ab initio MD codes. All numerical values based on CODATA 2014 (https://physics.nist.gov/cuu/Constants/).

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Thanks @valsson ! Would it be worth mentioning them in `src/setup/Units.cpp` as well so that they appear in the manual?

valsson:
@GiovanniBussi, yes I agree. I will add a note to the manual. 

Looking at the code I was also thinking that the unit parsing could be improved. Like for example allowing both to use "A" and "Angstrom". Maybe also make the parsing case-insensitive. Also give a better error message if the unit string is written incorrectly (currently it gives an error "energy units should be positive". 

I could try to make some these improvement and commit to this pull request. What do you think?

GiovanniBussi:
@valsson I am not sure allowing more choices is necessarily better.

In principle, now that constants are parsed using lepton, we could even redefine constants to have units in input values. E.g. `AT=3*nm_` where nm_ is set to 1 if units are nm, 10 if units are Angstroms etc. This is not possible yet since constants are global now, but once #250 will be closed it will be very easy (the set of constants will be contained in the PlumedMain object). Actually we never discussed about this, so maybe it is not even an interesting feature.

Anyway, if we increase the number of equivalent strings, allowing something like this would be a mess.

What do you think?

Of course this does not apply to your last suggestion (better error message), which I am sure would be absolutely an improvement.

valsson:
Ok, I see you point, to have better choices it not necessarily better, especially if something like you suggest would be implemented. I like the idea to be able to have unit in the input values, that could be useful. 

To be consistent with the rest, I think there should be only one option for Hartree energies, do you think that "Ha" is ok or is that too ambiguous. 

I will then make some changes to the error message and push to the pull request.

GiovanniBussi:
It looks like Ha is the [official abbreviation](https://en.wikipedia.org/wiki/Hartree) so I would use that (consistently with A/nm/etc).

Thanks a lot!

valsson:
I pushed some changes to the pull request:
- Improved error handling for setting units in tools/units.cpp
- Updated manual for UNITS, possible choices for units are listed
- Improved output for UNITS such that numerical values are given in the log file

PR #388, opened 2018-09-18
Flying Gaussian

spiwokv:
Dear all,
Here is the introduction of Flying Gaussian method into Plumed.

##### Description

* Flying Gaussian method  (Šućur & Spiwok J Chem Theory Comput 2016, 12(9), 4644-4650) is introduced here by Jana Hozzova (Pazurikova) and me as an extension of METAD keyword with FLYING_GAUSSIAN option (METAD .... FLYING_GAUSSIAN ....). It can be used a similar way as the MPI version of multiple walker metadynamics (requires WALKERS_MPI). When triggered, hills are not deposited but instead they move with evolution of walkers. Sampling enhancement is achieved by accumulation of multiple walkers in one minimum. We advise to use PACE=1 (hills position updated in every step). HILLS file is empty. Free energy surface can be calculated from COLVAR files as described in our recent article (Kříž, Šućur & Spiwok J Phys Chem B 2017, 121(46), 10479-10483).

* A combination of Flying Gaussian with parallel tempering (keyword FLYING_GAUSSIAN_PT) is
experimental and has not been published yet. We can remove this if necessary.

* Regtest are provided in basic/rt81.

##### Target release

I would like my code to appear in release 2.5

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
Hi maybe you can split the non PT pull request and the PT one in two different requests, because the first touches less files and patches

spiwokv:
Thanks @carlocamilloni,
I have removed the PT version since it is not published and I left only the non-PT version.
Regtest is there
Tested by Travis-CI
Vojtech

spiwokv:
Grazie mille!

PR #390, opened 2018-09-19
piv regtest 

SilvioPipolo:
##### Description

fixing problems in piv regtest

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@SilvioPipolo please check why some tests are failing: https://travis-ci.org/plumed/plumed2/jobs/430512850

carlocamilloni:
I think you need to run allgather only if(comm.initialized())

@GiovanniBussi I don't remember why comm.Sum does not require this

GiovanniBussi:
I think comm.Sum has a meaningful default, that is don't do anything:

https://github.com/plumed/plumed2/blob/6f5bb86560eb93fa7cdb2f1e1e3367740a18a6f4/src/tools/Communicator.cpp#L134-L140

comm.Allgather instead does not know what to do and crashes. Perhaps we could implement a meaningful thing also here, that I think would be copying the input array to the output one.

carlocamilloni:
Yes I think this would make much sense and simplify the implementation of mpi code (@SilvioPipolo so it is not enough to protect the code inside the if you also need to initialise your vector in the else{}) Even if I would prefer to make the change in Communicator

GiovanniBussi:
I agree it's better to implement it in the communicator. It could be something like:

````
plumed_assert(in.size==out.size);
plumed_assert(in.type==out.type);
if(s==NULL) return;
for(unsigned i=0;i<in.size;i++) r[i]=s[i];
````

Something similar could be done for Allgatherv

Can someone check if this fixes the issue?

GiovanniBussi:
No sorry, you cannot copy element by element, it's a void pointer... I have to check better

GiovanniBussi:
I think the type information is lost now when not using MPI
````
#ifdef __PLUMED_HAS_MPI
template<> MPI_Datatype Communicator::getMPIType<float>() { return MPI_FLOAT;}
template<> MPI_Datatype Communicator::getMPIType<double>() { return MPI_DOUBLE;}
template<> MPI_Datatype Communicator::getMPIType<int>()   { return MPI_INT;}
template<> MPI_Datatype Communicator::getMPIType<char>()   { return MPI_CHAR;}
template<> MPI_Datatype Communicator::getMPIType<unsigned>()   { return MPI_UNSIGNED;}
template<> MPI_Datatype Communicator::getMPIType<long unsigned>()   { return MPI_UNSIGNED_LONG;}
#else
template<> MPI_Datatype Communicator::getMPIType<float>() { return MPI_Datatype();}
template<> MPI_Datatype Communicator::getMPIType<double>() { return MPI_Datatype();}
template<> MPI_Datatype Communicator::getMPIType<int>() { return MPI_Datatype();}
template<> MPI_Datatype Communicator::getMPIType<char>() { return MPI_Datatype();}
template<> MPI_Datatype Communicator::getMPIType<unsigned>() { return MPI_Datatype();}
template<> MPI_Datatype Communicator::getMPIType<long unsigned>() { return MPI_Datatype();}
#endif
````

So at that stage you don't know anymore which type you are looking at and you have no way to copy the buffer. It can be done, but will take some time so maybe for the PIV stuff it is better to fix the Piv.cpp file.

SilvioPipolo:
@GiovanniBussi @carlocamilloni : I used comm.initialized() to avoid using the allgathers in the piv tests. I get no errors in my local version but travis still fails.

GiovanniBussi:
I think there are failures when using debug flags. That probably means that
you access an array out of the boundaries.

Search for the word "FAILURE" in the travis log

Also I think the cppcheck is failing (check the end of the log to see the
reason)

Giovanni

PR #398, opened 2018-10-13
V2.4 set omp

carlocamilloni:
##### Description

set the number of openMP threads from the MD code
furthermore call get_env only once (measurably faster)
gromacs patch updated to use it

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
Good points, I will update the openmp code on this line (safer static variable
Initialisation)  to remove the getenv call from 2.4 and add the api in plumed 2.5 (I think that users are not using openmp that much because they are not aware of it and the automated setting should help also in finding  performance regressions like the one just fixed)

C

Sent from my iPhone

carlocamilloni:
@GiovanniBussi API removed, I will implement it later in 2.5. Now the get_env has priority over whatever is gonna be set. get_env calls are called once.

Issue #328, opened 2018-01-09
Segmentation fault whenever more than 2 CVs are used for METAD with REWEIGHTING_NGRID


OleinikovasV:
PLUMED consistently breaks at a REWEIGHTING_NHILLS step whenever I run METAD with REWEIGHTING_NGRID using more than two arguments (it works fine with one and two). Does not drop any useful error messages other than "Segmentation fault".
I need this to obtain configuration weights using REWEIGHT_METAD for my simulation of METAD using 3 CVs.

valsson:
Hello, 

I have a feeling what might be the problem. Can you do the changes shown [here](https://github.com/valsson/plumed2/commit/a81f009de1f5d2bd1656b4024cd25bbcfbae2c84), recompile and check if it works. 

Omar

ladislav-hovan:
Hi Omar,

I was also looking at this issue, and with the changes you've suggested, the "Segmentation fault" message is replaced by "Floating point exception (core dumped)". However, I've noticed something else - on line 1862, an attempt is made to access the (i-1)th element of the rewf_grid_. The variable i starts with the value of rank, and there is no reason why it can't be zero. Indeed, if plumed is compiled without MPI, it will always be zero. This would instantly cause the segfault to happen (accessing the -1st element of the array). However, since I'm not sure about how the formula works, I am unable to fix it.

Ladislav

PR #399, opened 2018-10-17
Fix: Segmentation fault whenever more than 2 CVs are used for METAD w…

fiskissimo:
##### Description

Bugfix: Segmentation fault whenever more than 2 CVs are used for METAD with REWEIGHTING_NGRID #328

##### Target release

I would like my code to appear in release 2.5

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -1964,8 +1964,8 @@ void MetaD::computeReweightingFactor()
   for(unsigned i=rank; i<ntotgrid; i+=stride) {
     t_index[0]=(i%rewf_grid_[0]);
     unsigned kk=i;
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
@valsson I think you first implemented these lines. Could you double check the change?

If this is correct, I guess it should be backported to 2.4 at least (and perhaps 2.3, though we just declared it as non-supported).

valsson:
There is for sure some bug with the c(t) calculation for 3 or more CVs that has to be fixed. 

Now, it is actually @gtribello that implemented the integration where the bug/fix is. So it is better if he would take a look at the exact changes.  

Furthermore, I think it would be safer to change the unsigned int on lines 1964 and 1966 to  Grid::index_t. 

Looking a travis, is seems that the change breaks the regtest (rt67) for 2 CVs, there are some small numerical difference.

Issue #400, opened 2018-10-25
Restyling for reweighting implementation


invemichele:
Hi everybody,
I think there is room for improvement in the current implementation of the reweighting, and I would be happy to contribute.
In particular the way the integral for _c(t)_ is now calculated is, I believe, unnecessarily cumbersome. It comes directly from the `Grid::integrate()` function, which is not used anywhere else in the code (or at least not in the master branch). 
I think that a simple direct summation over the grid points is faster, works in any dimension (solving #328 ), and does not have any downside effect since the reweighting is only available when using a grid bias.
This also avoids calling `getBiasAndDerivatives` which unnecessarily returns the derivatives.

I implemented this changes in a pull request, together with the support for the non well-tempered case (I don't see why not having it) and a fix for the overflow of _c(t)_ in case of very high bias values.
Please let me know if I am missing a piece of the picture, and I am misunderstanding the code.

More in general I believe that also the user interface should be changed. In particular it does not make much sense to calculate the reweighting on a grid different from the bias one, and thus ask the user for `REWEIGHTING_NGRID`. especially because (if I understand the code correctly) the method called is `Grid::getValue(const vector<double> & x)`, which calls `Grid::getIndices(const vector<double> & x)`, which returns `floor((x[i]-min_[i])/dx_[i])`, thus as a matter of fact, if we double the number of reweighting bins, there is no difference in the value of _c(t)_, except for some small border effect.
Thus `REWEIGHTING_NGRID` could be dropped in favor of a flag like `CALC_RCT` or `CALC_REWEIGHTING`. It is fine to leave the `REWEIGHTING_NHILLS` option, but I would change the default frequency to 1, and let the user choose if increasing it.
Of course these are just suggestions, driven by my typical usage, but might not be best suited for all the possible scenarios.

I hope this can be useful.

valsson:
Hi @invemichele, 

I would agree with simplifying the integral calculation as you suggest. 

I just have one suggestion on the changes. Currently the integration is just done by summing over the grid points (and as c(t) is a ratio of two integrals there is no need to consider the volume of the grid bins). However, strictly speaking this is not correct numerical integration, it would be better to use something like trapezoidal rule for the integration so that the points at the edges have different weights. In many cases the integrated functions might be zero at the edges of the grid so this would not matter. But anyway I think it would be better to do it exactly. 

[...]

PR #401, opened 2018-10-25
restyling for the reweighting c(t) calculation

invemichele:
##### Description

Alternative implementation for the calculation of the reweighting constant _c(t)_. 
See #400 for details. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests

There are some small numerical differences in the multidimensional cases.


gtribello:
Hello @invemichele 

If you want to get rid of the double grid that is fine.  You are probably right that it doesn't make much of a difference and you can just go over the grid.

On the frequency of update of the update of c(t), however, I wouldn't make the default on this 1 as my understanding is that the performance of the algorithm is crap if you update the estimate of c(t) every time you add a hill.  In fact this is what I did the first time I implemented the algorithm and Omar and I could never get this setting to give a sensible estimate of the FES.  I would thus have the input be that the user has to set REWEIGHTING_HILLS.

valsson:
> I wouldn't make the default on this 1 as my understanding is that the performance of the 
> algorithm is crap if you update the estimate of c(t) every time you add a hill. In fact this is 
> what I did the first time I implemented the algorithm and Omar and I could never get this 
> setting to give a sensible estimate of the FES.

Regarding this, this was when using eq 14 in the JPCB 2015 paper, there the stride between updating c(t) matters as it is calculated as a finite difference. This is what we tried first where we saw this behavior. 

However, after consulting with Pratyush, we decided to use another more simpler equation to calculate c(t). This is equation 22 in the Ann. Rev. Phys. Chem 2016 paper and can be obtained by using equation 12 in equation 3 in the JPCB 2015 paper. There c(t) only depends on the current bias and is not calculated as a finite difference. Therefore, how often you update c(t) will not affect the results. It is only a question of overhead of how much the calculation costs. 

Omar

invemichele:
I believe the error introduced with the single grid is very small, and furthermore is a systematic constant shift, so from a physical point of view should not affect the result of the reweighting.

From the performance side, I agree that there should be a keyword to change the stride of _c(t)_ update. I tried to show in #400 that even for a fast system (alanine in vacuum) with a big 300x300 grid there is very little overhead, so I think the default stride should be 1, but I don't mind also leaving it to 50, for sure is not going to be a problem.

From what I understand, the history of the _c(t)_ reweighting is a quite twisted one, and the old implementation is somehow a reflection of this. It would be nice to make it more straightforward for the future!
Regards,

Michele

carlocamilloni:
I looked into the code and it is clear that the impact on performances is negligible, it is a few exp/log operations every time a Hill is added so unless this is done extremely frequently I don’t see any issue and also in that case one can still set the c(t) update to a higher period

GiovanniBussi:
@carlocamilloni does this change syntax in an incompatible way? If so we should also annotate it in the changelog.

carlocamilloni:
Yes the changelog must be updated

PR #403, opened 2018-11-02
Small fix to allow loading v2.5b

GiovanniBussi:
Since 4e3774ffbacb5e5836e3f97eb598a82ed9f423be the gromacs 2018 patch uses cmd "setNumOMPthreads", that is however not available in plumed 2.5b. I would like to merge this small fix so that it will be possible to load v2.5b from a gromacs version patched with the later patch.

@carlocamilloni what do you think?


carlocamilloni:
It’s good but maybe a bit an over killing for a beta version?

Sent from my iPhone

GiovanniBussi:
I agree this is not necessary I close it...

PR #404, opened 2018-11-05
Added patches for NAMD (versions 2.12 and git-latest 2.13b2)

dvdesolve:
Hello! I've created patches for the NAMD 2.12 and latest git version (2.13b2). NAMD compiles with them successfully and we're using it for about a year with no problems.

There is only one point for you to know. With the transition to the charm-6.8.2 in the latest git release of NAMD the shared-linking option during patching not working anymore - seems like new `charmc` have a nasty bug and when you're trying to compile NAMD with shared-linked PLUMED `charmc` can't find necessary Charm++ libs at the last stage of linking `namd2` executable. The only way to use PLUMED with the latest NAMD is to employ static linking.

PR #405, opened 2018-11-05
V2.4 namd patches

GiovanniBussi:
I would propose to update NAMD patches also in stable v2.4. Notice that NAMD 2.12 dates back to December 2016 and PLUMED 2.4 is one year younger.

I just rebased #404, kindly contributed by @dvdesolve, on v2.4 and removed the old patches. I would like to merge this on v2.4, let me know if anyone has objections. Thanks!

Giovanni


dvdesolve:
Also works for 2.3.x version

GiovanniBussi:
@dvdesolve technically 2.3 is not supported anymore so I would not add the patch there.

Notice that properly done patches are basically portable across versions. If you use `patch --runtime`, you should be able to patch an MD code using PLUMED 2.5 and later load PLUMED 2.0 setting `PLUMED_KERNEL` correctly.

Issue #406, opened 2018-11-05
NAMD 2.13 seems not to work with patch --shared


GiovanniBussi:
This has been reported in #404 by @dvdesolve.

It looks like we are making some mess in the Makefile while adding plumed flags.

dvdesolve:
Nevertheless I think that it's problem with static/shared charm-6.8.2. I've just rebuilt NAMD-2.13 with charm-6.8.2 with the next options:

## charm-6.8.2
`./build charm++ multicore-linux64 --build-shared --with-production`

## NAMD-2.13
```
./config Linux-x86_64-g++ --charm-arch multicore-linux64 --with-fftw3 --with-cuda
/opt/plumed/plumed24/bin/plumed patch -p -e namd-2.13
cd Linux-x86_64-g++
make
```

After that there are a lot of `*.so` files in `lib_so` directory and NAMD builds successfully (but some warnings are shown after linking)
[build-with-shared.txt](https://github.com/plumed/plumed2/files/2550333/build-with-shared.txt)

And there are dynamic dependencies for the binary `namd2`:
```
namd2
	linux-vdso.so.1 (0x00007ffe9a798000)
	/opt/plumed/plumed24/lib/libplumed.so (0x0000148b0ea87000)
	libmoduleGreedyRefineLB.so => /home/viktor/build/src/namd/Linux-x86_64-g++/.rootdir/charm-6.8.2/multicore-linux64/lib_so/libmoduleGreedyRefineLB.so (0x0000148b0ea74000)
	libmoduleCkLoop.so => /home/viktor/build/src/namd/Linux-x86_64-g++/.rootdir/charm-6.8.2/multicore-linux64/lib_so/libmoduleCkLoop.so (0x0000148b0ea63000)
	libmoduleCkMulticast.so => /home/viktor/build/src/namd/Linux-x86_64-g++/.rootdir/charm-6.8.2/multicore-linux64/lib_so/libmoduleCkMulticast.so (0x0000148b0ea5e000)
	libmoduleGreedyLB.so => /home/viktor/build/src/namd/Linux-x86_64-g++/.rootdir/charm-6.8.2/multicore-linux64/lib_so/libmoduleGreedyLB.so (0x0000148b0ea4e000)
	libmoduleRefineLB.so => /home/viktor/build/src/namd/Linux-x86_64-g++/.rootdir/charm-6.8.2/multicore-linux64/lib_so/libmoduleRefineLB.so (0x0000148b0ea41000)
[...]

PR #409, opened 2018-11-08
Tentative fix for #406

GiovanniBussi:
It should prepend the libplumed.so library with prefix -Wl,
so as not to disturb NAMD makefile. (suggested by @jcphill)


GiovanniBussi:
(the correct branch is v2.4-namd-patches3)

PR #411, opened 2018-11-21
Improved Python wrappers

GiovanniBussi:
I open this pull request to propose some changes in the way the python wrappers are installed. Currently, you should select the python version while installing PLUMED, and the wrapper will be installed inside PLUMED. In short, I would like to allow users to install the wrappers independently of PLUMED itself. This would allow me to correctly package the wrappers for MacPorts.

I would appreciate feedbacks especially from people using the python interface or potentially interested in using it (cc: @gtribello @ceriottm @dwhswenson @tonigi et al).

## Motivation

I am trying to package the python wrappers for MacPorts and I realized that the following cases are not working well with the current way the python wrappers are installed:

1. MacPorts. On MacPorts, one has to choose the compiler and MPI version specifying a variant (e.g. `port install plumed +clang70 +openmpi`). I could add further variants for PLUMED to include python wrappers (e.g. `port install plumed +py37`) but this would not allow installing simultaneously wrappers for different python versions (typical python packages in MacPorts are installed with something like `port install py36-numpy`). In addition, it would not allow to use MacPorts python wrappers with a custom version of PLUMED. Finally, it would mess up boost libraries, on which PLUMED depends (for `drr` module) and that in turn depend on python.
2. Pip. I think it would be nice to allow users to install python wrappers with pip, so as to use for instance virtualenvs. I don't think however that pip itself can compile the full PLUMED library.
3. Pypi. If pip works, it might be nice to upload PLUMED also on pypi, so that people can get it with `pip install plumed` (or maybe `pip install py-plumed`), without cloning our repository. I am not sure I am able to do that though.
4. Environment modules. When having environment modules on a cluster, it is inconvenient to commit to a specific python version when installing PLUMED. I think we should allow something like `module load python/3.7 py-plumed/python37 plumed/2.5.0`. Maybe even better: `module load python/3.7 plumed/2.5.0` followed by `pip3.7 install --user py-plumed`

## This pull request

In this pull request I propose a number of changes in the `python` directory of the `github.com/plumed/plumed2` repository. We have a range of options:

1. Ignore this pull request. I will just use the content of the `python` directory in the pull request to prepare a separate repository to allow for `MacPorts` (and perhaps `pypi`) installation. This is suboptimal since the repositories could go out of sync.
2. Merge this pull request and **leave** the rest of the PLUMED repo as is. This will allow both installation using `pip` and `MacPorts` and installation in the current way (wrappers installed inside PLUMED).
3. Merge this pull request and **remove** all python related stuff from `configure.ac` and `src/lib/Makefile`. Users will only be able to install PLUMED with `pip` or `MacPorts`. This is a bit extreme.

Notice that, even if we merge, it is possible that I won't be able to make a Macports package based on our master repository and I will possibly have to make a separate repository containing only the `setup.py` file and a few others. Something like [this](https://github.com/GiovanniBussi/py-plumed), probably renamed to `github.com/plumed/py-plumed` to be on the official github page. Still I would like to merge this pull request so that I will be able to update the new repository with a script (just copying the files) rather than maintaining a separate version.

My proposal is to **merge this branch on either v2.5 or master** following option 2 above. In principle, nothing should change for people using the current installation procedure, except that the PLUMED library will be loaded with dlopen.

## Detailed changes

Below a short summary of the relevant commits. See the git log for more extensive explanations.

Commit c4303330e749: Using C++ rather than C interface to better handle exceptions when mixing compilers.

Commit feb3fb9cf2e: Made python interface self-contained so that it can be installed before the plumed library.

Commit d980421ce05: This is optional, but I would merge it anyway. It adds a new constructor, that can be used as `p=plumed.Plumed(kernel="/path/to/libplumedKernel.so")`.

Commit bebec55cb09: Several improvements to the install script so that the same script can be used both to install the wrappers as in the current version and with pip/Macports. [Here](https://github.com/GiovanniBussi/py-plumed/tree/31f665b389ef6353c8ff97f3db1b597cb705ab12) and [here](https://travis-ci.org/GiovanniBussi/py-plumed/builds/457626993) you can see how it works using pip and MacPorts.


ceriottm:
The previous interface had a "precision" argument that has now been dropped. Seemed actually pretty pointless to me but wanted to ask if there's a rationale for having/not having the option.

tonigi:
Admittedly I only have a partial understanding of the problem. I am however wondering if instead of `pip` & co. one would prefer to go with a `conda` package. (The two are not mutually exclusive, but   `conda` makes `pip` maintenance effort somewhat redundant).

Pro:
* Conda packages contain pre-compiled binaries
* Packages & binaries can be built for the 3 major platforms
* Continuous integration with package deployment is provided by the `conda-forge` infrastructure
* Versions should be able to co-exist (via the standard `conda activate` mechanism)
* I *think* it will be possible to have separate (e.g. omp vs mpi) variants.

Con:
* The Conda infrastructure needs be installed. It's small and non-root, but still non-zero.
* Compilers are a "forced choice". This rules out e.g. icc-built binary packages.

There is a "proof of concept" package [https://anaconda.org/acellera/plumed2](here). It was made for other purposes but may serve as a starting point.

GiovanniBussi:
@tonigi are you speaking about the PLUMED library or about the python wrappers? Or perhaps you are suggesting to distribute both together via conda?

If it's for the whole library, that's something interesting but I would avoid to use that as a major channel, largely for the compiler problem you mentioned. Let me add that, at least as of now, PLUMED should be linked to the same instance of the MPI library as the MD code (I have a solution but it's a work in progress). It would be difficult to link PLUMED to an MD code that lives outside conda.

Giovanni

ceriottm:
Works seamlessly, it seems, and the possibility of specifying the path to the PLUMED library is a nice addition. I've really no objection as far as I'm concerned.

gtribello:
Hello @GiovanniBussi 

Sorry for taking a while to get back to you on this.   I think that it all looks fine and if the regtests are working then I am content to merge.  The only thing that I would ask is if the exceptions are working correctly?  I tried all manner of things to get the error messages passed back from PLUMED to python and really struggled to get this to work correctly.  I think this is partly why I was calling the c functions.  Can you possibly double check that you definitely get the plumed errors from python?  Maybe we can even add a regtest to double check this keeps working in the future.

Thanks
Gareth

GiovanniBussi:
@gtribello yes they should be reported correctly! There are two reasons that I know that could lead to failures in bringing the exceptions from C++ to python:

1. If you use a C++ compiler linked to a different standard library in plumed and in python. In this case, `std::exception` would be a different object in the two libraries (at least on OSX where there are separate namespaces for the linker). This should not happen with the original python wrappers, since the python wrapper was compiled with the C++ compiler chosen with `./configure`. So they should be consistent by construction. This is not true anymore (but the problem is solved in another way, see below).
2. Even if you use the same compiler, since you pass through C functions in principle the behavior of exceptions is undefined. Basically, we have a C++ function that calls a C function that calls a C++ function that throws. This is required in order to allow runtime linking, which cannot be done with C++. I observed this problem with some gcc version on OSX.

Both problems are solved in v2.5 where I implemented a special path for translating exceptions. In practice, `Plumed:cmd` calls `plumed_cmd_nothrow` that, if an exception happens within plumed, passes the information that is then rethrown within the MD code (or the python wrapper). Due to this enhancement, exceptions should be transmitted correctly now even if you have different compilers and libraries.

Notice that if you use the C interface instead (`plumed_cmd`), even if you compile as C++, the translation is not done. This was one of the reason for me to convert the python interface from C to C++.

However, if you have a case where you know things were not working and you want to try if they work correctly now we can double check

Giovanni

GiovanniBussi:
I made a few more improvements in order to allow installation on MacPorts and on Pypi without the need to create new dedicated repositories.

The MacPorts installation is already tested on Travis-CI on this branch. My plan is to submit the Portfile to the official MacPorts repository after we release version 2.5.

For Pypi I managed to add a tentative package [here](https://pypi.org/project/plumed/). This means that you can just use `pip install cython numpy` + `pip install plumed` and you are done. Version number is temporary (0.3 now) and will be synchronized with PLUMED version number after release (2.5). Also here, I plan to send the official package once 2.5 is released.

If I don't receive any other comment I will merge this branch to 2.5 at the end of the week.

gtribello:
That is great @GiovanniBussi.  Thanks for doing all this.

PR #416, opened 2018-12-06
V2.5 conda

tonigi:
##### Description

Here's a trivial change to make conda packages with minor modifications to your current build system.

##### Target release

Right now the code attempts to build packages on osx and linux. If successful, it uploads to a channel (to be configured) with a "testing" label. It would make sense to only do a "release" upload (not labeled testing) upon full releases.


conda/recipe/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# TODO: re-enable them and see. Also to do: install docs?
+
+./configure --prefix=$PREFIX --enable-shared --disable-python --disable-zlib --disable-external-lapack --disable-external-blas
+make -j4
+make install
+
+cd python
+make pip
+export plumed_default_kernel=$PREFIX/lib/libplumedKernel.so
```
GiovanniBussi: I think on MacOS this should be `$PREFIX/lib/libplumedKernel.dylib`

tonigi: Yes and I should test the import...

tonigi: Should be working now. Sorry for the commit spam. To test: `conda install -c tonigi/label/testing plumed2`

GiovanniBussi:
@tonigi I tried it on my workstation and it looks there is a problem with the path replacement done by conda.

In particular, the string containing the plumed root gets corrupted and, as a consequence, if you type `plumed -h` you do not see the script tools (such as `plumed patch`) listed.

I will check if it is possible to solve it.

tonigi:
Just an hypothesis: the `getEnvCommand()+" \""+root+"/scripts/"+` is substituted at compile time, so conda's patching of the binary (the horror) fails. If this is the case, a temporary variable may solve (e.g., plumed info --root seems to work).

GiovanniBussi:
@tonigi I think the problem is in the std::string initialized in `src/config/Config.inc.in`.

I think I solved it, I will push it tomorrow

carlocamilloni:
@GiovanniBussi @tonigi what is the status of this?

tonigi:
From my point of view, the following caveats hold
- a plumed2 conda channel should be created to hold the built packages
- release of a package should be done manually, reasonably in sync with the official releases (maybe some automation is desirable)
- packaging with xdrfile fails for unknown reasons
- there is no windows version
If they are not considered serious obstacles, it can be merged, and packages can be uploaded to conda. I can assist of course.

GiovanniBussi:
Here's the "official" channel: https://anaconda.org/plumed

Though I am not sure I know how to configure it... I guess we need a private token that should be added to the travis environment right? I am able to arrive at this page, not sure what I should do next

image

I would be happy to give "Owner" access to @tonigi as well, if Toni has the time and capability to configure it... Thanks!

tonigi:
Feel free to add me... but probably it's only useful for those who can change travis settings when needed. 

For the conda/travis bridge: long version is https://gist.github.com/zshaheen/fe76d1507839ed6fbfbccef6b9c13ed9 . Short version

- check at least both 'Allow read access to the API site' and 'Allow write access to the API site'.
- Create the token and copy it.
- in Travis settings: add an environment variable with the name CONDA_UPLOAD_TOKEN and value copied

GiovanniBussi:
@tonigi done! I added the token to travis-ci.org/plumed/plumed2

tonigi:
Ok! now i think we need a way to only run travis (or, more precisely, conda upload) when a release is needed. will write you offline

GiovanniBussi:
Testing upload [here](https://travis-ci.org/plumed/plumed2/builds/529647747).

GiovanniBussi:
@tonigi can you check if what I did in v2.5-conda on plumed/plumed2 repository is ok? The way it works now should be:
- When pushing on a branch named `test-*` or when doing a pull request or when pushing a tag, two jobs are used to build a conda binary for linux and osx. This will allow us to check if everything is ok.
- When pushing a tag, in addition, a version named like the tag is pushed on the plumed anaconda channel. Notice that dashes in tag name are replaced with underscores.
- All pushes are with `testing` label. I noticed the label can be removed a posteriori from the web interface. Maybe for now we can still do it manually. In perspective, when tag name matches our official tags (v[0-9][0-9]\.[0-9][0-9]*) we could remove the label automatically.

Is it ok to use a conda version such as `v2.5.1`? Should it be without the `v`?

Thanks!

Giovanni

GiovanniBussi:
Another doubt @plumed/all 
should the conda package be called `plumed` or `plumed2`? The difference will be in doing
````
conda install -c plumed plumed2
````
or
````
conda install -c plumed plumed
````

For consistency with MacPorts and AUR I would pick `plumed`.

carlocamilloni:
I am ok to keep consistency, I guess the version is anyway visible to the users

tonigi:
The release scheme seems just fine to me. Misc points:
 * I think conda versions like 2.x.y are ok. However, version should not have the "v".
 * Manually removing the label seems tolerable if just once per release. 
 * A potential problem with "plumed" instead of "plumed2" is the confusion arising from the various "unofficial" packages which are called "plumed2" instead. I'd vote for plumed2
* Speaking of which, there are several packages now on conda cloud. I can remove mine, and acellera can probably just copy the official one. In any case it should be evident that they are not the official channel. Not sure who "intbio" is... who seems to have made remarkable progress too

maxbonomi:
Well the code is called plumed not plumed2, so I would call it with its name

GiovanniBussi:
Great! I think I fixed the `v` in the name now. With "official" releases the version should be as expected (e.g. `2.5.2` or `2.6b`). If we push a tag with an arbitrary name then it will be with a possibly incorrect name (i.e. starting with a letter). However, it will end up in the `testing` label on anaconda, so it won't be reachable by default.

After the test finish I will merge this into v2.5. Thanks @tonigi !!

tonigi:
Very good! I found the tweak to link from @molsim 's https://github.com/intbio/plumed-conda ,  which also has mpi in conda should we want. But I'd say it's enough for now.

GiovanniBussi:
I close this since it was basically merged (though not completely)

PR #418, opened 2018-12-07
Test fix #417

### LINKED ISSUE #417: Issue with integrate_grid and openMP
@gtribello 
I think that decreasing the threshold for using openMP thread has highlighted a possible issue (threadsafety?) with integrate_grid (from plumed 2.4 on) all the regtests in plumed 2.3 go well

GiovanniBussi:
I think this fixes #417.

Regtests are passing now.


carlocamilloni:
I think we can merge it so that we have back travis working

PR #419, opened 2018-12-08
Small fix so that numerical forces in driver work with metadynamics

gtribello:
##### Description

I realized there was a bug in the routines that calculate the numerical forces that appears when you do metadynamics.  In essence as the sequence of plumed calls in driver is as follows:

prepareCalc
calc
<- store current value of bias ->
<- move atoms ->
prepareCalc
performCalcNoUpdate

you get a problem on a metadynamics update step as there is a huge difference in the value of the bias you get in the calc step and the prepareCalc step that happens after the atoms have been moved.  This difference is simply because there has been a call to update during the first call to calc as a Gaussian has been added to the metadynamics bias.

To be honest this is a pretty niche problem and I am not sure it needs fixing.  I have fixed it, however, by adding the following code after the call to calc in driver:

if(debugforces.length()>0) {
      virial.assign(9,real(0.0));
      forces.assign(3*natoms,real(0.0));
      p.cmd("prepareCalc");
      p.cmd("performCalcNoUpdate");
}

This ensures that the reference value of the bias that will be used when calculating the numerical derivatives is computed after the new Gaussian has been added.

##### Target release

It can probably be put in the latest release version.  I don't think it is that important though so we can just put it in master.

##### Type of contribution

##### Copyright

- [x ] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

##### Tests


gtribello:
Incidentally @GiovanniBussi I am currently getting some errors on this branch with travis-ci which I don't understand.

GiovanniBussi:
@gtribello it’s the problem related to #417. It’s fixed now (it doesn’t fail) thought there are parallelization problems (see Carlo’post).

I’ll chech this metad thing ASAP

GiovanniBussi:
@gtribello should we merge this?

gtribello:
@GiovanniBussi I think we might as well merge this.  So go ahead.

PR #428, opened 2018-12-18
Added integration weights for Grid class 

valsson:
##### Description

Added a function to the Grid class to get the integration weights for the grids. For multi-dimensional integration, the weights are a product of one-dimensional weights. Currently only Trapezoidal weights are supported. 

This is function that I have been using for doing integrations in the VES code. I have changed all calls in the VES code to now call the function in the Grid class. 

I also added the integration weights to the calculations of the c(t) in MetaD.cpp (#401 by @invemichele). The results of the regtest (rt67) are the same since for periodic CVs all the weights are the same. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -1942,10 +1942,11 @@ void MetaD::computeReweightingFactor()
 
   const unsigned rank=comm.Get_rank();
   const unsigned stride=comm.Get_size();
+  const std::vector<double> integration_weights = BiasGrid_->getIntegrationWeights();
   for (Grid::index_t t=rank; t<BiasGrid_->getSize(); t+=stride) {
     const double val=BiasGrid_->getValue(t);
-    Z_0+=std::exp(minusBetaF*val-big_number);
-    Z_V+=std::exp(minusBetaFplusV*val-big_number);
+    Z_0+=integration_weights[t]*std::exp(minusBetaF*val-big_number);
+    Z_V+=integration_weights[t]*std::exp(minusBetaFplusV*val-big_number);
   }
   if (stride>1) {
     comm.Sum(Z_0);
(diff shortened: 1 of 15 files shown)


GiovanniBussi:
@valsson @plumed/all this is (moderately) intrusive in the core. Should we merge it last minute to v2.5 or better merge to master?

valsson:
After private communication with @invemichele I decided to revert the change to the calculation of c(t) in MetaD.cpp such that the weights are not used in the calculation of the integrals.

We agreed that this would be better since (adapted from @invemichele email):
1. If the weights are trapezoidal, they are different from 1 only on the boundary of the grid, and the integrand, exp(-beta*F), is generally always exponentially small there. (except for periodic CVs where all weights are the same due to how the grid for periodic CVs are constructed)
2. Higher order weights, such as simpson's are apparently not always better (see https://en.wikipedia.org/wiki/Trapezoidal_rule#Periodic_and_peak_functions )
3. It introduces an overhead, that might be small or not, depending on the dimension of the grid. you create a vector as big as the grid itself every time you update c(t).

Therefore, as this introduces overhead without much improvements, it is better just to sum over the grid without taking the weights into account. 

Unless you can see a situation where point 1 is not valid, i.e. exp(-beta*F) is not exponentially small at the boundary of the grid (for non-periodic CVs).

GiovanniBussi:
@valsson does this mean that we should not merge this?

valsson:
This is something that I use a lot in the VES code and after I reverted the changes to the MetaD.cpp, it only affects the VES code. In general, I think this could be useful for other things. I would like to have it internal in Grid, it would simplify some of my code. 

So, I would like it to be merged but then only to the master I would say.

valsson:
I will close this pull request for now and look into it at a latter point.

PR #439, opened 2019-01-22
Virial minimization in EDS

whitead:
##### Description

This PR adds a new keyword to EDS, `VIRIAL`, which adds regularization to minimize the change to the system virial pressure. This PR is a work in progress; it requires doc and citation updates once my pre-print on the method is up.

@hockyg 

##### TODO

- [X] Fix Travis build
- [X] Add regtest for coverage
- [ ] Add citations and equations to documentation

##### Target release

I would like my code to appear in master

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@whitead if there are new keywords, I would rather merge this to master (v2.6). If you have bug fixes as well, could you split them as a separate PR (to merged on v2.5)?

Also consider the following issues:
- Check why the request is [failing on travis-ci](https://travis-ci.org/plumed/plumed2/builds/482702869).
- In "type of contribution" and "copyright" I think you should click the last box

Finally, I had a very quick and superficial look at the code. Notice that doing the virial with r times f only works if you do not apply PBC within the relevant action. I don't think this will stop us from merging (it's up to you), but you might want to solve the problem and/or clarify possible limitations in the documentation of EDS. Also, I might have misinterpreted something.

Giovanni

whitead:
Hi Giovanni,

Thank you for looking over this and I got the Travis problems solved. I will split into two PRs. 

Your comment about the virial term is correct. I didn't know about this issue. Could you help me understand how to fix it? According to https://www.sciencedirect.com/science/article/pii/S0009261406001552 eq 8, I believe I just need to add in the CV partials with respect to the box size? This is stored in the "virial" term of the ActionAtomstic I believe? Thanks for being so sharp on this; I have trouble with the pressure and virial mathematics.

GiovanniBussi:
Hi,

yes, I think you access it with modifyVirial.

The function `Colvar::setBoxDerivativesNoPbc` computes it in the same way you did, which is only correct if you do not use PBCs.

Functions with explicit use of PBCs (e.g. CoordinationBase) does the explicit calculation accumulating the external products of distances and derivatives

Giovanni

whitead:
@GiovanniBussi could you please check to make sure I've used the plumed virial correctly? I believe that the CV virial tensor is dcv / dv_ij where v_ij is a box vector component. Then the change in a CV wrt to change in volume would be dcv / dV = sum_ij dcv / dv_ij * dv_ij / dV = sum_ij dcv / dv_ij * (d det(V) / dv_ij)^-1. The term d det(V) / dv_ij can be computed from Jacobi's formula (adjugate matrix). For orthorombic, this becomes dcv/dv_ii * v_ii / V for the diagonal terms. [Here is how I computed this.](https://github.com/plumed/plumed2/pull/439/commits/9a4cde0ea55416f5e7ccd079800b760ce87c289b#diff-62354a01019fe4cd25360fab18bc6af7R807)
Is that logic correct? Thank you!

whitead:
This is ready to be merged from my end. @GiovanniBussi

GiovanniBussi:
@whitead great I'll merge it. There's still some problem on [travis-ci](https://travis-ci.org/plumed/plumed2/jobs/497582211) could you have a look at it? Thanks!

PR #440, opened 2019-01-27
Small test so that calculation always crashes if you try to add force…

gtribello:
…s on a value that does not have derivatives

##### Description

This is a small change to the code to fix the fact that plumed will currently run even if you try to bias a quantity that does not have derivatives.  Obviously, you would like plumed to throw an error in these cases.

##### Target release

Possibly I should have backported this onto as many previous versions of plumed as I could.  I think it is a bug.

##### Type of contribution

##### Copyright

##### Tests

It was not necessary to add any new regtests


gtribello:
@GiovanniBussi is there a smarter way that you can think of to check that this will not break.  Currently, this gives a nice error if you try to add a bias on a quantity that has no derivatives.  If you have a function of a quantity that has no derivatives, however, the error is not detected during readin.  The error might thus be a bit cryptic.

PR #444, opened 2019-02-19
Added tutorial on performance optimization.

GiovanniBussi:
I would like to add a new tutorial on optimizing simulations with PLUMED.

It is basically [this one](https://plumed.github.io/doc-cineca-2019/user-doc/html/cineca-2019.html) but renamed in a more neutral way. The tutorial is called "Optimizing PLUMED performance" and is referred to from the `Performances` page.


carlocamilloni:
Merge it

PR #447, opened 2019-02-25
Fix 446

### LINKED ISSUE #446: Bug in PLUMED 2.5 + Gromacs 2018 + OpenMP
I think there is a bug that shows up when GROMACS is invoked without specifying the number of OpenMP threads.

In particular, if I do not use the `-ntomp` flag, the GROMACS default is "0 threads", what means choosing the number of threads at runtime. However, PLUMED is then told to use zero threads and this leads to incorrect evaluation of some CVs (I found this with ALPHARMSD, but I suspect all multicolvars could be affected). Other CVs appear not to be affected (e.g. COORDINATION).

I will open a pull request with a tentative solution.

GiovanniBussi:
##### Description

This change makes it so that when `cmd("setNumOMPthreads",&x);` is invoked with `x<=0`, the number of threads is set to 1. This should fix #446, at least in the fact that we will not have erroneous results.

Notice however that the optimal fix would be to find out the correct number of threads chosen by GROMACS. However, I am not sure that at the time we initialize PLUMED this number has been already chosen by GROMACS, not I don't know how to find its value. Currently, we call `plumed_cmd(plumedmain,"setNumOMPthreads",&hw_opt.nthreads_omp);`. Maybe @carlocamilloni knows how to better solve this? Otherwise, this fix might be sufficient.

##### Target release

### DIFF
--- a/patches/gromacs-2018.5.diff/src/programs/mdrun/runner.cpp
+++ b/patches/gromacs-2018.5.diff/src/programs/mdrun/runner.cpp
@@ -1342,7 +1342,10 @@ int Mdrunner::mdrunner()
           /* detect plumed API version */
           int pversion=0;
           plumed_cmd(plumedmain,"getApiVersion",&pversion);
-          if(pversion>5) plumed_cmd(plumedmain,"setNumOMPthreads",&hw_opt.nthreads_omp);
+          if(pversion>5) {
+             int nth = gmx_omp_nthreads_get(emntDefault);
+             if(pversion>5) plumed_cmd(plumedmain,"setNumOMPthreads",&nth);
+          }
         }
         /* END PLUMED */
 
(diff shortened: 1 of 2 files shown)


carlocamilloni:
this is weird because as far as I understand at that point gromacs should have performed all the setup actions.

carlocamilloni:
@GiovanniBussi could you test this:

          int nth = gmx_omp_nthreads_get(emntDefault);
          if(pversion>5) plumed_cmd(plumedmain,"setNumOMPthreads",&nth);

in place of the current implementation?

GiovanniBussi:
@carlocamilloni it seems to work! Number of threads is now correctly detected both when using `-ntomp`, `OMP_NUM_THREADS`, and when gromacs picks it automatically.

I will merge **both** changes anyway, so that if someone forgets to update at least there are no errors in the result.

Thanks!

PR #448, opened 2019-03-21
Atom selection with mdtraj/mdanalysis

GiovanniBussi:
With some help from @tonigi I managed to implement advanced atom selections. The idea is to reuse existing packages (mdtraj and MDAnalysis), so that people can use a syntax that they know already. We checked if it's possible to incorporate also VMD language but apparently it isn't easy to do.

This is not a replacement of our `@phi-1,@chi-3` selectors, which are probably still the best choice for selecting specific angles. However, it provides a more general way to do large selections. Below I wrote something that I will paste in the manual.

There are some problems now on Travis-CI but I suspect they are related to Travis itself.

I plan to merge this on **master branch (v2.6)**. Let me know if you have comments on the syntax!

Giovanni

### Advanced atom selection with mdtraj or MDAnalysis

Since PLUMED 2.6 it is possible to use the expressive selection syntax of [mdtraj](http://mdtraj.org/latest/atom_selection.html) and/or [MDAnalysis](https://www.mdanalysis.org/docs/documentation_pages/selections.html):

````
MOLINFO STRUCTURE=helix.pdb PYTHON_BIN=python
g1: GROUP ATOMS=@mda:backbone 
g2: GROUP ATOMS={@mda:{resnum 1 or resid 3:5}}
g3: GROUP ATOMS={@mda:{resid 3:5} @mda:{resnum 1}}
g4: GROUP ATOMS={@mdt:{protein and (backbone or resname ALA)}}
g5: GROUP ATOMS={@mdt:{mass 5.5 to 20}} # masses guessed by mdtraj based on atom type!
g6: GROUP ATOMS={@mda:{resid 3:5} @mda:{resnum 1} 1-10}
````

Here `@mda:` indicates that `MDAnalysis` language is used, whereas `@mdt:` indicates that `mdtraj` language is used. Notice that these languages typically select atoms in order. If you want to specify a different order, you can chain definitions as in `g3` above (compare with `g2`). Selections can be also chained with standard PLUMED selections (see `g6`).

The double braces are required due to the way PLUMED parses atom lists. In particular:

- The outer braces are needed to show PLUMED where the `ATOMS=...` option ends.
- The inner braces are needed to show PLUMED where each selector ends.

MDAnalysis also supports geometric selectors based on atomic coordinates. These selectors **are static** and return lists computed using the coordinates stored in the `MOLINFO` pdb file.

In order to use this syntax you should check the following points at runtime:

1. `plumed --no-mpi config has subprocess` prints `subprocess on` (should be ok on most UNIX systems).
2. You have a python interpreter with mdtraj and/or MDAnalysis installed. You can check using:
   - `python -c "import mdtraj"`
   - `python -c "import MDAnalysis"`

   In order to install these packages refer to their documentation. Pip or conda install should be ok, provided you make sure the correct python interpreter is in the execution PATH at runtime. Notice that you will only need the package(s) related to the syntax that you want to use.
3. In case you installed these modules on a python with a different name (e.g. `python3.6`), the correct check is:
   - `python3.6 -c "import mdtraj"`
   - `python3.6 -c "import MDAnalysis"`

   If this is the case, you should use `MOLINFO PYTHON_BIN=default` at runtime and set the environment variable `export PYTHON_BIN=python3.6` or set `export PLUMED_PYTHON_BIN=python3.6` (higher priority). Alternatively, directly provide the interpreter with `MOLINFO PYTHON_BIN=python3.6` (even higher priority).
4. The PDB file that you provide to `MOLINFO` should have consecutive atom numbers starting from 1. This is currently enforced since reading atom numbers out of order (as PLUMED does) is not supported by other packages.

### VMD syntax

You can also use `@vmd:` to access to VMD syntax. This will require launching an instance of VMD, so `vmd` executable should be in the path.


tonigi:
HI, a couple of afterthoughts/clarifications on VMD:

1. launching VMD for each string to convert is possible, but slow given VMD startup times
2. "interrogating" it in real time as a subprocess is a headache
3. VMD's python modules are not compiled-in in recent builds.
4. as an alternative, one could collect all atomselection strings in the input, convert all of them in a single VMD run, and then substitute them. That, however, requires that (AFAIU) plumed input parsing occurs in two stages, which is not the case.

All in all, may look like a questionable effort in presence of MDT and MDA, which are somewhat comparable.

GiovanniBussi:
@tonigi option 4 I think is not compatible with the way we handle the input now.

In particular, we also allow MD codes to "push" input lines to PLUMED one by one. This is for instance done by OpenMM. I would be skeptical in changing this "interpreted" behavior (one line at a time) to a "compiled" behavior (where the full input is parsed first), if this is just to allow VMD selections.

Maybe option 1 could be implemented (as `@vmd:`), with a warning on the speed issue. This is a problem only if many selections are done in the same input file I think.

GiovanniBussi:
@tonigi I implemented the `@vmd` option.

It currently goes through python (see [here](https://github.com/GiovanniBussi/plumed2/commit/f5164b91f3c39b374aa84056d3e88d8137bf1082)). In order to empty the output buffer it is sufficient to add a number of `puts` commands. It seems to work, and launches VMD once, so the overhead is not dramatic.

Before merging I plan to change the implementation in the following way:
- launch a separate python interpreter for mdtraj and/or MDAnalysis, only when the user uses the related syntax.
- launch VMD directly from C++, so that python interpreter is not needed if using VMD syntax only.
- kill the interpreters at the first MD step. They can easily be launched again later if one pushes a new line.

In case the buffering solution with VMD is not portable, we could either drop VMD support or reimplement the communication with sockets (should better support unbuffered communication).

In the future, it could be easy adding [gromacs syntax](http://manual.gromacs.org/documentation/2018/onlinehelp/selections.html?highlight=select) as well.

However, the code it its current form should be working.

carlocamilloni:
It sound cool, I would not care about vmd syntax, I agree that currently is the only good software for trajectories but will it last? I think no... is already lagging behind all other tools in a number of aspects

Issue #435, opened 2019-01-14
Consistency in "-" and "_" in action and option names


GiovanniBussi:
I noticed that in action and option names we use a mixture of underscore (_) and dash (-), the latter being used rarely.

You can see all the uses with the command:
````
grep "^let b:plumedDictionary" $plumedroot/vim/syntax/plumed.vim | grep "\-"
````
They are:
````
MULTI-RMSD
PCARMSD SQUARED-ROOT
PROPERTYMAP DEBUG-CLOSE LOG-CLOSE
````

I would propose to switch to underscore only converting them to
````
MULTI_RMSD
PCARMSD SQUARED_ROOT
PROPERTYMAP DEBUG_CLOSE LOG_CLOSE
````
In the transient (or forever), we can allow the dash version as an alias (depending on how much are these keywords used).

PR #449, opened 2019-03-21
Fix underscores

GiovanniBussi:
This change adds aliases using "_" instead of "-" (see #435).

I would like to merge this on branch 2.4 before we release v2.4.5.

I think using only characters in `[A-Za-z0-9_]` has some advantage, since it means that all actions/keywords can be valid C++/Python identifiers. In the future we might decide to disable names not matching this pattern. Merging this change now would allow for better backward compatibility.

Please let me know if there is any objection to this merge. Thanks!

@gtribello @jpazurikova


carlocamilloni:
Good to me

carlocamilloni:
Actually in 2.6 I would remove the - version

PR #471, opened 2019-05-01
fixed derivatives in CoordinationNumbers

MoleOrbitalHybridAnalyst:
##### Description

A bug fix on multicolvar::CoordinationNumbers::compute(). The derivatives were missing two terms. The version 2.4 seems to be the oldest version this bug appears.

##### Target release

I would like my code to appear in release v2.4 and later

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@plumed/all in case @gtribello approves this, should we merge it to 2.4 even though it is not maintained?

carlocamilloni:
I think so, I mean patch related to correctness are very relevant and version 2.4 is still very recent so we could even think of making a new release if some other patches are worth porting

whitead:
Thank you @MoleOrbitalHybridAnalyst for spotting this and contributing the correction.

PR #474, opened 2019-05-09
Multicanonical and multithermal-multibaric simulations

PabloPiaggi:
##### Description

I am submitting 2 new target distributions for the VES module and **also** an action to calculate weights for reweighting.

The target distributions allow performing multicanonical and multithermal-multibaric simulations[1][2]. These simulations can be used to reweight at all temperatures and/or pressures in the chosen intervals. For this reason also a new reweighting action is needed. 

[[1]](https://journals.aps.org/prl/abstract/10.1103/PhysRevLett.122.050601) Piaggi and Parrinello, Phys. Rev. Lett. 122 (5), 050601 (2019)
[[2]](https://arxiv.org/abs/1904.05624) Piaggi and Parrinello, arXiv preprint arXiv:1904.05624 (2019)
 
##### Target release

I would like my code to appear in **any** release. As soon as possible in the master branch.

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@PabloPiaggi thanks for your request! I have a comment on the reweighting part. Now you use something like
````
REWEIGHT_TEMP_PRESS ARG=energy,volume
````
I think this is quite error prone. Couldn't it be
````
REWEIGHT_TEMP_PRESS ENERGY=energy VOLUME=volume
````
?

Even better, if you want to limit people to only use the real energy and the real volume, we could search for a way to find these values implicitly.

PabloPiaggi:
Thanks @GiovanniBussi. Yes, you are right, the syntax is much clearer in that way. I will make the changes tomorrow.

Perhaps the potential energy and the volume can be chosen as defaults but more complicated options can also be allowed. I will look into this. Thanks again!

PabloPiaggi:
I was implementing @GiovanniBussi 's suggestion and it came to my mind that it could be useful to discuss first what to do with REWEIGHT_TEMP. This reweighting action has the limitation that automatically takes as energy the potential energy + the bias potentials. Not only does this give no flexibility to the user but it also outputs weights that take from the *biased* ensemble at temperature T to the *biased* ensemble at temperature T'. In my experience one wants to obtain weights that take to the *unbiased* ensemble.

I propose two alternatives to the multitemperature-multipressure reweighting.
1) Have a REWEIGHT_TEMP action that calculates weights:
<a href="https://www.codecogs.com/eqnedit.php?latex=e^{(\beta-\beta')E}" target="_blank"></a>
and a REWEIGHT_TEMP_PRESS action that calculates weights:
<a href="https://www.codecogs.com/eqnedit.php?latex=e^{(\beta-\beta')E&space;&plus;&space;(\beta&space;P-\beta'&space;P')\mathcal{V}}" target="_blank"></a>
In this case to go from T to T' one uses REWEIGHT_TEMP and to go from T and P to T' and P' one uses REWEIGHT_TEMP_PRESS. Choosing P=P' makes REWEIGHT_TEMP_PRESS equal to REWEIGHT_TEMP. This is the path that we were taking now.

2) Have a REWEIGHT_TEMP action that calculates weights:
<a href="https://www.codecogs.com/eqnedit.php?latex=e^{(\beta-\beta')E}" target="_blank"></a>
and a REWEIGHT_PRESS action that calculates weights:
<a href="https://www.codecogs.com/eqnedit.php?latex=e^{(\beta&space;P-\beta'&space;P')\mathcal{V}}" target="_blank"></a>
In this case to go from T to T' one uses REWEIGHT_TEMP and to go from T and P to T' and P' one uses REWEIGHT_PRESS in combination with REWEIGHT_TEMP.


In all cases one should add weights through REWEIGHT_BIAS to go from the biased ensemble to the unbiased one. I have already provided examples that show how to do this.

What do you think?

PabloPiaggi:
Just a reminder. Could you please give the green light to program option 1 or 2?

Thank you.

carlocamilloni:
actually I would prefer to have a single action REWEIGHT_TEMP_PRESS that supersedes REWEIGHT_TEMP, alternatively I would go for option 2 @plumed/all

PabloPiaggi:
I implemented the suggestions of @GiovanniBussi and @carlocamilloni . Please let me know if there's anything else that needs to be changed. It was difficult to write the documentation for REWEIGHT_TEMP_PRESS because now it handles 4 different scenarios. I hope you like the way I wrote it.

@valsson could you please take a look at the new target distributions?

Thanks!

carlocamilloni:
I think there may still be a precision issue in one test, check if it is enough to decrease the precision by one or two digit

PabloPiaggi:
Thanks @carlocamilloni . Now it passed all checks

carlocamilloni:
Great, thanks!

Issue #478, opened 2019-05-21
add DL_POLY_4 HISTORY support for driver


alinelena:
would be nice the plumed driver to be able to read DL_POLY_4 HISTORY file

alinelena:
at the moment the implementation add --idlp4

timestep, masses and charges are read from trajectory file

GiovanniBussi:
Closed, see #481

PR #479, opened 2019-05-21
add support for Dl_POLY_4 trajectory in driver

alinelena:
##### Description

adds support to the driver to read dl_poly_4 history files, see #478 

##### Type of contribution

##### Copyright

##### Tests

 - to be added


GiovanniBussi:
Thanks! Could you add a regression test as well? Let us know if you need help in setting it up.

alinelena:
Ciao Giovanni,

no problem, is there any documentation on the workflow you use for reg testing?

Regards,
Alin

alinelena:
> Thanks! Could you add a regression test as well? Let us know if you need help in setting it up.

Regtest added. unfortunately I get two failures.
1. is related to astyle, so i suspect my formatting is not in line with the expected. Unfortunately no explicit error is shown can you help to put me on the right track?
2. fails to build some conda, that is odd.

GiovanniBussi:
@alinelena thanks for implementing the test!
- about the conda failure don't worry. I know what it is, I thought I fixed it already but apparently it didn't work. I can merge your PR anyway.
- about astyle you should just type `make astyle` in the main directory then `git commit`

I noticed that your driver flag also reads charges and masses. I would suggest to use the following in the test:
````
DUMPATOMS ATOMS=1-864 FILE=test.xyz
DUMPMASSCHARGE FILE=mcfile
````
Then run the following commands:
````
cd regtest/basic/rt-dlp4
make
cp tmp/test.xyz test.xyz.reference
cp tmp/mcfile mcfile.reference
git add .
git commit
````
(reference files are checked for differences when running tests).

Also if possible could you decrease the number of atoms? The HISTORY file is > 2Mb, and as the project grows we would like the repository size not to grow too much. Perhaps 30 atoms and a few frames are sufficient to make sure the file is correctly read.

Thanks again!

GiovanniBussi:
@alinelena I think I solved the conda issue on master branch, so your next commit on the pull request should not fail anymore.

alinelena:
@GiovanniBussi ok Giovanni all points taken I will make clean PR and close this one.
unrelated question. dlpoly trajectory may contain forces too. does it make any sense to read them if available?

alinelena:
this pr is superseeded by #481

GiovanniBussi:
@alinelena about forces I am not sure... Certainly the driver could pass forces and one could see that using ENERGY CV they get scaled. If you want you can put them in the forces array (I think now they are set to zero), I don't expect this would create problems anywhere else. However, I am not sure which would be a practical application.

Issue #467, opened 2019-04-29
Problems with Grid/SparseGrid hierarchy


GiovanniBussi:
I think there is a problem in the Grid/SparseGrid hierarchy. Now we have
````
Grid -> SparseGrid
````
which means that every SparseGrid is also a Grid. However, Grid contains member `grid_`, whereas SparseGrid contains member `map_` and does not use member `grid_`. This can lead to problems. Indeed, we added a lot of methods to `Grid` that use `grid_` and that thus, when used from a `SparseGrid`, lead to unpredictable behavior. I want to change the structure as follows:
````
GridBase -> Grid
         -> SparseGrid
````
where GridBase only declares virtual methods that are available both in Grid and SparseGrid. In this way, in MetaD we can safely use pointers to `GridBase` using the available methods and allowing for sparse grids. Other methods not interested in SparseGrid might directly use Grid, which will be always dense.

I noticed a single bug induced by this, namely in `METAD` with `TARGET` and `GRID_SPARSE` (due to the fact that `getMaxValue` is not overridden in `SparseGrid`). I will make an ad hoc fix for v2.5, but I think that on master we should change the hierarchy as I proposed above.

@invemichele @valsson I will need to make tiny adjustments to ves, which is the only module using grid. I will open a pull request so that you will be able to cross check them.

carlocamilloni:
This sounds dangerous, maybe we should remove sparse grids if they are not used. 
Anyway your solution sounds good

Carlo

Sent from my iPhone

GiovanniBussi:
I think they are useful with MetaD in multiple dimensions, so I'll try to fix it (I have a draft already)

PR #483, opened 2019-05-27
Fix 467

GiovanniBussi:
##### Description

This pull request addresses #467. Since this affects code in external modules, I would like @valsson and @invemichele to have a look before I merge.

Notice that the problem with grids lead to a bug in `METAD` + `TARGET` + `GRID_SPARSE` that I addressed separately in v2.4 (branch v2.4-fix-467) and v2.5 (branch v2.5-fix-467). I will also merge those branch once this is merged.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
This pull request **fixes 2 alerts** when merging fc3fa0bcc71f90d1b98a5faf4c518a87921d8366 into 9629d2dbe6e557f14011641fc29d4d7a1aec1c07 - [view on LGTM.com](https://lgtm.com/projects/g/plumed/plumed2/rev/pr-33988e7f1cc238c2e10e90e39265a3c2b7719db6)

**fixed alerts:**

* 2 for Virtual call from constructor or destructor

---

*Comment posted by [LGTM.com](https://lgtm.com)*

invemichele:
@GiovanniBussi the new implementation seems a good idea to me. 
I thought my VESdeltaF class was already ok, since I force the user to use non-sparse grids (line 224) to avoid any issue. But I guess with the lines you added it will remain safe even if something will change in the future... so it's fine for me.

In the class bias/External, line 142, I guess it should be `GridBase::create` instead of  `Grid::create`, even though probably there is no practical difference between the two.

GiovanniBussi:
This pull request **fixes 2 alerts** when merging 68f4e7abf899da203826de4f39d03b1d3d2e71dc into 9629d2dbe6e557f14011641fc29d4d7a1aec1c07 - [view on LGTM.com](https://lgtm.com/projects/g/plumed/plumed2/rev/pr-cbb7486041772fbbb5b7823c0368c473b11b624b)

**fixed alerts:**

* 2 for Virtual call from constructor or destructor

---

*Comment posted by [LGTM.com](https://lgtm.com)*

valsson:
Hello @GiovanniBussi the changes to VES look fine to me so as far I am concerned it can be merged. 

Maybe one thing is that I would never intend the user to use sparse grids in the VES code, and the user never has the option to select a sparse grid. Perhaps the only place where this could be an issue is in TD_GRID where the user loads an external grid file, there I should make it such that it is not possible to use a sparse grid.

Another thing is that my PhD student @bpampel has done some additions to the Grid class by creating a VesGrid class which derives from Grid and implements various additional features, mostly for doing linear interpolation that he needs. We think that would be nicer to have this natively in the Grid class, and also include some other grid features that we use in the VES code and are now included in the VES code. We could clean this up and make a PR and then you can see if it is fine for you to include it in the main code. We can also discuss this better in person. 

Omar

GiovanniBussi:
@valsson thanks!

I see that all the calls to Grid::create in VES indeed set sparse to false.

Maybe the GridBase pointers in VES should all be converted to Grid pointers and used as I did for @invemichele 's stuff?

In the meanwhile I will merge this version, then of course you might be willing to change later the way it is used in VES.

About the additional stuff, is it something that would only make sense for dense grid? If so you can inherit from Grid. Otherwise, I think it would be better to inherit from the base class.

invemichele:
@GiovanniBussi I just noticed that in the code there are still many references to `Grid::index_t`, that probably should be changed to `GridBase::index_t`

PR #503, opened 2019-07-28
Caseinsens

carlocamilloni:
##### Description

Make the parsing of the input case insensitive by
- forcing directives to be converted in uppercase
- using interpretLabel to convert in uppercase directives
- implement a case insensitive comparison for parsing tools

in this way I can overcome some problem with const Action in ActionRegister and avoid code duplication I think

##### Target release

v2.6

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/regtest/basic/rt5/plumed.dat
+++ b/regtest/basic/rt5/plumed.dat
@@ -1,5 +1,6 @@
-ang1: ANGLE ATOMS=2,6,3
-ang2: ANGLE ATOMS=2,6,7,3
+angle atoms=2,6,3 label=ang1
+ang2: AnglE ATOMS=2,6,7,3
+anglE ATOMS=2,6,7,3
 
 RESTRAINT ...
   ARG=ang2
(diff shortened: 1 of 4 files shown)


GiovanniBussi:
Neat solution!

Do you know if there are actual actions that use small letters? I checked with
````
git grep "+PLUMEDOC"| awk '{print $NF}' | grep "[a-z]"
````

and I see only command like tools or other pages.

Maybe it would be better to give an error instead of translating to upper case?

The reason for this is that I am afraid there could be problems if someone *thinks* he is registering a lower case name (and use it for `PLUMEDOC`).

I will also have to take care about vim syntax, but this can be done later.

carlocamilloni:
I agree we can force an error, i will commit the change and then if we are happy i’ll
Merge it

carlocamilloni:
@GiovanniBussi it now throw an error in case of small letter actions, currently there are not

PR #504, opened 2019-07-29
Update travis build to use Ubuntu 16 xenial

GiovanniBussi:
##### Description

Update travis build to use Ubuntu 16 xenial

##### Target release

##### Notes

I made some fix already, namely:
- changed a check on autoconf since ubuntu 16 has a custom autoconf
- install by hand boost libraries when compiling in debug mode, since linking a code with debug flags with non-debug boost gives problems.

There is however one thing that I cannot fix. Regtest `mapping/rt-adapt` fails. I am not sure if this is a real bug or just some numerical instability, but it would be great if @gtribello can have a look at it.

Thanks!

Giovanni

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -129,7 +129,14 @@ install:
     fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then sudo apt-get update -qq ; fi
 # install some package - these are fast, we install them anyway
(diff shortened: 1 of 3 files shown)


GiovanniBussi:
I close this since considering #513 I will try to directly jump to ubuntu 18 bionic.

PR #509, opened 2019-07-31
Override

GiovanniBussi:
##### Description

This is a massive commit, but it does something simple: it just adds `override` to all the functions where it makes sense. This makes the code easier to read, since you can immediately see when a function is overriding a virtual function in a base class. I might have missed some (and I didn't touch contributed modules), but there are no false positives since the compiler would give an error.

##### Target release


carlocamilloni:
good to me!

PR #510, opened 2019-08-01
Fix for updating of bias potential in VES_LINEAR_EXPANSION

valsson:
##### Description

This fixes an issue with updating of bias potential in VES_LINEAR_EXPANSION that is present for certain integrators that call the calculation of the bias multiple times and replica exchange. 

See discussion on PLUMED group: 
https://groups.google.com/d/msg/plumed-users/kPZu_tNZtgk/LrkS0EqrCQAJ

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/CHANGES/v2.5.md
+++ b/CHANGES/v2.5.md
@@ -198,4 +198,4 @@ For users:
   - Force using cython when compiling from source. Still using the pre-generated cpp file
     when installing from PyPI, to avoid cython dependency.
   - Using python 2 to create the cpp file uploaded on PyPI (this will change to python 3 in 2.6, see \issue{502}).
-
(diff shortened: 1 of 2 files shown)


valsson:
When testing these things I noticed that it possible to call update() multiple times within the same step which most likely  will lead to incorrect results. I don't think anyone would do something like that, but anyway I add a warning for this function.

valsson:
@invemichele, you should look into this, briefly looking at your VesDeltaF it seems that it also suffers from the same issue that the averages needed for the gradient are accumulated in calculate(), and not update(). This will create problems with certain integrators (see [here](https://groups.google.com/d/msg/plumed-users/kPZu_tNZtgk/LrkS0EqrCQAJ)), like the one used in i-Pi, and replica exchange.

invemichele:
Yes, my code has the same issue. I will fix it as you did, storing the values in a pointer and throwing a warning if update() is called before a new calculate().
thanks

GiovanniBussi:
@valsson you would like to merge this to 2.5 as a bug fix, however the pull request points to master. Can you change it? Thanks!

valsson:
Sorry about that, I have changed the PR to v2.5

PR #511, opened 2019-08-01
V2.5 fix h36

GiovanniBussi:
##### Description

Make sure that h36 encoding is used when writing atom numbers in PDBs.

I would like to merge this but I am afraid since this part of the code is not covered (see [here](https://www.plumed.org/coverage-master/coverage/tools/PDB.cpp.gcov.html)).

@gtribello could you have a look? It would be ideal if you add a regtest of this, but if you think it is correct we can merge anyway.

##### Target release

I would like my code to appear in release __v2.5__


gtribello:
ciao @GiovanniBussi 

I am not sure what you want me to look at here.  I looked at the PDB.cpp code coverage report.  The following functions are not covered:

setPositions, hasFlag, getBoxAxs, getBoxAng, getAtomsInChain, checkForResidue, checkForAtom

with the exception of hasFlag none of these functions are used anywhere in the code.  My view is thus that if you want to delete them that is fine with me.  

As for hasFlag - this has to stay.  It is doing something relatively simple though so I am pretty confident that you will not have broken it.

GiovanniBussi:
The part that is not tested is the else at line 543 (I think it is the case without MOLINFO). Can you confirm that my edit is correct and/or change a test to cover those lines? Thanks! Giovanni

gtribello:
Hello @GiovanniBussi 

I changed the test in 

regtest/dimred/rt-pca

so that it should test this line in the pdb input file.  I made the changes to the test on the v2.5 branch so you should be able to test the code that is in this branch.

Let me know if you need anything else.

PR #514, opened 2019-08-13
Override for VES modules

valsson:
##### Description

Adds `override` statement to virtual functions in the VES modules. Related to PR #509. 

##### Target release

I would like my code to appear in release v2.6

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
I’m happy to merge this as is. However, notice that it is recommended not to use both virtual and override (https://stackoverflow.com/questions/39932391/virtual-override-or-both-c). Let me know if you want to change this or if I should merge this PR as is. Thanks! Giovanni

valsson:
Thanks for letting me know @GiovanniBussi. This should be fixed in the latest commit such that virtual is only used in base classes while override is used in derived classes. 

When looking at the code, I noticed that there is one file (adjmat/ActionWithInputMatrix.h) where both virtual and override is used.

PR #520, opened 2019-09-04
Backport fix 441 to 2.4

GiovanniBussi:
I tried to backport the fix and it looks that tests are passing (but I will wait for this PR to be fully tested on travis-ci).

Since this is a bug fix, we might also release a new v2.4.7 together with v2.5.3


carlocamilloni:
yes I agree we may release a 2.4.7 or maybe, since it will be for a single bug, just wait a bit longer and release 2.4.7 at the end of the years (irrespectively of wether we fix more things or not)

PR #528, opened 2019-10-18
PYCV for PLUMED v2.6

tonigi:
This pull request is mostly for testing on Plumed's travis and setting a "known good mark". No need to merge (but feel free to). 

 Main reasons *against* merging would be:
* It required some (minor) changes to travis and configure.  
* Python linking is fragile; it may complicate maintainability later on.
* A proper implementation would probably mirror the C++ API + provide a wrapper.
* Possible interference with the other Python functionality.
* Stuff not yet apparent.

##### Description

PYCV enables PLUMED2 Collective Variables (CVs) and arbitrary functions to be defined and auto-differentiated in the Python language. [See also here.](https://giorginolab.github.io/plumed2-pycv/)

##### Target release

Travis-checks against v2.6 (but should work in v2.5.2 straight away).

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
I agree this should not be merged as is. I already discussed with @tonigi several of the points above. Anyway, this is certainly an excellent starting point so I would keep it here as a work in progress until we make a real plan for python integration.

PR #546, opened 2020-01-23
Some fixes to VES optimizer file output

valsson:
##### Description

This PR fixes two issues with file output in the VES optimizer: 

- Made sure that all relevant output files are written out at the final step when shutting down the simulation. This solves issues reported by @PabloPiaggi with restarting when there is a mismatch been the output of files and the number of MD steps.

- Fixed an off-by-one bug in the output of target distribution averages. The bug only affects the output and does not affect results. The bug also affected the output of coefficients when using a bias cutoff. This is a minor issue that should not affect users. 

I have done the changes to v2.4 as discussed with @GiovanniBussi, but they should be merged to v2.5, v2.6, and the master

##### Target release

I would like my code to appear in release v2.4, v2.5, v2.6, master

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@valsson you opened by mistake a pull request to master branch.

I'll take care of it. I am running right now tests for your branch separately [here](https://travis-ci.org/plumed/plumed2/builds/640998562). I'll be back if I see failures or conflicts in merging

Thanks!

GiovanniBussi:
@valsson it looks like there are failures on travis ci. Can you double check? [this is the relevant build](https://travis-ci.org/plumed/plumed2/builds/640998562)

valsson:
@GiovanniBussi sorry for mistake in doing PR to the master instead of v2.4, I didn't pay enough attention. Thanks for taking care of it! 

I looked at the failing regtest, it is numerics issue with one of the grid files, the last digit in one line is -1. In principle, I should have a format option for these files to avoid these kind of issues, but it is not implemented at the moment (I might do it on the master soon), and I haven't observed before that I was getting different results on my Mac and on travis. To fix it, I just changed the reference file by hand. I will take a look at the build later to see if that fixes it.

valsson:
Okay, then the issue is that it fails for the MacOS test. I have just removed the reference file for the failing file, so the failure is avoided. Anyway, this particular file is not an important part of this regtest, so it should not matter if it is not checked.

PR #558, opened 2020-02-27
small fix for single precision timestep

invemichele:
##### Description

A possible fix for the small annoyance of non rounded time step if MD engine uses single precision (see https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/plumed-users/qv-rqQvdODE/aSfcMM_dCAAJ ).
I implemented this for myself long time back, now I saw the discussion on plumed-users, so I decided to make this PR. Feel free to ignore it if you think is not useful.

This is the rationale: the MD time step is set by the user in some input file in decimal notation, and it is extremely unlikely that the user needs more than 6 significant digits. The MD code then stores it in single or double precision. Plumed might want to stay close to the original decimal time step, instead of simply casting it to double, especially since it is only used when outputting the time, and does not play a role in calculations.

##### Target release

I would like my code to appear in release 2.7 (maybe also in older versions?)

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/Atoms.cpp
+++ b/src/core/Atoms.cpp
@@ -442,6 +442,11 @@ void Atoms::updateUnits() {
 
 void Atoms::setTimeStep(void*p) {
   MD2double(p,timestep);
+// The following is to avoid extra digits in case the MD code uses floats
+// e.g.: float f=0.002 when converted to double becomes 0.002000000094995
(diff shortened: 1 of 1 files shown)


carlocamilloni:
thanks!

PR #573, opened 2020-05-27
Removed amber18 patch

GiovanniBussi:
##### Description

@dvdesolve I would remove amber18 patch from the plumed source code starting with version 2.7, if you agree.

##### Target release


dvdesolve:
I think this is ok. in any case we always could rollback to this commit

Issue #567, opened 2020-05-04
Fix tail corrections in gromacs?


GiovanniBussi:
It looks like the energy communicated by gromacs does not include long tail corrections. Maybe this should be fixed if possible?

See [mailing list](https://groups.google.com/d/msgid/plumed-users/CAGBhpicEGTq%3DATN9P3%3DY05u4u51hyYLsY5A8mnA3eC%3DkhqO5Ew%40mail.gmail.com?utm_medium=email&utm_source=footer).

invemichele:
I learned this from @PabloPiaggi who probably has suggestions on the matter

PabloPiaggi:
I thought that the lack of tail corrections in the interface with Gromacs was intentional. The tail corrections are special because they have a contribution to the virial but no contribution to the forces. I have a vague memory that I tested in LAMMPS biasing the energy with tail corrections and didn't succeed in passing the sanity checks. As a matter of fact, I added a [warning](https://github.com/lammps/lammps/blob/master/src/USER-PLUMED/fix_plumed.cpp#L438) in LAMMPS if one runs a constant pressure simulation and uses a potential that includes tail corrections. For this reason I have always biased the energy without tail corrections and then used the full energy to do reweighting.

However, I wrote the equations again today and I don't see any reasons why it shouldn't be possible to fix this. It seems that it would suffice to add the tail corrections to the energy and the appropriate term to the virial. 

It would be good to check the interface with LAMMPS to see if tail corrections are handled properly. I think we were only testing an example without tail corrections [here](https://github.com/plumed/test-lammps-and-plumed/tree/master/rt-engforce-npt).

invemichele:
Should we add a warning about this bug in the ENERGY manual page?

From what I understand, the plumed ENERGY is (slightly) wrong when in gromacs `DispCorr Ener` or `DispCorr EnerPres` is used, and in lammps when `pair_modify tail yes` is used.
And a similar issue is probably present also in other MD codes.

biradicaloid:
It makes me confused that the ENERGY in plumed is exactly same with the potential energy in lammps when the "pair modify tail yes" is used.  As expected, ENERGY in plumed should equal to (Potential energy - Etail) in lammps. Is it correct to use the current ENERGY (including Etail) to do biasing and reweighting?

PR #577, opened 2020-06-04
Update doc of ENERGY CV

PabloPiaggi:
The documentation of the ENERGY CV states that it calculates the total energy. For most people total energy means potential+kinetic energy. I think this should be clarified and that there is an implicit convention, followed at least by LAMMPS and Gromacs, that the energy CV is the **total potential energy** (without tail corrections as discussed in issue #567). I also believe that most if not all papers that use some energy as CV use the potential energy. Furthermore, the two citations in the ENERGY CV are papers that use the potential energy.

For the reasons stated above this PR proposes to change **total energy**  to **total potential energy** in the documentation of the ENERGY CV.

Please disregard this PR if the plan is to use the total energy in the future.

This issue was pointed out to me by Tom Gartner @tgartner .

##### Target release

I would like my code to appear in any release

##### Type of contribution

##### Copyright

##### Tests

- [x ] No regtests required

### DIFF
--- a/src/colvar/Energy.cpp
+++ b/src/colvar/Energy.cpp
@@ -32,7 +32,7 @@ namespace colvar {
 
 //+PLUMEDOC COLVAR ENERGY
 /*
-Calculate the total energy of the simulation box.
+Calculate the total potential energy of the simulation box.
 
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
Very good point @PabloPiaggi ....

I would actually backport it so that it appears also in older copies of the manual

PR #586, opened 2020-06-19
Bias reweighting for EMMI

tlhr:
##### Description

This patch adds the possibility of using the bias potential from a replica-based enhanced sampling protocol like metadynamics to be used to re-weight the EMMI ensemble average.

@carlocamilloni can you have a look at this? I had to change the subclassing to allow passing the bias to `ARG`. The code basically follows the Metainference implementation, but it seems that there is no re-weighting required in the sigma estimation (i.e. no sigma_SEM estimation via weighted variance), only in the forward model averaging, can you confirm this? The actual important part (derivatives) are lines 1659 - 1668.

##### Target release

No preference here!

##### Type of contribution

##### Copyright

##### Tests


tlhr:
The regtest failed on Travis (the derivatives are very slightly off the reference), but it runs fine on my local machine. Is this just due to the slightly changed system architecture?

carlocamilloni:
@tlhr I'll have a look next week, about derivatives it may be a matter of precision but also of mpi/openmp?

tlhr:
> @tlhr I'll have a look next week, about derivatives it may be a matter of precision but also of mpi/openmp?

So it seems to be the compiler / Linux version, previously the Trusty build (gcc 4.8.4) was working, while the Bionic one (gcc 7.4.0) wasn't, and now it's the other way round. Is there a protocol for dealing with this?

carlocamilloni:
So in general we now rely less and less on numerical derivates, it is enough to know that derivatives are correct and then only dump the derivatives from the analytical version. Furthermore you can decrease precision

tlhr:
@carlocamilloni I further reduced the precision, but the virial has a very different magnitude and is preventing the regtest from working. What do you think the best course of action is? Should I heavily reduce the bias felt by the system to get the magnitudes closer?

carlocamilloni:
@tlhr so, I checked the code even if not in super detail but I mean it is the same of MetainferenceBase so it should be fine. The virial derivatives are huge, are they comparable with the original version without reweighing? If this is the case then you should decrease them by some order of magnitude either increasing the sigma values or decreasing the bias in the regtest so to get something numerically more stable

carlocamilloni:
To me this is fine
@maxbonomi are you happy with this?

tlhr:
@carlocamilloni @maxbonomi Can you have a look at the new implementation? I did some checks using fake EM maps with Ala2 and also did a quick test on a larger, more realistic system and get the correct results. Note that there is still an issue with the numerical derivatives, but I'm unsure about how to fix it.

carlocamilloni:
@tlhr @maxbonomi I am happy with the changes and I think it could be merged

Issue #134, opened 2015-04-14
Possibility to read multiple MOLINFO


GiovanniBussi:
It is not currently possible to use MOLINFO more than ones, e.g.:
MOLINFO STRUCTURE=chain-a.pdb
cv1: TORSION ATOMS=@alpha-A7

MOLINFO STRUCTURE=chain-a2.pdb
cv2: TORSION ATOMS=@alpha-A2

MOLINFO STRUCTURE=chain-a3.pdb
cv3: TORSION ATOMS=@phi-A3

does not work as expected (See #133). We should make it possible also in view of the recent changes in the PDB format to allow very big structures with a lot of chains.

PR #593, opened 2020-07-01
Allow multiple MOLINFO

GiovanniBussi:
##### Description

This is to allow multiple `MOLINFO` commands in the same file (see #134).

A typical use case would be for torsion selection with modified nucleotides. One could use a "fake" pdb file, where residue names are standard ones (to allow torsion search to work), and later on a "real" pdb file where residue names are not standard, to allow DUMPATOMS to contain the correct residue names. E.g.
```
MOLINFO STRUCTURE=AA_fake.pdb # here we use standard residue names

chi: TORSION ATOMS=@chi-1

[...]

MOLINFO STRUCTURE=AA.pdb # here we used the modified names

DUMPATOMS ATOMS=@allatoms FILE=out.gro
```

##### Target release

##### Tests

- [ ] I verified that all regtests are passed successfully on Travis-CI. *(waiting for the full test suite)*


GiovanniBussi:
The way I implemented it now, each MOLINFO supersedes the previous ones. In order to do what you suggest we should also add some keyword to specify which MOLINFO should be used at each atom selection. I think this might complicate the syntax. Or were you thinking about some easy way to do it?

One possible addition would be something like `MOLINFO END` to be used such as
```
MOLINFO STRUCTURE=AA.pdb
MOLINFO STRUCTURE=AA_fake.pdb # here we use standard residue names
chi: TORSION ATOMS=@chi-1
MOLINFO END
# past this line we use AA.pdb
```

This would allow to have "self contained" include files, that "push" a topology, use it, and "pop" it. This would be relatively easy to implement (add a `END` flag and change the function that searches the last `MOLINFO`), but I am not sure how useful it would be.

carlocamilloni:
The idea of an optional MOLINFO END keyword is nice I think, but is anyway good as already is

PR #599, opened 2020-07-08
fixed exponential overflow of rct in case of non-tempered metad

invemichele:
##### Description

This fixes the issues reported in the mailing list https://groups.google.com/g/plumed-users/c/maD46ZOfss4/m/gXtqFcwoCAAJ

##### Target release

I would like my code to appear in release v2.5 and above 

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -1946,21 +1946,21 @@ void MetaD::computeReweightingFactor()
     minusBetaF=1;
     minusBetaFplusV=0;
   }
-  const double big_number=minusBetaF*BiasGrid_->getMaxValue(); //to avoid exp overflow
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
This looks ok to me and I would like to merge it.

@invemichele just to make sure I understand the fix, there's no expected change in the result expect when there are overflows right? Can I assume this to be a bug fix and backport it to v2.5? I ask this because you wrote that you want this on v2.5 but you opened a pull request on master branch.

Thanks!

Giovanni

invemichele:
Yes, it is a bug fix so should be backported (good to know that I was supposed to change it in the v2.5 branch, next time I'll do it that way).
The previous version had 2 separate issues, but only when using non-tempered metad, that is not tested in regtest. One is the easy overflow and the other a missing kbt factor.

PR #615, opened 2020-09-30
Parsing integers with lepton

### LINKED ISSUE #614: Support to simple math also when parsing int
When parsing a `float`, it is possible to use a simple math expression in the plumed input file.
It would be nice to extend this feature also to the parsing of `int` types.

As of now it is is what happens, e.g. in METAD:
  HEIGHT=2\*10 -> ok
  PACE=2\*10  -> Error: keyword PACE could not be read correctly

Maybe it could be done by adding to `Tools::convertToAny` a call to lepton similar to the one in `Tools::convertToReal`. But I am not sure, it might be more complicated and require extra checks...

GiovanniBussi:
##### Description

Closes #614.

I am opening this PR to run the complete test suite on travis-ci.

@invemichele could you double check if you think the checks I added are sufficient for corner cases?

The minimum requirement is that anything that was parsed before this change will still be parsed with the same result. This should be true I think.

##### Target release

If there is any doubt, I will just postpone merging to v2.8.

##### Tests


src/tools/Tools.cpp
```diff
@@ -46,24 +48,60 @@ bool Tools::convertToAny(const string & str,T & t) {
...
+   
+    // do the actual conversion
+    auto tmp=static_cast<T>(std::round(r));
+
+    // it should be *very close* to itself if converted back to double
+    double diff= r-static_cast<double>(tmp);
+    if(diff*diff > 1e-20) return false;
+    // this is to accomodate small numerical errors and allow e.g. exp(log(7)) to be integer
+
+    // it should be change if incremented or decremented by one (see https://stackoverflow.com/a/43656140)
```
invemichele, line 92:
```suggestion
    // it should change if incremented or decremented by one (see https://stackoverflow.com/a/43656140)
```

invemichele:
It looks fine to me!

PR #622, opened 2020-10-17
OPES for expanded ensembles

invemichele:
##### Description

This adds the OPES_EXPANDED bias action, together with all the needed expansion CVs, as described in https://arxiv.org/abs/2007.03055
It also includes an experimental OPES_METAD_EXPLORE action.

It is still a **work in progress**, I made the pull request mainly to run all the automatic tests.

##### Target release

I would like my code to appear in release **2.8**
##### Type of contribution

- [x] ~~new module contribution or~~ edit of a module authored by you

##### Copyright

##### Tests


invemichele:
Among other more serious things, I have a `-0.0` issue in my regtests (very small numbers might be randomly printed as `0.0` or `-0.0`). The problem is that `regtest/scripts/fixzeros.sh` substitutes the minus sign with a white space, while I would need to remove it in order to always pass the regtest. 
I guess I could fix this by setting `FMT=% f` in my plumed file, but I can't find a way to make plumed properly parse the white space...
Any suggestion? Is there a simple solution for this?

GiovanniBussi:
@invemichele if you literally mean `FMT=% f` with a space after the `%`, then this should work:
`FMT={% f}`
(I didn't know about this format)

invemichele:
Perfect, from my tests this seems to solve the issue!

Another possibly more serious problem is that I get this error from codecheck:
`[opes/OPESmetad.cpp:220] (error) :unknownMacro: There is an unknown macro here somewhere. Configuration is required. If PLUMED_REGISTER_ACTION is a macro then please configure it.`

This is probably linked to a little trick that I do. I want to define a new action OPES_METAD_EXPLORE, that is basically identical to OPES_METAD, but for a bool that is set to true. For various reasons I would really prefer to have two separate actions, instead of simply adding a flag to OPES_METAD. 
What I did is the following: I defined a template class `OPESmetad<mode>` that is then statically registered twice, with two different values of the above mentioned bool. This worked well in all my tests, but apparently causes this weird `unknownMacro` error, that appears the second time PLUMED_REGISTER_ACTION is called. I really don't understand why this happens.

One way to fix this is of course to simply copy and paste all the code and have two distinct classes, but it would be less handy to maintain...

valsson:
You could try to use ActionShortcut, see for example src/analysis/WhamWeights.cpp so OPES_METAD_EXPLORE would call OPES_METAD with an additional FLAG (perhaps hidden) that would set the boolen to true

valsson:
To explain better what I mean. In the past I have done something similar to this after instructions from @gtribello: 

```cpp
class MyShortcut : public ActionShortcut {
public:
  static void registerKeywords( Keywords& keys );
  MyShortcut(const ActionOptions&);
};

PLUMED_REGISTER_ACTION(MyShortcut,"MY_SHORTCUT")

void MyShortcut::registerKeywords( Keywords& keys ) {
  ActionShortcut::registerKeywords( keys );
}

MyShortcut::MyShortcut(const ActionOptions&ao):
  Action(ao),
  ActionShortcut(ao)
{
  std::string input_string = convertInputLineToString();
  readInputLine( getShortcutLabel() + ": MY_ACTION " + input_string );
}



}
```

This is just a stupid example that would not do anything, just call MY_ACTION with the same keywords as MY_SHORTCUT was called. But you can try to modify the input_string.

invemichele:
Thanks a lot Omar, this looks exactly like what I wanted to do!
I'll give it a try.

I am still puzzled by the `unknownMacro` error, but whatever...

GiovanniBussi:
@invemichele I think it's a bug of cppcheck. In other places we solve it with a suppression:
```
> git grep unknownMac
src/crystallization/Q3.cpp:// cppcheck-suppress unknownMacro
src/crystallization/Q4.cpp:// cppcheck-suppress unknownMacro
src/crystallization/Q6.cpp:// cppcheck-suppress unknownMacro
```

invemichele:
Ok, this is good news. 
Then I prefer to keep my current implementation, since from what I understand in order to use ActionShortcut I would need to rewrite the registerKeywords function.
Thanks @valsson and @GiovanniBussi !

invemichele:
Travis is failing for some error not related to my code.

I consider this pr ready to be merged. However, since the code is still relatively young, to be on the safe side I would prefer to have it merged in v2.8, not in the next plumed stable release.

GiovanniBussi:
@invemichele I think you still have python scripts in a subdirectory. Would it be possible to remove them and, if you wish, put them as part of a separate tutorial? If you put the script there nobody will find it. Thanks!

invemichele:
For the future I am working on putting together a more decent postprocessing. Writing is as a plumed cltool turned out to be more complicate than I first thought, so I will probably go for a tutorial. However, this might take a while since it is not a top priority, because technically with REWEIGHT_BIAS one can already do all the necessary OPES postprocessing.

In the meantime, what I did was to add a link in the doc of [OPES_METAD](https://www.plumed.org/doc-v2.7/user-doc/html/_o_p_e_s__m_e_t_a_d.html)  to that directory with the scripts.
Unfortunately the link goes to the same opes branch I am using for this PR (I had forgotten the folder was there), so the link will break once I delete the folder. But it is not a big problem, I will add it back as soon as the PR is merged and the link will work again.
Do you think it is ok?
Otherwise I can create another PR to v2.7 where I change the link to point to my plumed-nest OPES repository, where I have a copy of the folder with the scripts.

Anyway this is temporary, in a few months I will remove that link and make the tutorial.
Sorry about the mess!

GiovanniBussi:
I would put the material in a separate repository so that you can update it
if you wish.

Giovanni

invemichele:
On my side this PR is ready.
(I tend to always make small improvements to the doc, but they are not really important)

GiovanniBussi:
Thanks @invemichele !!

PR #623, opened 2020-10-19
Fixing a simple LGTM warning about multiplication overflow

jandom:
##### Description

Found a couple of low-hanging fruits in LGTM output, mainly multiplications overflows. This PR attempts to fix them. 

https://lgtm.com/projects/g/plumed/plumed2/?mode=list

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/ReweightWham.cpp
+++ b/src/bias/ReweightWham.cpp
@@ -134,7 +134,7 @@ double ReweightWham::getWeight( const unsigned& iweight ) const {
 }
 
 void ReweightWham::calculateWeights( const unsigned& nframes ) {
-  if( stored_biases.size()!=nreplicas*nframes ) error("wrong number of weights stored");
+  if( stored_biases.size()!=(unsigned long) nreplicas*nframes ) error("wrong number of weights stored");
   // Get the minimum value of the bias
   double minv = *min_element(std::begin(stored_biases), std::end(stored_biases));
   // Resize final weights array
--- a/src/core/MDAtoms.cpp
+++ b/src/core/MDAtoms.cpp
@@ -122,7 +122,7 @@ void MDAtomsTyped<T>::setUnits(const Units& units,const Units& MDUnits) {
 
 template <class T>
 void MDAtomsTyped<T>::getBox(Tensor&box)const {
-  if(this->box) for(int i=0; i<3; i++)for(int j=0; j<3; j++) box(i,j)=this->box[3*i+j]*scaleb;
+  if(this->box) for(int i=0; i<3; i++)for(int j=0; j<3; j++) box(i,j)=(double) this->box[3*i+j] * (double) scaleb;
   else box.zero();
 }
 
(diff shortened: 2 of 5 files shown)


GiovanniBussi:
@jandom thanks!

**About those in core/MDAtoms.cpp:**

One very small doubt I always add about this suggestion is that in principle `float*float` is faster than `double*double`, plus we don't really expect overflows here. Perhaps adding an explicit cast to the product is sufficient to remove the warning?
-    positions[index[i]][0]=(double) px[stride*i] * (double) scalep;
+    positions[index[i]][0]=static_cast<double>(px[stride*i] * scalep);

Or can we make sure there's not a measurable impact on the performance? I am afraid that these operations are applied on all atoms in the MD code (not just those used in PLUMED), that could be a very large number.

Also notice that C++ style conversion is clearer (either with `double(.)` or `static_cast<double>(.)`) since it's more clear what the conversion refers to.

**About the `unsigned long` conversions:**

Since the warnings are raised by a comparison with a size, perhaps [size_t](https://en.cppreference.com/w/cpp/types/size_t) is more appropriate? I think they should be identical on most platforms, but formally it is more correct. One should also make sure one of the header defining std::size_t is included (e.g. `cstddef`).

jandom:
@GiovanniBussi sorry about this – it's a semi-useless PR but just starting off here, and these looked simple enough - i should have known better ;-) 

**Double casts**

This sent me down an SO rabbit hole with the question of speed

https://stackoverflow.com/questions/4584637/double-or-float-which-is-faster

In the absence of any existing speed test, that would say if it indeed matters, I'd rather not touch this code. 

Happy to go with the C++ conversion but I'm out of my league when it comes to performance

**Size_t casts**

That indeed seems to be less of a minefield, should I focus my efforts here?

jandom:
Hmmm, how about we close this one and I'll try anther PR for the size_t thread (seems a little less muddy). 

This PR is broken anyway, something in the conda build broke but nothing obviously related to the files I touched 

```
compiling  ../analysis/Average.cpp

compiling  ../analysis/AverageVessel.cpp

mkdir: deps: File exists

make[4]: [../maketools/make.rules:15: ../main/main.o] Error 1 (ignored)

mkdir: deps: File exists

mkdir: deps: File exists

make[4]: [../maketools/make.rules:16: ../analysis/Average.o] Error 1 (ignored)

make[4]: [../maketools/make.rules:16: ../analysis/AverageVessel.o] Error 1 (ignored)

mv: rename ../main/main.d to deps/../main/main.d: No such file or directory

make[4]: *** [../maketools/make.rules:19: ../main/main.o] Error 1

make[4]: *** Waiting for unfinished jobs....

mv: rename ../analysis/Average.d to deps/../analysis/Average.d: No such file or directory

make[4]: *** [../maketools/make.rules:19: ../analysis/Average.o] Error 1

mv: rename ../analysis/AverageVessel.d to deps/../analysis/AverageVessel.d: No such file or directory

make[4]: *** [../maketools/make.rules:19: ../analysis/AverageVessel.o] Error 1

mv: rename ../analysis/AnalysisBase.d to deps/../analysis/AnalysisBase.d: No such file or directory

make[4]: *** [../maketools/make.rules:19: ../analysis/AnalysisBase.o] Error 1

make[3]: *** [Makefile:112: all] Error 2
```

PR #625, opened 2020-10-21
Playing around with googletests

jandom:
##### Description

This is my silly attempt at demonstrating how unit tests (tests per class) can be used to write sappy and fast test cases.

Their principle advantages are two fold
- they are very fast to run
- allow to access corner cases that are difficult to target with regression tests

It has a dependency on googletests , I know that plumed has the capacity to detect these optional external libraries but I'm too much of a build-chain noob to get it to work. Also my Makefile changes are a mess #HelpNeeded 

To build googletest on a MacOS

```
git clone https://github.com/google/googletest/
cd googletest
mkdir build
cmake ..
make 
```

Then build plumed (I don't know how to do it in a cleaner way)

```
./configure \
    LDFLAGS="-L /Users/jandom/workspace/googletest/build/lib" \
    CPPFLAGS="-I /Users/jandom/workspace/googletest/googletest/include -I /Users/jandom/workspace/googletest/googlemock/include" \
    LIBS="-lgtest -lgmock"
make src -j4
make unit_tests
```

Finally run unit tests (they are just an executable) 

```$ ./unit_test 
Running main() from /Users/jandom/workspace/googletest/googletest/src/gtest_main.cc
[==========] Running 3 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from Foo
[ RUN      ] Foo.Is42
[       OK ] Foo.Is42 (0 ms)
[ RUN      ] Foo.IsStill42
[       OK ] Foo.IsStill42 (0 ms)
[----------] 2 tests from Foo (0 ms total)

[----------] 1 test from AtomNumber
[ RUN      ] AtomNumber.Constructor
[       OK ] AtomNumber.Constructor (0 ms)
[----------] 1 test from AtomNumber (0 ms total)

[----------] Global test environment tear-down
[==========] 3 tests from 2 test suites ran. (0 ms total)
[  PASSED  ] 3 tests.
```

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,9 @@ $(SRCDIRS):
 # compile plumed before tests:
 test: src
 
+unit_test: src
+	ld -lSystem -L /Users/jandom/workspace/googletest/build/lib -lgtest -lgmock -lc++ -lgtest_main src/tools/bar.o src/tools/bar_test.o src/tools/AtomNumber.o src/tools/AtomNumberTest.o -o unit_test
+
 # doxygen
 doc:
 	$(MAKE) -C user-doc
--- a/src/tools/AtomNumberTest.cpp
+++ b/src/tools/AtomNumberTest.cpp
@@ -2,11 +2,11 @@
 #include "gmock/gmock.h" // for the matcher
 #include "AtomNumber.h"
 
-namespace PLMD { 
-  TEST(AtomNumber,Constructor) {
+namespace PLMD {
+TEST(AtomNumber,Constructor) {
 
-    AtomNumber *atomNumber = new AtomNumber();
+  AtomNumber *atomNumber = new AtomNumber();
 
-    EXPECT_EQ(atomNumber->index(), 0);
-  }
+  EXPECT_EQ(atomNumber->index(), 0);
+}
 }
--- a/src/tools/AtomNumberTest.cpp
+++ b/src/tools/AtomNumberTest.cpp
@@ -4,9 +4,18 @@
 
 namespace PLMD {
 TEST(AtomNumber,Constructor) {
-
   AtomNumber *atomNumber = new AtomNumber();
-
+}
+TEST(AtomNumber,Index) {
+  AtomNumber *atomNumber = new AtomNumber();
   EXPECT_EQ(atomNumber->index(), 0);
+  atomNumber->setIndex(42);
+  EXPECT_EQ(atomNumber->index(), 42);
+}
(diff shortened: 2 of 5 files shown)


jandom:
Guys, I think I'm going to throw in the towel on this PR – but I'm still thinking this would be a nice feature, it just really doesn't play nice with the existing plumed boilerplate. Have you guys had a look? What did you think? 

Asked a friend for a bit of help, maybe he can guide this better.

GiovanniBussi:
Dear @jandom, thanks for providing the working example!

I will write below the two reasons why I am a bit skeptical of incorporating this.

**Introducing an extra dependency**

I see that installing the extra library on Travis might be easy and done once for all. However, typical users will not have it installed on their systems. Thus, they will not be able to perform this test. We have historically encountered troubles related to compiler bugs with heavy optimization or errors in the code only showing up with specific compilation flags. The only way to detect these problems is that users test their production library. I think this will not happen if it requires installing an extra dependency. If the extra library is written in C++ (I didn't check, but I guess it's the case) this will also add the complication that it should be compiled with the same flags as PLUMED (e.g., if PLUMED is built with debug flags, that should also be built with debug flags - that's the reason for [these lines](https://github.com/plumed/plumed2/blob/master/.travis.yml#L128-L130) to build boost, for instance). So, the price that we would pay is high, and the gain would be the possibility to write something not so different from `plumed_assert(bar()==42)`.

**Keeping tests within the source tree**

Keeping the code in the source tree does not allow you to run the tests of one version of PLUMED using a different version. Whereas this is something that is expected to fail in some cases (we have non-compatible changes from time to time, and the interfaces of the functions not accessible from outside are changing frequently), it is still something really useful. For this reason, I would prefer to have the code of the tests stored in a separate directory. This clearly can be done also with google tests.

Given this, I see also an advantage: **When tests are performed, the log is nicer**. If this is a strong enough motivation, perhaps we could try to include in PLUMED some simple macro and function that allows to simplify the syntax of the tests and to produce the nice log? I don't want to re-implement the google tests library, but just to get inspiration and perhaps hard code the few things we would use within PLUMED.

jandom:
Completely agreed with all of the above @GiovanniBussi, and I'm closing this. 

What other advantages are out there? It could be just the speed of the test, they are ultimately developer facing allowing them to quickly diagnose if something major is broken. These tests are absolutely no substitute for the super-robust regression test suite you guys already have (and it's really an impressive suite).

PR #633, opened 2020-10-31
Implementing make_unique

GiovanniBussi:
##### Description

I added an implementation of make_unique compatible with C++11.

See [here](https://stackoverflow.com/questions/37514509/advantages-of-using-stdmake-unique-over-new-operator) for a discussion about the advantages and [here](https://stackoverflow.com/questions/17902405/how-to-implement-make-unique-function-in-c11) for the C++11 compatible implementation.

In addition to adding the implementation as `Tools::make_unique`, I replaced all the `new` statements in the default modules where this was appropriate.

I open a pull request in case anyone has comments and to run the full test suite. If there are no objections, I will merge in a week or so.

##### Target release

### DIFF
--- a/src/adjmat/AdjacencyMatrixBase.cpp
+++ b/src/adjmat/AdjacencyMatrixBase.cpp
@@ -162,7 +162,7 @@ void AdjacencyMatrixBase::finishMatrixSetup( const bool& symmetric, const std::v
   vesselbase::VesselOptions da("","",0,param,this);
   Keywords keys; AdjacencyMatrixVessel::registerKeywords( keys );
   vesselbase::VesselOptions da2(da,keys);
-  std::unique_ptr<AdjacencyMatrixVessel> ves( new AdjacencyMatrixVessel(da2) );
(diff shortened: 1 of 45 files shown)


carlocamilloni:
This is good to me, but why don’t we move to c++14 altogether for plumed 2.8?

GiovanniBussi:
That could be done. But notice it would raise the required GCC from 4.8 to 5, and plumed will not compile with the default compiler in Centos 7 (the one we have at SISSA ;-)) without installing additional compilers. Whereas GROMACS is using already c++14, many other codes do not have this requirement. Also considering that there are not so many new features in c++ 14, it might not be worth..

carlocamilloni:
no we can stay with c++11 if there isn't something in c++14 that would be really worth. I think that gromacs now uses c++17 but that for example CUDA are compatible only with c++14, so I think that c++11 strikes a good balance between compatibility and being future proof.
On this gromacs is considering to move to MPI 2 (I am not really expert but it looks there are nice features). Said that to me we can merge this

PR #635, opened 2020-11-05
Update OPESmetad.cpp

carlocamilloni:
@invemichele I think this should fix the periodicity issue when merging kernels, could you double check it?


invemichele:
Thanks for the suggestion! 
Unfortunately it is slightly more complicate, and in fact I probably should have put the `\\TODO` in the `kernel::merge_me_with` function, say at line 1145.
The rule to merge the kernels is the following (from the SI of the paper):
merging
where h is the height and **s** the position of the center.

So, to implement merging through PBC one should do what you did plus something like the following:
```
if s_1-s_2 != difference(s_1,s_2):
  change s_2 to proper PBC image
merge kernels 1 and 2
```

Now that I think about it, probably the proper PBC image can be set with `s_2=s_1+difference(s_1,s_2)` but this might fall outside the periodicity, so it has to be brought back again, `s_2=bringBackInPbc(s_2)`. 

I can try to implement this, and run some tests to check that it actually behaves like I expect!

invemichele:
sorry, the bringBackInPbc should be done on the new center s', not on s_2

carlocamilloni:
Ok good to know, i will have a further look, in the meantime, playing with the code I have also a branch with kernels neighbor list and some heuristic, see opes-neighb

Sent from my iPhone

invemichele:
great, I will check it out!

carlocamilloni:
I tested this and it works correctly, I merge it and I also merged it into the opes-neighbour branch. There I ran a test on ALADP where the results are essentially identical (but I want to do a more quantitative test) but run ~6 times faster. At least on ALADP now the impact of OPES is only slightly more than that of METAD

invemichele:
This sounds great! I have been lazy in implementing the neighbor list because for more complex systems, where one runs multiple cores in parallel, the overhead from the kernel summation is usually not so big. But for something like alanine it makes a big difference, thanks a lot!
I'm a little busy these days, but I will definitely have a look during the weekend

carlocamilloni:
Thanks. actually I have implemented them because otherwise OPES kills the performances when the "complex system" run on GPU and plumed on CPU ;-), I think they may be also a good alternative to SPARSEGRID in METAD.

PR #640, opened 2020-11-08
[WIP] OPES neighbor list

invemichele:
@carlocamilloni I looked at the neighbor list and made a few tests... it's amazing! thanks a lot!
In particular in higher dim (I tested ala4 with 6 angles) it makes a huge difference. But it speeds up also 1D systems, so I probably will want to have it on by default, maybe after a few months, just to be sure everything works fine.

I have a question, how do you chose the 2.3 and 0.6 factors for cutoffs? 
In my tests oftentimes the neighbor list was rebuilt much before a full PACE, but the performance was good. I tried to modify those factors to have longer-lived neighbor lists, but then I ended up with too many kernels in the list and thus worse performance....

##### Description

I did some rewriting of the code, mainly for aesthetics and to better understand it myself. I though it would be better to use as keyword NLIST, as it is done elsewhere in PLUMED. Is it ok for you?

I also added an MPI version of the list creation, and pushed it to my repo just before you pushed your MPI version...
I see that you flatten the kernels to communicate them, I instead opted for communicating the index and then filling in the kernels only after, what do you think about this solution?
I will get the OpenMP part from your code, and also try to use Allgatherv (which should be more efficient) instead of Sum, like I do now.

##### Target release

I would like my code to appear in branch **opes-neighbor**

##### Type of contribution

##### Copyright

##### Tests


invemichele:
@carlocamilloni I see that you are still working on it. I will take a look tomorrow or in the next days (let me know when you are done) and fix the  details so that this PR can be merged.

carlocamilloni:
@invemichele thanks, yes I was still working a bit around it because combining openMP, MPI and non trivial data is always a nightmare, anyway I like your solution of using the index because it allows to reuse the neighbour list also in update(). About using NLIST naming is fine, I usually don't use it when the implementation is not based the NeighborList.cpp class, but we don't have clear conventions. About the 2.3 and 0.6 parameters my idea was that using 4 and 0.5 should be the most conservative choice, in terms of performances is not too bad, 2.3 is just to keep low the number of kernels that is what actually matters for the performances (sqrt(2.3)~1.5) with a small errors in particulars because in OPES the automatic choice for the truncated gaussians is pretty conservative. The 0.6 is to give some chance for the system to "go back towards the center" before updating the neighbour list; said that somehow this choice is pretty good. In the last iteration I have increased 2.3->3 and 0.6->0.5 in view that by parallelising the cycles it may be better to have a few more kernels updated slightly less often. 

with yesterday night changes I think that the openMP stuff now works, so I think it may be worth to merge your changes and then we can have some more tests and iterations if needed

invemichele:
Ok, then I could implement the whole neighbor list thing just by masking the `kernels_` vector. Basically the NL will be an array of indexes and every loop that before was running over all the kernels now will run only on the NL indexes (except of course the loop to create the NL). It should be safe to use it also when adding a new kernel and checking for merging. The only thing I have to be cautious with is multiple walkers, because they will have different NL.
I guess implementing NL as array masking is not good practice, because it can kill cache optimization, but given that kernels are a nontrivial data type probably there was no such optimization in the first place... But the only way to know is to try.

About the NL parameters, I think we could set some reasonable default value (I'll check out the new ones), and add a keyword to change them for users with specific needs.

invemichele:
Ok, I have done the changes. 
Now the neighbor list is used also when a new kernel is added, to speedup the merging. By default any new kernel is added to the nlist, and it is not updated, so it can run longer than a PACE.

When using multiple walkers the nlist of each walker merge before adding the new kernels, so that the merging works properly. There is a risk that this makes the nlist quite big and they might not be updated for a few steps. I think this is system specific, so I added a keyword NLIST_PACE_RESET to force the nlist to reset at each pace (as before), so one can set it if needed. I'll do some tests on this.

The only thing that is missing now is the OpenMP, after that I will make the PR mergeable

invemichele:
I added the OpenMP parallelization. I have never used it, and I did not test it so I would be glad if you could have a look @carlocamilloni 
I decided to split the loops to avoid some unnecessary operations in case `OpenMP::getNumThreads()==1` .

From my point of view this could be merged to v2.7, leaving the NLIST as an optional "experimental" feature. But if after using it for a while we don't see problems, then I would definitely set it as default and just leave an option to turn it off.
I like this neighbor list a lot, I naively didn't expect such a drastic change in performance. Thanks again for the contribution!

carlocamilloni:
@invemichele Great! I am merging it a run some test. I think it may be worth to include it as experimental in v2.7 and move it to "production" in 2.8. I would also add some check on the NLIST parameters because they may result in completely wrong results. On a side note I think it may be useful to add a SIGMA_MIN keyword that would do the same of FIXED_SIGMA but automatically once the minimum value is reached

invemichele:
you know, I used to have this SIGMA_MIN option in an old version of the code, but somehow it got lost and didn't make it to the final version. I will add it back!

carlocamilloni:
@invemichele merged in v2.7, I have run a couple of tests with mpi/openMP and it works well! I have also added a test on the parameters for NLIST with some heuristic, I think it is safer in this way

invemichele:
Thanks! I will also add some regtest and do the manual merging of this file with the master branch, to be sure it goes properly

PR #649, opened 2020-11-20
Fix to ves_md_linearexpansion with periodic potential

bpampel:
##### Description

The periodic boundary conditions of the ves_md_linearexpansion tool previously shifted the potential range so it is symmetric with respect to the origin. (Example: `[0,1] -> [-0.5,0.5]`)
This PR changes the way the pbc is calculated, so if the specified range is not symmetric, it is actually kept.

Also includes a new regtest to verify the correct behaviour.

This should not change any simulations where a symmetric range was specified, so it is fully backwards compatible.

The VES maintainer @valsson has seen and agreed to the changes.

##### Target release

I would like my code to appear in release v2.5 and all newer

##### Type of contribution

##### Copyright

##### Tests

- [ ] I verified that all regtests are passed successfully on GitHub Actions.


GiovanniBussi:
Looks reasonable to me. @valsson do you confirm? Thanks! Giovanni

valsson:
Yes, I approve of the PR. Omar

PR #650, opened 2020-11-21
OPES: added some more OPENMP parallelism

carlocamilloni:
- Zed
- Mergeable
@invemichele


invemichele:
Thanks!
is there a way to test OpenMP in regtest? what should I put in the config?

invemichele (approved):

GiovanniBussi:
> Thanks!
> is there a way to test OpenMP in regtest? what should I put in the config?

@invemichele regtests on GitHub Actions should already run with 2 threads (see [here](https://github.com/plumed/plumed2/blob/63008b0183800427b11f00071b2cfa049a002432/.github/workflows/ci.yml#L11)). You can change the number of threads changing the `PLUMED_NUM_THREADS` env var in the `config` file of a specific regtest

invemichele:
OK, perfect, then there is no need to add a new regtest!
We already have one running with the SERIAL keyword, that removes parallelization

PR #653, opened 2020-11-27
Typesafe interface

GiovanniBussi:
##### Description

This is an attempt to add type checks to the C++ and Python interfaces. In particular, the idea is that when someone calls the functions `p->cmd("whatever",ptr)` with the wrong pointer type an error should be reported. This is particularly relevant for the Python interface that can be used interactively (and an exception is preferred to a segfault), but would also be useful for the C++ interface.

*Implementation details*. The way it is implemented is by defining several overloads of the cmd function in `wrapper/Plumed.h` and send to PLUMED a code identifying the pointer datatype. The check will only be done if the library is recent, but the calls will be backward compatible as usual (i.e. it will be possible to call PLUMED 2.7, of course without the benefits of type checks). I implemented so that it is possible to pass the base type (e.g. `int` or `FILE`) and const qualifiers. Const qualifiers conversions are permitted as in C++. I ignored the different between `int` and `unsigned`, basically identifying ints only based on their size. More types could be added (e.g. `long double`?), but I don't expect we plan to pass much more. Within PLUMED, instead of brining around `void*` I pass a `TypesafePtr` that can be used as follows:
```
step=*val.get<const int>();
```
This example checks that the pointer is of the correct type (in this case, it will accept `const/no const int/unsigned`). Otherwise, throws an exception. (**NOTE**: after 0031ca8, the instruction above should be changed to `step=*val.get<const int*>();`)

~**TODO** Check python interface. I am afraid I discovered a small issue. @gtribello I notice that integers where passed as `int64` but then interpreted as `int`. In x86 it works (luckily! the first 4 digits of a int64 are the same of a int), but might break on other architectures.~ (DONE, se comment below)

*Possible extension: static or dynamic disabling.* Currently checks are always done. I could modify the code so that they are e.g. disabled when compiling the MD code with `-DNDEBUG`. Perhaps an env var would be better.

*Possible extension: boundary checking.* One could call `p->cmd("setPositions",pos,natoms)` and there could be a boundary check in PLUMED. Currently it is possible to pass the size of the array but is not used anywhere (easy to add). **(DONE, see comments below)**

*Possible extension: FORTRAN wrappers.* I would like to investigate the possibility to enable it also in FORTRAN wrappers, since FORTRAN allows overloads similar to C++. Might be useful since we have a lot of FORTRAN codes using PLUMED around.

*Possible extension: control ownership.*  The current version only checks type. I think that ownership transfer between the MD code and PLUMED is beyond what we can do portably on multiple languages (e.g. passing a `std::unique_ptr`). However, we could allow some basic checks such as: tell from the MD code that a pointer should not be used anymore. I implemented a draft for this, but I am not sure how useful it is. If this is not useful, I will remove the related code before merging. **(Added pass-by-value make sure plumed does not take the address, still not sure if it's worth allowing users to pass vectors with controlled ownership)**

Feedbacks are welcome!

##### Target release


GiovanniBussi:
Regarding this:
>TODO Check python interface. I am afraid I discovered a small issue. @gtribello I notice that integers where passed as int64 but then interpreted as int. In x86 it works (luckily! the first 4 digits of a int64 are the same of a int), but might break on other architectures.

It looks like the only integer passed from the python interface in the tests was for `cmd("getDataRank")` which actually wants a pointer to long int. In particular: when passing python ints, they are interpreted as C int, then passing numpy int array they are interpreted as long ints. The code in the example is working correctly, except that it is probably not possible to use "setStepLong" (the only function that takes a long int in input). Similarly, it would not be possible to use arrays that PLUMED interprets as ints.

Type checking will make life easier in these cases I guess.

carlocamilloni:
This looks very important! Go ahead!

GiovanniBussi:
I made further progress.

*Bound checks are possible*. Just pass something like `cmd("setVirial",virial,8)` and you will receive an error since PLUMED will try to access 9 elements at this pointer. Something similar will happen in Python, although in that case we already assume numpy array and thus the size will be inferred. That is:

```
virial=np.zeros((3,2)) # <- incorrect, should be 3,3!
p.cmd("setVirial,virial)
```
will trigger an exception.

*Pass-by-value is possible*. This is the most basic way to control ownership. In particular, you can now pass an object by value and be sure that PLUMED will not store a pointer to it. Something like:

```
int nstep=1000;
int flag=0;

p.cmd("setStep", nstep); // <- this is ok by value

p.cmd("setStepLong", nstep); // <- error!
// should have been:
p.cmd("setStepLong", (long) nstep);
// This is required to be able to pair with PLUMED <=2.7,
// that would not be able to do the conversion.

p.cmd("setStopFlag",flag); // <- error!
// should have been:
// p.cmd("setStopFlag",&flag);
```

The same check is done when passing ints or floats from python. Since this is basically covering some special cases that were checked in the python interface (`getBias` and `getDataRank`), I removed the explicit checks (in particular, they should both be references now). There is still no check on ownership of vectors or strings. I will think if there is a simple way to introduce it.

GiovanniBussi:
OK I think this is kind of done. There are still a few things that I want to cleanup but the interface works. I cc people that I know are using / used / looked at the C++ (@carlocamilloni @gtribello @dvdesolve @akohlmey @PabloPiaggi @invemichele @peastman ) or Python interface (@gtribello @ceriottm @dwhswenson @tonigi) in case someone wants to comment (sorry for the spam). I plan to then cleanup the implementation, perhaps remove a couple of redundant options (see below), and merge in master (to be v2.8). Once this is merged, I will have a look at the possibility to have checks also in FORTRAN. Thanks!

**C++ implementation**
Here's a summary of the checks performed in C++.  `old` means current behavior, `new` means this branch.
```
int natoms=10;
double natomsd=10.0;

// *** type check ***

p.cmd("setNatoms",&natoms);
// old: ok
// new: ok

// incorrect type
p.cmd("setNatoms",&natomsd);
// old: random results
// new: throw exception

// MPI_Comm cannot be strictly type checked since its type depends
// on the MPI variant
p.cmd("setMPIComm",&mpi_comm);
// old: ok
// new: ok

// *** pass by value ***

// Note: no conversions are performed, user has to pass the
// right type or an exception will be thrown

p.cmd("setNatoms",natoms);
// old: not compile
// new: ok

// incorrect type
p.cmd("setNatoms",natomsd);
// old: not compile
// new: throw exception

// not checked
p.cmd("setMPIComm",mpi_comm);
// old: MPICH not compile / OPENMPI compiles + random results
// new: MPICH ok (it's a int!) / OPENMPI compiles + random results

// *** boundary check ***

std::vector<double> pos(29);
// NB: 10 atoms require 30 elements!
p.cmd("setPositions",&pos[0]);
... // nothing happens until the pointer is used calling "calc" ...
p.cmd("calc");
// old: random results
// new: random results (array size not known)


std::vector<double> pos(29);
// NB: 10 atoms requires 30 elements!
p.cmd("setPositions",&pos[0],29);
... // nothing happens until the pointer is used calling "calc" ...
p.cmd("calc");
// old: not compile
// new: throw exception

```

Notice that NAMD patch has to be fixed (at [this line](https://github.com/plumed/plumed2/blob/4ce8dc4c898b31c9250d5c581b157c4314be7e72/patches/namd-2.13.diff#L98) the pointer should be `const int* p;` for const correctness; now it is `int* p`).


**Python implementation**

From Python the thing is pretty similar, with the improvement that when passing numpy arrays plumed detects the shape and check if the number of elements is appropriate.
```
p.cmd("setNatoms",10)
...
pos=np.ones((9,3)) # wrong!
p.cmd("setPositions",pos)
# old: random results
# new: raise exception

pos=np.ones((3,10)) # wrong!
p.cmd("setPositions",pos)
# old: random results
# new: random results - only number of elements is checked and not shape
```

**Other functionalities**

There are some other functionalities already implemented, but I might remove them to simplify the code unless there is some particular interest:

- It is possible to add a callback function that is called when PLUMED does not need a pointer anymore. Could be used to pass in `std::unique_ptr` (the callback function could delete the object). Not sure how useful it would be. MD codes using this will not be compatible with PLUMED<2.8

- MD codes can signal to PLUMED that a pointer has expired and should not be used anymore (an exception will be raised if the pointer is used again). This would be backward compatible, but I am afraid it would be difficult to use.

Other things that would be relatively easy to add, I would add them if there is particular interest:

- Passing `std::vector` from C++ by reference, with automatic bound checking
- Checking the full shape in numpy arrays.
- The exception object could be extended to bring information about the required type/size. This could allow, e.g., Python to automatically convert to the right types if needed.

tonigi:
Sounds great. My only minor comment is that  

> Checking the full shape in numpy arrays.

may be useful, as it is easy to get C/Fortran atom ordering wrong. :)

PabloPiaggi:
This is very useful @GiovanniBussi . Good to know that now the C++ and Python interfaces will have type checks. Since this is backwards compatible I don't think there is anything that has to be changed, for instance in the [Lammps interface](https://github.com/lammps/lammps/blob/master/src/USER-PLUMED/fix_plumed.cpp), right?

GiovanniBussi:
> This is very useful @GiovanniBussi . Good to know that now the C++ and Python interfaces will have type checks. Since this is backwards compatible I don't think there is anything that has to be changed, for instance in the [Lammps interface](https://github.com/lammps/lammps/blob/master/src/USER-PLUMED/fix_plumed.cpp), right?

@PabloPiaggi in principle nothing, except in cases where incorrect types where passed. I found a const error (actually, probably my fault) in the NAMD interface, so it is possible that there are errors out there also in other codes that incorporate calls to PLUMED. Type errors are less likely, since they should have already been reported with catastrophic failures (although passing a long int as an int might have been used without being detected). Size (and perhaps shape, as suggested by @tonigi) will only be checked when provided, so should not break anything.

The version in the current pull request does not check `void*` (that is, if you call `cmd("setNatoms",(void*)&natoms)` it works), but I plan to remove this and to allow exactly the conversions that are allowed in C++ (`anything*` to `void*` is ok, but  `void*` to `anything*` is not). But I would be surprised if MD codes use this syntax.

In any case, it will be possible to shut down the check at run time with an env var. Probably I should add that to the error message, so that a user finding a problem and knowing that the code is working correctly would be able to override the check.

PR #659, opened 2020-12-21
Faster and safer OPES_EXPANDED

invemichele:
##### Description

Various improvements to OPES_EXPANDED:
- removed `long double` and used instead a safe log-sum-exp, which brings to a speedup of more than a factor 3 in some (extreme) cases. It is also more stable. Added a regtest with a setup that would fail with old implementation
- added options to store and restart from STATE, similar to OPES_METAD
- each MPI process now works on a different set of DeltaFs, and the full vector is only gathered for printing it
- minor cleanup

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Thanks @invemichele !

PR #662, opened 2021-01-12
DRR: add a new option for specifying the output name of --merge.

HanatoK:
##### Description

The old behavior of generating output filename for --merge is not
suitable in the cases that the input filenames contain "/", which
indicates the input files are in different sub-directories. Simply
joining the input filenames as the name for output does not work,
so I add a new option --merge_output for these cases. After applying
this patch, the output name can be specified as following:

`plumed drr_tool --merge dir1/win.drrstate,dir2/win.drrstate --merge_output xxx`

##### Target release

I would like my code to appear in release v2.7

##### Type of contribution

##### Copyright

##### Tests

- [ ] I verified that all regtests are passed successfully on GitHub Actions.


HanatoK:
@carlocamilloni 
This is the same PR as #661 but for v2.7. I tried to rebase #661 to v2.7 directly on github but failed, so I made a new one.

PR #663, opened 2021-01-20
opes: fix for too small initial sigma estimates

carlocamilloni:
@invemichele

### DIFF
--- a/src/opes/OPESmetad.cpp
+++ b/src/opes/OPESmetad.cpp
@@ -911,8 +911,10 @@ void OPESmetad::update()
       for(unsigned i=0; i<ncv_; i++)
         av_M2_[i]/=(1-bias_prefactor_);
     }
-    for(unsigned i=0; i<ncv_; i++)
+    for(unsigned i=0; i<ncv_; i++) {
       sigma[i]=std::sqrt(av_M2_[i]/adaptive_counter_*(1-bias_prefactor_));
+      if(sigma[i]<1e-6) sigma[i]=1e-6;
+    }
   }
   if(!fixed_sigma_)
   {


carlocamilloni:
We got a case were the CV was almost always 0 in the initial stage and the code segfaulted

invemichele:
Thanks!
Do you think this is a problem that can arise only with the initial estimate or could happen also in the middle of the simulation?

In the first case I would put an error, so the user knows there is something wrong with the adaptive sigma. But if it can happen in the middle of a simulation it is probably better not to kill it, but just send a warning.

In both cases we could implement it so that if SIGMA_MIN is used it will not be triggered.

I have been adding up various small improvements, I could put this fix in as well and do a pull request

invemichele:
I have done both things here https://github.com/invemichele/plumed2/commit/e049b7e45e2dbbeeb3141fd69d0ff50c55b39f13
what do you think @carlocamilloni ?

carlocamilloni:
@invemichele I am wondering if actually sigma_min is enough to fix:           
const double s_rescaling_i=std::max(s_rescaling,sigma_min_[i]/sigma[i]);
if sigma here is 0 you get a segfault irrespectively of sigma_min right? (I mean, at least before your correction)

invemichele:
you are right, it would crash if `sigma[i]=0`, I thought this could never happen...
I changed it so it sets sigma=sigma_min before trying the rescaling. This should make it robust to failure.

While I was looking at it, I realized that I was only adjusting the height when sigma changed due to the shrinking, and not when it was changing because it was adaptive...
Now I rewrote the code so that any change in sigma does affect the height, thus keeping everything properly normalized.
I did some tests, and this does bring some improvements! in particular it makes less severe the cases where the adaptive sigma starts with a quite bad guess, because it is faster to readjust it.

PR #665, opened 2021-01-23
some OPES_METAD improvements

invemichele:
##### Description

Various minor improvements to OPES_METAD. 

The most relevant is that now the kernel heights are correctly rescaled also according to adaptive sigma changes (see [here](https://github.com/invemichele/plumed2/blob/v2.7/src/opes/OPESmetad.cpp#L979)). Some tests have shown that this improves recovering from a bad initial sigma guess. 
Now an extra parameter (the initial sigma guess, `sigma0`) is stored in the STATE file, see e.g. [this regtest](https://github.com/invemichele/plumed2/blob/v2.7/regtest/opes/rt-opes_metad-bis/Restart.data#L12), and needed for restarting from STATE an adaptive sigma simulation. Thus to restart form a STATE file obtained with the previous OPES_METAD version, one has to manually add to it the very first value of sigma that is stored in the KERNELS file.
Restarting a simulation without adaptive sigma or from KERNELS file is instead back compatible.

Other changes:
- fix issue with extremely small adaptive sigma (alternative to #663)
- now the STATE file is dumped after new kernels are added. This makes restarting from STATE and from KERNELS files more similar
- some improvements to the log
- removed useless `inline` and other cleanup

##### Target release

I would like my code to appear in release **v2.7**
**CAUTION: _this cannot be automatically merged to v2.8, I will make another PR with the manually-merged code_**

##### Type of contribution

##### Copyright

##### Tests


invemichele:
@carlocamilloni if you could have a look it would be great!

carlocamilloni:
@invemichele sounds good, but I haven't yet checked it in detail (tomorrow). only concern is that writing the STATE file as often as kernel will impact the performances. I would keep the option to write it less often because this can be a large file

invemichele:
Yes, sure, no rush!
Sorry, my sentence was ambiguous: with "now the STATE file is dumped after new kernels are added" I meant that the `dumpStateToFile` function is now _after_ the `addKernel` one, instead of _before_. But it is not called every time, only once every STATE_WSTRIDE, as previously.

PR #687, opened 2021-05-04
Attempting to add GHBFIX and regtest-ghbfix

ThorbenF:
##### Description

This is an attempt to add the colvar GHBFIX, which essentially is a switching function for atom pairs, being -1 for small distances and 0 for large distances with a smooth interpolation inbetween them.

This variable computes the ghbfix interaction among GROUPA and GROUPB as defined in Kührová et al.(2016): Computer Folding of RNA Tetraloops: Identification of Key Force Field Deficiencies. JCTC.
It can be scaled as desired by specifying interaction scaling parameters and energy units.

This is for the attention of @GiovanniBussi

##### Target release

In the next new plumed version.

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@ThorbenF a few comments below:
- There's a failure on the codecheck job on GitHub Actions: `[colvar/GHBFIX.cpp:0] (error) :plmd_astyle: astyle not satisfied`. Please apply astyle with `make astyle`
- Don't worry about other failures on GitHub Actions. There's something wrong with installing boost libraries. I will check it again tomorrow.
- You marked this `new module contribution or edit of a module authored by you`. If you do not add a full module you should mark the first option `changes to code or doc authored by PLUMED developers` (it's not very clear, we should change it in the template). The third options is only for adding a full module.
- Regarding the tests, I would recommend adding more lines to `regtest/basic/rt-ghbfix/scalingParameters.dat` and also using strings instead of numbers, so as to check that the matching is done correctly.

Thanks!

GiovanniBussi:
@ThorbenF the problem installing boost should be solved now. If you rebase your branch on the current master, tests with debug flags and test on centos7 should not fail anymore

GiovanniBussi:
@ThorbenF I am ready to merge. A couple of possible changes:
- `std::map<std::string,unsigned> MapTypesTable;` is only used in the constructor. I would declare it inside the constructor rather than as a class member (this is pure aesthetic)
- The following code can be optimized:
```
  double distance=std::sqrt(distance2);
  double result;
  const double rdist = (distance-d0);
  if(distance>dmax) {
    result=0.;
    dfunc=0.0;
    return result;
  }
```
change to (to be tested):
```
  double result;
  if(distance>dmax_squared) {
    result=0.;
    dfunc=0.0;
    return result;
  }
  double distance=std::sqrt(distance2);
  const double rdist = (distance-d0);
```

where `dmax_squared` is an additional class member, defined as `dmax*dmax`.

This will avoid the square root calculation whenever two atoms are not in range.

GiovanniBussi:
Thanks @ThorbenF !

PR #690, opened 2021-05-14
Removed some tests with numerical derivatives

GiovanniBussi:
##### Description

@valsson I would like to remove some tests that lead to numerical problems with the Intel compiler. Please let me know if you are OK with this.

Would it make sense to also remove other NUMERICAL_DERIVATIVES tests in the ves module? Usually these tests are useful while developing variables and functions, but are not necessary as regression tests.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


valsson:
Hello @GiovanniBussi, 

This would be fine for me. It would also be fine to remove the other `NUMERICAL_DERIVATIVES` tests in the ves regtests. It is true that this is only useful when developing. 

However, instead of deleting the relevant lines in the `plumed.dat` files, I would prefer to only comment them out. It is good to have it there for future development. 

I have done this in a different branch (https://github.com/valsson/plumed2/tree/ves-remove-num-deriv). I would suggest that I make a pull request for that branch and that we close this pull request. 

Regards,
Omar

GiovanniBussi:
Good point @valsson and thanks for preparing the other (more complete) request. I merged that one and I'll close this one.

PR #698, opened 2021-06-10
Fortran03 interface

GiovanniBussi:
This is an attempt to write a FORTRAN2003 module with interfaces for PLUMED calls. It is based on #653 (already merged) so as to allow the use of type and shape runtime checks.

I cc a set of people that (based on some search) could be using PLUMED from FORTRAN codes, in case they have comments or suggestions. In particular, I have no idea about which is the best way to *install* this module so that other codes can use it. I am also open to suggestions in the code, especially the FORTRAN part.

cc: @dvdesolve @aradi @danielhollas @praiteri @jppiquem @edoapra @marci73 @lscalfi

**Usage**

The interface is very similar to the classical FORTRAN interface. Here's a practical example:

```
PROGRAM main
  USE PLUMED_MODULE
  IMPLICIT NONE
  TYPE(PLUMED) :: p
  INTEGER :: i
  INTEGER :: natoms
  REAL(8), ALLOCATABLE :: positions(:,:)
  REAL(8), ALLOCATABLE :: forces(:,:)
  REAL(8), ALLOCATABLE :: masses(:)
  REAL(8) :: box(3,3)
  REAL(8) :: virial(3,3)

  ! this example shows how to run a command line tool from fortran
  p=plumed_f03_create()
  call plumed_f03_cmd(p,"CLTool setArgvLine","plumed --help")
  call plumed_f03_cmd(p,"CLTool run",i)
  call plumed_f03_finalize(p)

  ALLOCATE(positions(3,10))
  ALLOCATE(forces(3,10))
  ALLOCATE(masses(10))
  positions=0.0
  box=0.0
  virial=0.0
  forces=0.0
  do i=1,10
    positions(1,i)=1.0+10*i
    positions(2,i)=2.0+10*i
    positions(3,i)=3.0+10*i
  enddo

! global interface
  call plumed_f03_gcreate()
  call plumed_f03_gcmd("setNatoms",10)
  call plumed_f03_gcmd("init")
  call plumed_f03_gcmd("readInputLine","p: POSITION ATOM=2")
  call plumed_f03_gcmd("readInputLine","PRINT ARG=p.*")
  call plumed_f03_gcmd("setStep",1)
  call plumed_f03_gcmd("setPositions",positions)
  call plumed_f03_gcmd("setForces",forces)
  call plumed_f03_gcmd("setMasses",masses)
  call plumed_f03_gcmd("setBox",box)
  call plumed_f03_gcmd("setVirial",virial)
  call plumed_f03_gcmd("calc")
  call plumed_f03_gfinalize()

! object interface
  p=plumed_f03_create()
  call plumed_f03_cmd(p,"setNatoms",10)
  call plumed_f03_cmd(p,"init")
  call plumed_f03_cmd(p,"readInputLine","p: POSITION ATOM=2")
  call plumed_f03_cmd(p,"readInputLine","PRINT ARG=p.*")
  call plumed_f03_cmd(p,"setStep",1)
  call plumed_f03_cmd(p,"setPositions",positions)
  call plumed_f03_cmd(p,"setForces",forces)
  call plumed_f03_cmd(p,"setMasses",masses)
  call plumed_f03_cmd(p,"setBox",box)
  call plumed_f03_cmd(p,"setVirial",virial)
  call plumed_f03_cmd(p,"calc")
  call plumed_f03_finalize(p)

END PROGRAM main

```

Differences with respect to the classic FORTRAN interface:

- `use plumed_module` should be added to import the interface.
- `plumed_f_` becomes `plumed_f03_` to avoid confusion or mixing of the two interfaces.
- `char(0)` is not needed at the end of the command string nor at the end of the passed argument (it's implicitly added to strings). Actually, adding it as well would have no effect.
- `plumed_f03_gcmd("calc")` can be used without a dummy argument.
- The list of types that can be passed to `plumed_f03_gcmd()` is limited to `CHARACTER` (with `LEN=*`, and no dimension beside this), `REAL` (of kinds `c_float`, `c_double`, and `c_long_double`) and `INTEGER` (of kinds `c_int`, `c_short`, and `c_long`) as scalar or arrays up to rank 4. Notice that PLUMED currently only needs arrays up to rank 2, `float`, `double`, `int`, `long`, and `char`.
-  Type checks are performed at runtime (see #653). For instance, if you pass an `INTEGER` array to `setPositions` you receive an error.
-  Shapes are checked as well at runtime. **Notice that virial and box should be `(3,3)` and not `(9)`**.
-  In the object interface, a type `TYPE(PLUMED)` can be used to store the PLUMED instance. This changes slightly the form of some calls, making them more similar to their C counterparts (I actually don't think FORTRAN people use the object interface in practice). Anyway, these small changes are a good reason to change the function names to `_f03_`, so as to avoid compatibility issues.

**Implementation details**

This is obtained by adding:

- a few more wrapper functions in the `src/wrapper/Plumed.h`, meant to include type information, and
- a FORTRAN module in the `fortran/` directory. The FORTRAN module is generated with a python script, but the resulting code is included in the repository so it can be used straight away.

Currently, a code that needs to use this interface should first compile the `fortran/plumed.f03` module and then use it. I didn't find any standard way to compile and install a FORTRAN module file. I was thinking about two possible procedures:

- For codes that link PLUMED as a library (thus require PLUMED to be installed in advance) I could install the file `fortran/plumed.f03` somewhere, so that it can be accessed by a Fortran code. E.g. it could go into `prefix/lib/plumed/fortran/plumed.f03`. I guess this path could be stored in the `prefix/lib/plumed/src/lib/Plumed.*` files so as to be usable in a Makefile of CMake file, similarly to what is currently done with the path to the library.
- For codes that include PLUMED wrappers (`Plumed.h` and `Plumed.c`) and load PLUMED kernel at runtime (e.g.: AMBER), the `fortran/plumed.f03` file could be redistributed as well. This would allow also loading older PLUMED versions in the usual manner (setting `PLUMED_KERNEL`).

Notice that [dftbplus](	https://github.com/dftbplus/dftbplus/blob/87b00c77272efb44a854bc9749a4cac6f91de130/prog/dftb%2B/lib_extlibs/plumed.F90) implements its own wrapper class, which is a kind of alternative to my `fortran/plumed.f03` file, with more advanced features. @aradi if you are interested we could try to merge the efforts, or I can help you in adding compatibility with type checks in your own class. I can also try to make the typechecking more FORTRAN friendly, since it was designed with C++ in mind. However, any change to that would have to be included before releasing v2.8 and should be backward compatible.

**Testing**

If you want to test the code, this is what you should do:

```
git clone https://github.com/plumed/plumed2.git
cd plumed2
git checkout fortran03
./configure && make

cd fortran
# if you wish, edit main.f03
make FC=gfortran # put here the name of your FORTRAN compiler

./test

# alternatively, test runtime loading with:
export PLUMED_KERNEL=/path/to/libplumedKernel.so
./test-runtime
# this should be able to load older PLUMED versions as well
```

##### Target release


aradi:
@GiovanniBussi Dear Giovanni, yes I'd be interested to combine efforts. A few remarks:

- I'd suggest, that your f03-like functions call the C-functions directly, as we do in DFTB+. We would need a module which declares the interface of the C-routines using the [`bind(C)` mechanism](https://github.com/dftbplus/dftbplus/blob/a573e644cbe696abac7e509736f1c6bf7a07331a/prog/dftb%2B/lib_extlibs/plumed.F90#L50). That way, we won't need the horrible C-macros adding `_` suffices and convert things to lower case. Also, the burden of maintenance would become lower, as it would use the same interface directly, which Python and C also use.

- We should definitely NOT use `character(32)` to store a pointer! We should use the `type(c_ptr)` type, which correctly represents a pointer on the given system. Again, this could save the complicated and error prone bit arithmetic magic in `plumed_f2c()`.

- Does providing global routines makes sense at all? If you come up with a new interface, let's make it right from the beginning and ask people to instantiate a type. It would automatically make all applications instance-safe with respect of Plumed, even if they won't make use of it. (And you could still leave the old Fortran interface around for those, who do not want to change).

- As you already require Python for generating the Fortran source file, you may think about packaging the single file pre-processor [fypp](https://github.com/aradi/fypp) and use it to generate them from a [more readable template](https://github.com/dftbplus/dftbplus/blob/a573e644cbe696abac7e509736f1c6bf7a07331a/prog/dftb%2B/lib_extlibs/plumed.F90#L112). Fypp is used in several projects (fortran standard lib, CP2K) and would make mantainance probably easier. (Disclaimer: I am the main author of it...)

- I'd prefer to add an object-oriented interface to Plumed (as we do in DFTB+). There are a lot of recent scientific projects written in Modern Fortran, and I think, it would make sense to support this movement. You can still maintain your old F77 interface for all those, who want to stick with it, but the new one should use all the feature F2003 offers.

Maybe, we should set up a short video conference discussing the issues? I am very happy to help and contribute, so that it goes into the direction of a modern Fortran interface according to **current** Fortran best practices. :wink:

GiovanniBussi:
@aradi thanks a lot for your feedbacks!

> * I'd suggest, that your f03-like functions call the C-functions directly, as we do in DFTB+. We would need a module which declares the interface of the C-routines using the [`bind(C)` mechanism](https://github.com/dftbplus/dftbplus/blob/a573e644cbe696abac7e509736f1c6bf7a07331a/prog/dftb%2B/lib_extlibs/plumed.F90#L50). That way, we won't need the horrible C-macros adding `_` suffices and convert things to lower case. Also, the burden of maintenance would become lower, as it would use the same interface directly, which Python and C also use.

I see, my plan was to allow pre-FORTRAN03 compilers to work. Notice that technically the Python interface is calling C++ (through Cython), and C++ has manually defined mapping to C functions (e.g. `PLMD::Plumed::cmd()` maps to `plumed_cmd`). Similarly, the old-FORTRAN functions are manually mapped to the C functions. In my draft FORTRAN03 (included in this pull request), the FORTRAN03 functions were calling the old-FORTRAN functions. But I understand they could call directly the C functions. I am planning to add C functions with knowledge of the argument type in the next future, to allow type checking with C11 `_Generic`. These new functions might be callable from FORTRAN03 with `bind(C)` and allow for type checking.

> * We should definitely NOT use `character(32)` to store a pointer! We should use the `type(c_ptr)` type, which correctly represents a pointer on the given system. Again, this could save the complicated and error prone bit arithmetic magic in `plumed_f2c()`.

Also in this case, this was the only solution that I found to allow compilers with no knowledge of C interfaces to work. But I see that there's no point in continuing in this direction. Except for maintaining the old interface.

> * Does providing global routines makes sense at all? If you come up with a new interface, let's make it right from the beginning and ask people to instantiate a type. It would automatically make all applications instance-safe with respect of Plumed, even if they won't make use of it. (And you could still leave the old Fortran interface around for those, who do not want to change).

On this I am not sure. I checked on GitHub, and literally nobody is publicly using the FORTRAN interface with non-global routines. People working with FORTRAN are so used to global data that I think they would fill more confident with this. The duplication is very limited, since all the calls have exactly the same form. Also in C++ now, you can do `p.cmd()` or `gcmd()`.

> * As you already require Python for generating the Fortran source file, you may think about packaging the single file pre-processor [fypp](https://github.com/aradi/fypp) and use it to generate them from a [more readable template](https://github.com/dftbplus/dftbplus/blob/a573e644cbe696abac7e509736f1c6bf7a07331a/prog/dftb%2B/lib_extlibs/plumed.F90#L112). Fypp is used in several projects (fortran standard lib, CP2K) and would make mantainance probably easier. (Disclaimer: I am the main author of it...)

I would avoid also the dependence on Python actually (I prefer to store the f03 file indeed). These things tend to be in the way once you install on machines with different operating systems, or on containers where you need to install all needed components, etc. This is the `autoconf` philosophy (developers install `autoconf`, users only need `sh`), that I consider very practical.

> * I'd prefer to add an object-oriented interface to Plumed (as we do in DFTB+). There are a lot of recent scientific projects written in Modern Fortran, and I think, it would make sense to support this movement. You can still maintain your old F77 interface for all those, who want to stick with it, but the new one should use all the feature F2003 offers.

If it's possible to define constructors/destructors/assignment operator then probably we could do something in the spirit of the C++ interface.

> Maybe, we should set up a short video conference discussing the issues? I am very happy to help and contribute, so that it goes into the direction of a modern Fortran interface according to **current** Fortran best practices. 😉

Good idea! I will wait a few days to see if I can collect more feedback and then be back.

Since you seem expert in modern FORTRAN, I have a further question: which would be considered the best way to emulate C++ exceptions? Starting with PLUMED 2.5 we have a way to map exceptions thrown within PLUMED to exceptions thrown by the C++ wrapper. If there's something analogous to "throwing an exception" with FORTRAN we could do it as well.

Thanks again!

Giovanni

aradi:
> I see, my plan was to allow pre-FORTRAN03 compilers to work. Notice that technically the Python interface is calling C++ (through Cython), and C++ has manually defined mapping to C functions (e.g. `PLMD::Plumed::cmd()` maps to `plumed_cmd`). Similarly, the old-FORTRAN functions are manually mapped to the C functions. In my draft FORTRAN03 (included in this pull request), the FORTRAN03 functions were calling the old-FORTRAN functions. But I understand they could call directly the C functions. I am planning to add C functions with knowledge of the argument type in the next future, to allow type checking with C11 `_Generic`. These new functions might be callable from FORTRAN03 with `bind(C)` and allow for type checking.

Oh, I see, I was not aware that Python uses directly the C++ interface... The Fortran 2003 features are implemented in all actively developed compiler for several years by now. But probably it make sense to define, how many years backward should the new implementation be compiler compatible. And if people want to have more, they still could stick with the old Fortran interface, if you make the new one optional on build...  As for the `_Generic`. I am not sure, how you want to use it in C. It seems to be a macro (evaluated at compilation time), while Fortran would call C-API after of the compiled library, so only run-time checks could be made... (But, I am definitely not a modern C expert, so teach me if I am wrong.)

> > * We should definitely NOT use `character(32)` to store a pointer! We should use the `type(c_ptr)` type, which correctly represents a pointer on the given system. Again, this could save the complicated and error prone bit arithmetic magic in `plumed_f2c()`.
> 
> Also in this case, this was the only solution that I found to allow compilers with no knowledge of C interfaces to work. But I see that there's no point in continuing in this direction. Except for maintaining the old interface.

Exactly. We have now a compiler independent robust and interoperable solution.

> On this I am not sure. I checked on GitHub, and literally nobody is publicly using the FORTRAN interface with non-global routines. People working with FORTRAN are so used to global data that I think they would feel more confident with this. The duplication is very limited, since all the calls have exactly the same form. Also in C++ now, you can do `p.cmd()` or `gcmd()`.
> 

OK, I understand... (Maybe I anyway tend to educate people when designing interfaces :wink: )

> I would avoid also the dependence on Python actually (I prefer to store the f03 file indeed). These things tend to be in the way once you install on machines with different operating systems, or on containers where you need to install all needed components, etc. This is the `autoconf` philosophy (developers install `autoconf`, users only need `sh`), that I consider very practical.

I am with you on that. What I meant is that instead of the rather difficult to read `gen.py` one could use a readable template and a template engine to generate the code at the development phase. Latter could be fypp, as it is used all over now, or any other template generator able to deal with loops and some quirks of Fortran.

> If it's possible to define constructors/destructors/assignment operator then probably we could do something in the spirit of the C++ interface.

Yes, it is all possible. Question is, how much effort one should put into that. Note, that since we are communicating via the C-interface, one would have to write all the OO-mambo-jumbo for Fortran in Fortran once more. And had to maintain it extra. Probably, first, we should go for a robust, interoperable, type-safe interface for your next release, using a container type and procedures (optionally type bound procedures -- methods) to manipulate/query the container. Then, one could evaluate, whether further OO on the Fortran side is worth the effort or not.

> Good idea! I will wait a few days to see if I can collect more feedback and then be back.

Sure. You can also directly mail me on my [Uni Bremen address](https://www.bccms.uni-bremen.de/cms/people/b-aradi/).

> Since you seem expert in modern FORTRAN, I have a further question: which would be considered the best way to emulate C++ exceptions? Starting with PLUMED 2.5 we have a way to map exceptions thrown within PLUMED to exceptions thrown by the C++ wrapper. If there's something analogous to "throwing an exception" with FORTRAN we could do it as well.

That's a tricky one, as Fortran does not have exceptions (yet). (And even if it had, I usually prefer not to cross language borders with exceptions. And since we communicate via C, that would be difficult, anyway, I guess.)  But how do you handle the errors in C? We could just use that in the Fortran wrapper routines to detect errors, and maybe pass the error further up in the Fortran-part using either error codes or even better an allocated derived type (struct) with detailed error information.

GiovanniBussi:
>But how do you handle the errors in C?

When calling from C, the exception is not caught and the code crashes. When calling from C++, instead, if an exception is raised within PLUMED, we call a callback function that [stores a code](https://github.com/plumed/plumed2/blob/f017447899a99a4b474bb45f1b47bfa2b82e1075/src/core/PlumedMainInitializer.cpp#L77-L172) associated to the exception type and we then [rethrow the exception](https://github.com/plumed/plumed2/blob/master/src/wrapper/Plumed.h#L1254-L1318) from the C++ wrapper. It's a boring part of the code, but does not need any maintenance since it never changes. And works in difficult cases such as linking the C++ against a version of the standard library and PLUMED against another version.

> We could just use that in the Fortran wrapper routines to detect errors, and maybe pass the error further up in the Fortran-part using either error codes or even better an allocated derived type (struct) with detailed error information.

It would be relatively easy to do that in FORTRAN as well. A possible scheme would be:
```
call p%cmd("setPositions",incorrect_array) ! would fail, since the user is not checking the outcome

type (plumed_error) :: err
call p%cmd("setPositions",incorrect_array, error=err) ! would not fail, the user will be responsible for that
if err%code /= 0
then
  print err%code,err%message
  stop ! or something more clever, if possible
endif
```

But this is only worth if it would be perceived as natural from a FORTRAN point of view. I didn't implement anything like that in C because I thought there is no standard way of doing it.

aradi:
> It would be relatively easy to do that in FORTRAN as well. A possible scheme would be:
> 
> ```
> call p%cmd("setPositions",incorrect_array) ! would fail, since the user is not checking the outcome
> 
> type (plumed_error) :: err
> call p%cmd("setPositions",incorrect_array, error=err) ! would not fail, the user will be responsible for that
> if err%code /= 0
> then
>   print err%code,err%message
>   stop ! or something more clever, if possible
> endif
> ```
> 
> But this is only worth if it would be perceived as natural from a FORTRAN point of view. I didn't implement anything like that in C because I thought there is no standard way of doing it.

Well the "standard" way would be to pass integers, but I prefer derived types as it can carry also a human readable message. 
Maybe, the best is to use an allocatable one and check on its allocation status:
```
 type (plumed_error), allocatable :: err
call p%cmd("setPositions",incorrect_array, error=err) ! would not fail, the user will be responsible for that
if (allocated(plumed_error) then
   print *, err%code,err%message
end if
```
It had a few advantages over the static one, and one could even enforce it via a destructor (if one really wanted), that it stops the code, if the caller does not clear it explicitly...

Ps. By the way, the language we discuss about is now called **Fortran** (and not FORTRAN). :wink:

GiovanniBussi:
@aradi I managed to add error objects as this:
```
  call plumed_f03_cmd(p,"setStep",1.0,error=error)
  if(error%code /= 0) then
    print * , "ERROR:",error%code,error%what
    call plumed_f03_cmd(p,"setStep",1,error=error)
    if(error%code /= 0) then
      stop "should never arrive here"
    endif
  endif
```
I had some problem with the `allocatable` version that you suggested, it's in the TODO list anyway

GiovanniBussi:
I close this, superseded by #708

PR #703, opened 2021-06-18
Module for the computation of the solvent accessible surface area of proteins

andrea-arsiccio:
##### Description

I have added a new module (sasa) for the calculation of the solvent accessible surface area (SASA) of proteins using either the fast algorithm by Hasel et al. (Tetrahedron Computer Methodology Vol. 1, No. 2, pp. 103-116, 1988) or the LCPO algorithm (Weiser J, Shenkin PS and Still WC. J. Comput. Chem. 1999 (20), 217-230). The script could be used to include the SASA as a collective variable in metadynamics simulations, and also for implicit solvent simulations as described for instance in: "Arsiccio and Shea, Protein Cold Denaturation in Implicit Solvent Simulations: A Transfer Free Energy Approach, J. Phys. Chem. B, 2021". I have tested the script on a database of proteins, and used it with both Gromacs and Amber. I have also added a new regtest to validate changes.

##### Target release

I would like my code to appear in release v2.7 (actually I have no real preference, if you think the code is valuable please feel free to add it to any release of your choice at your earliest convenience)

##### Type of contribution

- [ x ] new module contribution or edit of a module authored by you

##### Copyright

- [ X ] the module I added or modified contains a `COPYRIGHT` file with the correct license information. Code should be released under an open source license. I also used the command `cd src && ./header.sh mymodulename` in order to make sure the headers of the module are correct. 

##### Tests

- [ X  ] I added a new regtest or modified an existing regtest to validate my changes.
- [ X  ] I verified that all regtests are passed successfully on [GitHub Actions](https://github.com/plumed/plumed2/actions).


GiovanniBussi:
@andrea-arsiccio thanks for this interesting contribution! I didn't check it yet carefully, but I have some comment:
- `plumed.getActionSet().select<SetupMolInfo*>();` The correct way to retrieve this in recent PLUMED is with `selectLatest` (see [here](https://github.com/plumed/plumed2/blob/43b4507cb81543b308d5ee021134b12441ac1f5a/src/generic/DumpAtoms.cpp#L247-L248)). This allows a user to include multiple MOLINFO, and every action uses the latest defined one.
- I noticed your tools needs to read data files (`SASAFile.open(path+"src/sasa/HASEL/hasel.dat");`). I am not sure these files would be installed. We could think about a way to install additional files for contributed modules. An alternative that might be simpler is to use `xxd`. Can you have a look at that? Check for instance [here](https://github.com/plumed/plumed2/blob/master/src/cltools/Makefile#L8-L9) and [here](https://github.com/plumed/plumed2/blob/master/src/cltools/Completion.cpp#L78-L81). An advantage would be that the implementation would work even if the resource directory is not installed.
- I've seen some compilation error now, please check the GitHub actions log

andrea-arsiccio:
Thank you so much for getting in touch, I am a huge fan of Plumed and
really thrilled to give my contribution.

Sure, I will try to address the points you raised as soon as possible, and
pull another request to Plumed.
Thanks and regards,
Andrea

Il giorno ven 18 giu 2021 alle ore 10:01 Giovanni ***@***.***>
ha scritto:

andrea-arsiccio:
Hello,
thanks a lot for your precious comments and suggestions.
I solved some issues I was having with MolInfo (I apologize, when I tested my code I was using plumed 2.4, so the script was not updated for the latest releases of plumed). Looking at the GitHub actions log it seems there may still be some issues with the namespace (sasa) I selected, but I am not sure why this is the case. Could you please give me a hint?

As for xxd, I did some tests, but I am unfortunately not an expert and I think I am not sure I understood perfectly how it works. I apologize about that, I will try again as soon as possible. Do you think reading from external data files would be problematic?

Thank you for your time and consideration.
Best regards,
Andrea

src/sasa/Sasa.h
```diff
@@ -0,0 +1,32 @@
...
+   misrepresented as being the original software.
+3. This notice may not be removed or altered from any source distribution.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ */
+#ifndef __PLUMED_sasa_Sasa_h
+#define __PLUMED_sasa_Sasa_h
+#include "core/Colvar.h"
+#include "config/Config.h"
+
+namespace PLMD {
+namespace Sasa {
```
GiovanniBussi, line 26: please use `sasa` instead of `Sasa`

GiovanniBussi:
- namespace should be `sasa`, not `Sasa`
- you should apply astyle. Do this:
```
cd plumed2/src
make astyle
git commit
```

Related to xxd, you should do this:
- put the extra file (e.g. `hasel.dat`) in the `src/sasa/` directory
- in the `src/sasa/Makefile` add the following lines at the end:
```
sasa_HASEL.cpp: hasel.xxd
hasel.xxd: sasa.dat
<TAB>cat ./hasel.dat | ../maketools/xxd > hasel.xxd
```
(make sure `<TAB>` is actually a tab space)
- In the `sasa_HASEL.cpp` file, in the constructor of the class put:
```
  static const char hasel_xxd [] = {
#include "hasel.xxd"
// cppcheck-suppress syntaxError
    , 0x00
  };
  std::istringstream hasel_stream(hasel_xxd);
```
At this point, `hasel_xxd` will be a single long string containing your file. You can add a printf to double check it. `hasel_stream` is a istringstream object from where you can read using `>>` as you are doing now.

An alternative to this xxd thing would be to hard code the parameters as C++ code as it is done for instance [here](https://github.com/plumed/plumed2/blob/master/src/isdb/SAXS.cpp#L568-L1632). I would say that the optimal thing depends on how you want to generate these parameters in case of an update, and which is the advantage of the text format you are currently using.

Do astyle again after every commit to make sure the files are properly formatted.

andrea-arsiccio:
Thank you very much again for your precious help and availability.

I have changed the namespace to sasa, and opted to hard code the SASA parameters as C++ code. I think this would make it easier for me to modify the code if needed in the future. 
Thank you and best regards,
Andrea

src/sasa/sasa_HASEL.cpp
```diff
@@ -0,0 +1,2716 @@
...
+}
+
+
+//reads input PDB file provided with MOLINFO, and assigns atom and residue names to each atom involved in the CV
+
+void SASA_HASEL::readPDB() {
+  auto* moldat = plumed.getActionSet().selectLatest<GenericMolInfo*>(this);
+  AtomResidueName[0].clear();
+  AtomResidueName[1].clear();
+
```
GiovanniBussi, line 243: Here moldat could be nullptr. I would suggest to give an error in this case.

src/sasa/sasa_HASEL.cpp
```diff
@@ -0,0 +1,2716 @@
...
+  auto* moldat = plumed.getActionSet().selectLatest<GenericMolInfo*>(this);
+  double Si, sasai, bij;
+  double sasa = 0;
+  vector<Vector> derivatives( natoms );
+  for(unsigned i = 0; i < natoms; i++) {
+    derivatives[i][0] = 0.;
+    derivatives[i][1] = 0.;
+    derivatives[i][2] = 0.;
+  }
+
```
GiovanniBussi, line 2514: Also here I would recommend a check on moldat before using it

src/sasa/sasa_HASEL.cpp
```diff
@@ -0,0 +1,2716 @@
...
+  setValue(sasa);
+  firstStepFlag = 1;
+  ++nl_update;
+  if (nl_update == stride) {
+    nl_update = 0;
+  }
+// setBoxDerivativesNoPbc();
+}
+
+}//namespace PLMD
```
GiovanniBussi, line 2715: I guess the two comments are swapped (close sasa namespace first)

src/sasa/sasa_LCPO.cpp
```diff
@@ -0,0 +1,2324 @@
...
+  setValue(sasa);
+  firstStepFlag = 1;
+  ++nl_update;
+  if (nl_update == stride) {
+    nl_update = 0;
+  }
+// setBoxDerivativesNoPbc();
+}
+
+}//namespace PLMD
```
GiovanniBussi, line 2323: I guess the two comments are swapped (close sasa namespace first)

andrea-arsiccio: You are definitely right! Sorry about that

src/sasa/sasa_HASEL.cpp
andrea-arsiccio: Right, I modified that. Thanks for the suggestion

src/sasa/sasa_HASEL.cpp
andrea-arsiccio: I added that. Thanks for the suggestion

src/sasa/sasa_HASEL.cpp
andrea-arsiccio: Thanks a lot for this comment. I added this.

GiovanniBussi:
Thanks! So, the last thing before merging is to add a description of the module. This is not very much documented, but it should be sufficient to add a single file names SASAMOD.md in user-doc. See e.g. [this](https://github.com/plumed/plumed2/blob/master/user-doc/FISSTMOD.md).

andrea-arsiccio:
Thank you so much again for your valuable comments! I have added a description of the module in user-doc. Please let me know if this could work.

GiovanniBussi:
Merged thanks for your contribution!!!

andrea-arsiccio:
Thank you so much, I was really happy to contribute! Hope this module will
prove useful for the Plumed users!
Regards,
Andrea

Il giorno lun 21 giu 2021 alle ore 21:09 Giovanni ***@***.***>
ha scritto:

valsson:
Hello, 

Just a short comment. I noticed that in the manual for the master branch, the page for this module is displayed in the root and not under modules (https://www.plumed.org/doc-master/user-doc/html/_s_a_s_a_m_o_d.html). Furthermore, the formatting for the example DeltaG file in https://www.plumed.org/doc-master/user-doc/html/_s_a_s_a__h_a_s_e_l.html seems slightly broken. 

Regards,
Omar

valsson:
You can use 

\verbatim
....
\endverbatim

For the example DeltaG file, then the formatting should be fixed

andrea-arsiccio:
Hello,
thank you very much, this is extremely helpful. I will fix the formatting
today and pull another request to the development version of plumed. I will
also adjust the references in the module description using \cite.

Sorry to bother you with one more question: how can I make the sasa module
appear under modules, rather than in the root?

Thanks and regards,
Andrea

Il giorno ven 25 giu 2021 alle ore 09:50 Omar Valsson <
***@***.***> ha scritto:

valsson:
Regarding the making the sasa module appear under modules, I honestly don't know. I quickly looked at your `SASAMOD.md` file and it appears fine to me, it appears similar to other module files (e.g. `PIVMOD.md`). Thus, I don't know why the manual generation tool doesn't pick it up as an additional module.

andrea-arsiccio:
Thanks. I modified the description and pulled a new request. Thank you very
much for letting me know about these formatting issues, hope they are
solved now.

Best regards,
Andrea

Il giorno ven 25 giu 2021 alle ore 10:16 Omar Valsson <
***@***.***> ha scritto:

PR #704, opened 2021-06-18
Fortran typesafe

GiovanniBussi:
### Fortran typechecks
Here's a second pull request, alternative to #698. Instead of designing a new set of functions, which is by itself a complicated task, I decided to add typechecks to the current one (`plumed_f_*`). The idea is the following:

1. Old Fortran code (not using this Fortran module) will continue to work after adding `use plumed_module`, provided it was correct.
2. New code written with `use plumed_module` will continue to work after removing `use plumed_module`.
3. Using `plumed_module` should only add extra safety, as listed below.
4. For all subroutines there is now an explicit interface. One cannot miss an argument or use too many arguments.
5. For subroutines `plumed_f_cmd` and `plumed_f_gcmd` there is additional runtime typechecking. This implies check on type, size, rank, and shape.
6. I made sure the code compiles with gfortran 10, that forbids calling the same function twice with different arguments.

To satisfy point 1, I didn't add a new `type :: plumed` and I didn't change function names to `plumed_f03`.

To satisfy point 2, I didn't add [exception handling](https://github.com/plumed/plumed2/pull/698#issuecomment-859659144) and I still require users to add `char(0)` after every string.

The new module is [here](https://github.com/plumed/plumed2/blob/fortran-typesafe/fortran/plumed.f90).
There's also a [regtest](https://github.com/plumed/plumed2/blob/fortran-typesafe/regtest/basic/rt-make-fortran/main.f90) now that tests simple fortran program.

@aradi could you please check if there's something that should be done differently? Main doubts:

- Installation. Currently, I "install" the `plumed.f90` file. Users of this are expected to recompile it with their own Fortran compiler. Is there a standard way to install Fortran modules instead?
- Compliance with Fortran standard. I realized we were passing `character(len=*)`, which is not legal. One should manually translate that to `character, dimension(:)` and than pass that. I thought that the approach we used to far always worked, and so it is convenient to continue in this way.
- Given the point above, using `bind(C)` would have been impossible. The C functions called now are still repeated with all the 6 possible name mangling.
- The code in [dftbplus](https://github.com/dftbplus/dftbplus/blob/87b00c77272efb44a854bc9749a4cac6f91de130/prog/dftb%2B/lib_extlibs/plumed.F90) uses keywords that I not know how to use (`contiguous`, `target`, `pointer`, `c_loc` ...). Is there something I should have done here as well (without breaking backward compatibility)?

Thanks a lot!

##### Target release


aradi:
> 
> @aradi could you please check if there's something that should be done differently? Main doubts:
> 

I think, this is OK as a first step, and you can not do much more using the old interface. So go ahead with it, but please, let also create a new alternative interface (co-existing with this improved one), which

* hides the fact, that the library is written in C++
* adds error handling (which a library always should)
* uses Fortran-types to represent instances (with automatic finalizers/destructors)
* passes C-pointers around, where needed

>   * Installation. Currently, I "install" the `plumed.f90` file. Users of this are expected to recompile it with their own Fortran
> compiler. Is there a standard way to install Fortran modules instead?

Fortran modules files are compiler (and sometimes even compiler version) dependent, similarly to C++ ABI. As there are several actively developed Fortran compilers out there, this is a real problem in the Fortran world. So, yes, probably the easiest solution is to distribute that one Fortran file, and each project can compile it with its own compiler.

>    * Compliance with Fortran standard. I realized we were passing `character(len=*)`, which is not legal. One should manually 
> translate that to `character, dimension(:)` and than pass that. I thought that the approach we used to far always worked, and so it 
> is convenient to continue in this way.

Actually, `character(len=*)` is not C-interoperable, indeed, because depending on the compiler, you may receive a pointer to a descriptor (which also contains the length of the string) instead of a pointer to the character data. So, by not defining a `bind(C)` interface, you do not solve the problem. (And you create for your users a very hard to debug problem, in case they are using the "wrong" compiler).

The solution is to create an interface to the C-routines with `bind(C)` and declare the arguments 
```
 character(kind=c_char), intent(in) :: key(*)
```
which is C-interoperable. Note, that Fortran allows to pass a scalar value to arrays of assumed size (unknown size) as above. So, a procedure with the dummy argument above, can be called with a `character(len=*, kind=c_char)` actual argument. You would not have to change your Fortran inteface, and would get nevertheless warranted C-operability.

>  * Given the point above, using `bind(C)` would have been impossible. The C functions called now are still repeated with all the 
> 6 possible name mangling.

But, as explained above, this is not needed as `bind(C)` is possible. :wink: 

>     * The code in [dftbplus](https://github.com/dftbplus/dftbplus/blob/87b00c77272efb44a854bc9749a4cac6f91de130
> /prog/dftb%2B/lib_extlibs/plumed.F90) uses keywords that I not know how to use (`contiguous`, `target`, `pointer`, `c_loc` ...). Is 
> there something I should have done here as well (without breaking backward compatibility)?
> 

No, you do not need attributes in your implementation. It is necessary in mine, because I am calling the same C-function in all my wrappers and pass only c-pointers around.

One last issue is the 32-byte handler/descriptor you pass around. I don't see, why this is warranted to work robustly on a 64-bit architecture, whichever magic you do with it. I know, that changing this to something else would break your current interface, but this would be a reason for me to advice people against using the current interface, if I were asked about it.

So, just to conclude: Add the `bind(C)` definitions and then this PR is IMO ready to go and you made your old Fortran interface as robust as possible, with the caveat that the handler passing is absolutely fragile by design. And then, let's create an optional new interface, which does this properly and can be used by new users of your library.

GiovanniBussi:
Thanks @aradi for the tip! It seemed to work. Now:
- C functions are all defined with `bind(C)`
- Their `char(len=*)` arguments have been changed to `char :: p(*)` as you suggested (see e.g. [here](https://github.com/plumed/plumed2/blob/e612096691895f90fe6fc0ac05586158f9566d23/fortran/plumed.f90#L23-L28)). This trick seemed to work.

For some reason I am not able to use the same trick on the Fortran functions. I mean, if [here](https://github.com/plumed/plumed2/blob/e612096691895f90fe6fc0ac05586158f9566d23/fortran/plumed.f90#L413-L420) I try to use `character(kind=c_char), intent(in)    :: p(32)` the compiler complains:
```
Error: There is no specific subroutine for the generic 'plumed_f_cmd' at (1)
```
It looks like it is not able to resolve the overload. The reason to do this would have been to change the documentation and suggest people to directly use `character :: p(32)` as a handle, which would have worked equivalently both with `use plumed_module` and without. Anyway, it is not so important I guess.

I also took this sentence literally:
> Fortran allows to pass a scalar value to arrays of assumed size

and I thought I can defined a single C function that accepts both a scalar and an array using a (*) argument in its prototype. But I then get this error:
```
Error: Rank mismatch in argument 'val' at (1) (rank-1 and scalar)
```
The solution I found for this is to have one C function that accepts a scalar (by reference) and another that accepts a (*) (see e.g. [here](https://github.com/plumed/plumed2/blob/e612096691895f90fe6fc0ac05586158f9566d23/fortran/plumed.f90#L200-L217)). This duplicates the number of C functions, but it's fine with me if there's not an easier solution.

Finally, related to the 64 bit question, notice that this `character(32)` contains actually 8*32=256 bits, so it is highly conservative! The infamous `plumed_c2f` [function](https://github.com/plumed/plumed2/blob/7fb5a99e706e763c99b7b239f3b9c4c7fd9ff76b/src/wrapper/Plumed.h#L3176) "prints" the pointer on a string with some convention. I did this to make sure the string only contains printable characters, with the doubt that a smart Fortran implementation could be fooled by zeros appearing in the middle of the string. Maybe I could use hexadecimal format for the string, so that people could interpret it more easily.

I will come back with more questions on the other thread, where I will try to make a better "modern Fortran" interface.

Thanks again!

aradi:
>     * C functions are all defined with `bind(C)`

Cool! :fireworks: 

> For some reason I am not able to use the same trick on the Fortran functions. 
> 
> ```
> Error: There is no specific subroutine for the generic 'plumed_f_cmd' at (1)
> ```
> 

No, if a generic overloaded name is used, the rules are stricter to ensure disambiguity.

> and I thought I can defined a single C function that accepts both a scalar and an array using a (*) argument in its prototype. But I then get this error:
> 
> ```
> Error: Rank mismatch in argument 'val' at (1) (rank-1 and scalar)
> ```

Yes, the trick only works with character variables, where a scalar character variable is considered as it it were an array of size one.

Actually, since Fortran 2018, you can make proper rank-agnostic C-bindings, such as
```
subroutine plumed_f_gcmd_safe_int(key, val, pass_shape) bind(C)
      import
      character(kind=c_char), intent(in)    :: key(*)
      integer(kind=c_int)                           :: val(..)
      integer(kind=c_size_t) :: pass_shape(*)
    end subroutine plumed_f_gcmd_safe_int
  end interface
```
However, I have no experience how many compilers support it reliably, as I did not use this feature myself yet.

> 
> Finally, related to the 64 bit question, notice that this `character(32)` contains actually 8*32=256 bits, so it is highly conservative! 

That's my bad, not being able to distinguish between bits and bytes, sorry. It should be safe enough, indeed, but still an ugly hack. (But let's fix and hide that in the modern Fortran interface...)

GiovanniBussi:
> I am wondering, why you need the `asynchronous` attribute here. Do you plan to change the content of `val` also once the routine call is over? Or does `val` becomes part of an asynchronous I/O operation? (Also it is a Fortran 2018 construct, so I am not sure, which compilers have it already implemented...)

Thanks! Indeed I wanted to double check with you. I've seen this used in the MPI interface, which is quite similar to ours. In particular, when you do `cmd("setForces",forces)` you expect the forces array to be modified after the function returns. and with `cmd("setPositions",positions)` you expect the forces array to be read after the function returns. I noticed that this gives a compilation error when passing e.g. `positions((/1,2,3/),:)` that is a temporary copy of the positions. In that case, if I don't use the module, the code compiles and gives random results.

For portability, it works with gfortran 4 (on [Centos7](https://github.com/plumed/plumed2/actions/runs/963274929)).

I was checking if there's something that can be done to check for user mistakes, but if you think `asynchronous` might be non portable I'll remove it.

PR #710, opened 2021-06-28
Fix incorrect EXPANSION module in manual

valsson:
##### Description

This fixes an issue with the additional modules where at the moment there is a listing for `EXPANSION` that is incorrect (between VES and MAZE in the list on https://www.plumed.org/doc-master/user-doc/html/_add_mod.html). 

It seems that this is due to the usage of `EXPANSION_CV` in the OPES manual. I have changed this to `OPES_EXPANSION_CV` in the relevant places and this seems to fix the issue (as tested on my laptop). 

@invemichele should check that he agrees with these changes. 

##### Target release

I would like my code to appear in release v2.8 (master)

##### Type of contribution

##### Copyright

### DIFF
--- a/src/opes/ECVcustom.cpp
+++ b/src/opes/ECVcustom.cpp
@@ -20,7 +20,7 @@ along with plumed.  If not, see <http://www.gnu.org/licenses/>.
 namespace PLMD {
 namespace opes {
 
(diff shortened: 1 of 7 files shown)


invemichele:
I didn't realize there could be this kind of issue! thanks Omar for spotting and correcting it

PR #713, opened 2021-07-02
minor improvements to OPES_METAD

invemichele:
##### Description

Several minor improvements to the OPES module that gathered up over time, such as:
- better doc
- work is now defined as in METAD

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@invemichele it looks this is breaking backward compatibility. Are you sure you want to include this in v2.7? If so, please add a comment [here](https://github.com/plumed/plumed2/blob/v2.7/CHANGES/v2.7.md) (just start a new header for v2.7.2, in case we can fix the formatting later). Thanks!

invemichele:
Are you referring to the fact that some regtests have been changed? 
Yes, it's true, but the changes only concern the definition of the output `work` (which is optional and rarely used). I realized that  I was printing the work done in one PACE and so I changed to print instead the accumulated work, to be consistent with what METAD does.

I consider this a minor change and I think it would be better to have it in v2.7, just to avoid any possible confusion.

GiovanniBussi:
Thanks! This is useful so that if users ask questions about version 2.7.1 or 2.7.2 we will know what to answer.

PR #722, opened 2021-07-19
Make sure incorrect conversions trigger an exception

### LINKED ISSUE #717: Issue with GRID_MIN/GRID_MAX in HISTOGRAM action
By chance, I noticed an issue with `GRID_MIN/GRID_MAX` in `HISTOGRAM` action.

It is possible to give a random string as a value to `GRID_MIN/GRID_MAX` and the code still runs without an error. Instead, it seems to take 0.0 as the grid value. 

For example, you can see this in `regtest/analysis/rt-histo`: 

If you use 

HISTOGRAM ...
  ARG=x
  GRID_MIN=0.0
  **GRID_MAX=blabla**
  GRID_BIN=ffa
  BANDWIDTH=0.1
  LABEL=hA1
... HISTOGRAM

Gives 

#! FIELDS x hA1 dhA1_x
#! SET normalisation    1.0000
#! SET min_x 0.0
**#! SET max_x blabla**
#! SET nbins_x  100
#! SET periodic_x false
   0.0000   1.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000
   0.0000   0.0000   0.0000

I guess it would be better to add a check on either the parsing of GRID_MIN/GRID_MAX that the parameters gives are proper numbers. Or maybe better in the `setBounds` function in `gridtools/HistogramOnGrid.cpp` to avoid this problem happening elsewhere.

GiovanniBussi:
This is an attempt to fix #717. I am applying it directly to version 2.8 but it might be backported.

##### Description

**OUTDATED, see below**

I modified the way the functions `Tools::convert()` work. Before this change, they were returning a bool (false meant an error). This is a poor design, since the user might ignore the error easily, and was used since in the first PLUMED 2 releases we were not using C++ exceptions yet.  After this change, these functions return void and raise an exception in case of an error. This makes it more difficult to inadvertently ignore conversion errors.

Basically, these modifications required me to modify all the places where the return value was correctly checked. All these places were immediately detectable as compilation error since the return value of these functions was accessed. For instance, something like this:

```
if(Tools::convert(a,b)) do_something();
else do_something_else();
```
now would not compile unless it is converted to:
```
try {
  Tools::convert(a,b);
  do_something();
} catch(ExceptionConversionError& exc) {
 do_something_else();
}
```

In fixing the code I had to adjust some checks in VES and OPES (cc: @valsson and @invemichele).

In addition, I added a workaround to the code since there's a [place](https://github.com/plumed/plumed2/blob/8f0d06287967946b2bb7a100fab94cf26203e69e/src/gridtools/GridVessel.cpp#L95-L108) where conversions were not checked and should not be checked to pass current regtests. More precisely, the original code was *not* checking the returned value, and, consistently, the new code ignore any exception triggered here. @gtribello please check if the fix is ok or if there's something else to do here. In particular, I think that to pass test `multicolvar/rt-dens-average` it would be sufficient to allow for empty strings (rather than to allow for arbitrary strings as it is now).

##### Target release

I would like my code to appear in release __v2.8__ but could be backported

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Note: there's something wrong with GitHub actions tests on Linux, I have to check them...

GiovanniBussi:
After having seen that there is a performance penalty in catching exception, I made an alternative implementation where exceptions are only used when needed. In the new implementation:
- `convertNoexcept(a,b)` is identical to the old `convert`. It returns a `bool` that one is then expected to check manually. This can be used in place where we were correctly checking the returned value
- `convert(a,b)` basically calls `plumed_assert(convert(a,b))`. It will automatically be used anywhere we do not check the output

The advantage with respect to the previous implementation is that we only throw exceptions when the is a non-detected error. We can still repeatedly call `covertNoexcept` to check if a given number can be converted without performance penalties.

The changes to the code are also simpler since there's no need to add `try/catch` statements.

@gtribello there is still [this point](https://github.com/plumed/plumed2/blob/a773ab62aed2888bdccd6ef3fea76498443b1ff8/src/gridtools/GridVessel.cpp#L95-L99) to double check. In order not to break a test, I am ignoring the result here.

PR #739, opened 2021-09-10
Let GROMACS runtime logs show PLUMED patch active

### LINKED ISSUE #737: allow GROMACS log files to report that the code is patched by PLUMED
In GROMACS, we have for a few years enabled users to do `cmake -DGMX_VERSION_STRING_OF_FORK=PLUMED` and that will show up in the GROMACS version string e.g. in log files. This is useful for users running their own forks making sense of their log files later on, as well as for developers reading forum posts or bug reports knowing what versions were used. Troubleshooting GROMACS users' performance questions can go much better if the co-existence of PLUMED is known.

In PLUMED's approach, it seems straightforward to patch the relevant cmake file in GROMACS that declares that variable so that its default value is `PLUMED` and the GROMACS mechanism will work the same way. Is such a patch to PLUMED welcome? If so, I will prepare it.

Would the PLUMED developers prefer also that the PLUMED version appear there? If so, is there a mechanism for that to be updated each time the PLUMED version is updated?

mabraham:
##### Description

Knowing which PLUMED version has patched GROMACS is potentially valuable information for GROMACS and PLUMED users and developers troubleshooting their runs.

Closes #737

##### Target release

I would like my code to appear in release 2.7 and master

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/patches/gromacs-2021.config
+++ b/patches/gromacs-2021.config
@@ -29,5 +29,8 @@ EOF
 
 plumed_before_patch(){
   plumed_patch_info
+  
+  mv cmake/gmxVersionInfo.cmake cmake/gmxVersionInfo.cmake.preplumed
+  awk -v version="$PLUMED_VERSION" '/^set\(GMX_VERSION_STRING_OF_FORK/{gsub(/""/, version)}; {print}' cmake/gmxVersionInfo.cmake.preplumed > cmake/gmxVersionInfo.cmake
 }
 


mabraham:
@GiovanniBussi Here's my attempt at patching the latest GROMACS config. Ideally we'd put a pair of "" around the version, but I couldn't work out how to do that with awk. So long as the PLUMED_VERSION doesn't contain spaces then the current code is OK. Once we're happy with this, then porting to the other configs and perhaps back to the 2.7 branch would be desirable.

Only recent GROMACS versions have this GMX_VERSION_STRING_OF_FORK variable, but if it's not found then the awk operation is a no-op, so we don't need extra complexity to handle that.

GiovanniBussi:
Thanks @mabraham !

I'll merge and backport to v2.7 and to gmx 2020 patch (the modified file seems already present [there](https://github.com/gromacs/gromacs/blob/release-2020/cmake/gmxVersionInfo.cmake)).

mabraham:
> @mabraham I added a "plumed-" prefix to the string, does it make sense?
> 
> See [9b50841](https://github.com/plumed/plumed2/commit/9b50841c8d8cc0801740a104d1502a1db3561363)

Sure, that's good!

mabraham:
> Thanks @mabraham !
> 
> I'll merge and backport to v2.7 and to gmx 2020 patch (the modified file seems already present [there](https://github.com/gromacs/gromacs/blob/release-2020/cmake/gmxVersionInfo.cmake)).

Yes, present since 2020

GiovanniBussi:
@mabraham is it possible that GROMACS uses the modified version to download the set of regression tests? I am asking because a user reported [this](https://groups.google.com/d/msgid/plumed-users/a9a3b8b8-b8c4-4fd2-91df-02313f32ffbbn%40googlegroups.com?utm_medium=email&utm_source=footer) on our mailing list. Thanks!

mabraham:
Thanks for the question. Yes, that is what is happening. Fix proposed for
2022.1 at https://gitlab.com/gromacs/gromacs/-/merge_requests/2587.

Unfortunately the issue won't be fixed in 2020 or 2021 branches (out of
support and not serious enough, respectively), so I suggest you suggest
users do a similar fix themselves, or put such a fix into the PLUMED patch,
or set a correct REGRESSIONTEST_VERSION value in the cache (which will not
be overridden by the current buggy logic).

Mark

mabraham:
That fix is now incorporated in the branch from which 2022.1 will be made, in due course, so the fix is expected to appear there.

PR #749, opened 2021-11-02
Handling exceptions in Fortran08 interface

GiovanniBussi:
##### Description

This is an attempt to include exceptions handling in the modern Fortran interface (see #708).

In this rewrite, the two implementations (classic and modern) are separated so that they do not depend on each other anymore. The classic one (see #704) only provides an explicit interface and type checks for the Fortran77-like interface. The modern one has now a number of improvements:

- Can handle errors and provide more info about the passed array.
- Internally, it uses a C structure (instead of `character(len=32)`) (this is not visible from outside).

Here a sample code:
```
type(plumed) :: pl
type(plumed_error) :: pl_error

! this would just crash (passing a real instead of a int)
call pl%cmd("setNatoms",4.0)

! this allows handling the exception instead:
call pl%cmd("setNatoms",4.0,error=pl_error)
if(pl_error%code!=0) then
  write(0,*) pl_error%code,pl_error%what
endif

! this will fail when calling "calc", since forces should be modifiable
call pl%cmd('setForces",forces,const=.true.)

! this will fail immediately, since PLUMED wants to store a copy of this pointer for later usage
call pl%cmd('setPositions",positions,nocopy=.true.)

! the two last cases could be handled passing a plumed_error object

```

@aradi I feel confident in the code, but if you want to have a look I would be happy :-) I plan to merge in a week or so. Critical points might be:
- The use of a [dummy type](https://github.com/plumed/plumed2/blob/fortran-exceptions/fortran/plumed_f08.f90.template#L17-L19) to enforce keyword-only arguments (e.g. [here](https://github.com/plumed/plumed2/blob/7e25ee16c39ffa130d1844b416a98dad3b33811a/fortran/plumed_f08.f90.template#L259)).
- [This callback function](https://github.com/plumed/plumed2/blob/7e25ee16c39ffa130d1844b416a98dad3b33811a/fortran/plumed_f08.f90.template#L226-L251) that manipulates the error message.
- The syntax for the additional arguments (`const` and `nocopy`, see e.g. [here](https://github.com/plumed/plumed2/blob/7e25ee16c39ffa130d1844b416a98dad3b33811a/fortran/plumed_f08.f90.template#L261-L262)).

In case you see something that might be useful to add or change, this is the best timing since we are close to releasing PLUMED 2.8.

Thanks a lot in advance!

##### Target release


aradi:
Dear Giovanni, this is a very nice effort to make the library robust! After having a quick glance over the implementation, I have one question and one comment.

* Question: Why do you need a callback function to populate the error structure? Theoretically, you could also populate it at the C/Fortran interface, as the C part has all error info, I guess. Nothing wrong with it, I am just curious about the motivation behind this decision.
* Comment: Unfortunately, the current interface still does not cure one of the big flaws of the original interface: You can pass an expression, where actually a variable (a pointer) would be needed, because you have no `intent` attribute on the value. 

This toy example demonstrates this, the second call would make the program crash:
```fortran
module test
  implicit none

contains

  subroutine main()
    integer :: ii

    ii = 5
    call routine_without_intent(ii)
    call routine_without_intent(1)

  end subroutine main


  subroutine routine_without_intent(ii)
    integer, asynchronous :: ii

    print *, "Got:", ii
    ii = ii + 2  ! This crashes, if you pass an expression instead of a variable

  end subroutine routine_without_intent

end module test


program test_program
  use test

  call main()

end program test_program
```
This is why I added two different kind of plumed calls in DFTB+ (although I made a stupid typo there, now that I look once more over it): one, which declares an `intent(in)` for values, and `intent(inout)` for pointers/references. Similarly, it would be much safer to have two different interfaces:
*  `pl%cmd_val(..., val,... )` with `val` having the `target, intent(in)` attribute. Here, user can safely pass expressions or constants, and that would be checked at *compile time*.
* `pl%cmd_ref(..., val, ...)` with `val` having the `target, intent(inout)` attribute. This would force the caller to only pass modifiable objects. So, if you try to call `pl%cmd_reference(..., 1, ..)`, the compiler will catch it! Also, you would not need the optional arguments, which again only ensure run-time checking instead of compile-time checking. (And, of course, it should be documented for every plumed command, whether it needs a value or a pointer/reference, so users know which type to call, but probably that is already).
I think, one should make as many checks as possible at compile time, and by choosing two different interfaces (instead of operating with optional arguments on an interface which is to generic) one can enforce this.

The same for the `nocopy`. I would rather introduce a third interface `pl%cmd_ptr` where `val` has the `pointer, contiguous, intent(in)` attribute. This would ensure, that no copy in/copy out is made when the routine is called (and that the pointer does not represent a non-contiguous stride). So people would have to add at least the `target` attribute to the objects the pass to that routine. (Without that, anything could happen.) Again, the same intention as with the optional attribute, but with compile-time checks instead of run-time ones.

GiovanniBussi:
@aradi thanks for the quick feedback.

**Regarding the callback function.**

This is required to manage resources (in this case, allocating a string) from the calling language (in this case, Fortran), but the same is true even when calling PLUMED from C++. This is the most flexible way to make sure it will be extensible in the future with the only overhead of passing two pointers (one to a function and one to a context). The function is only called when an exception is raised.

**Regarding const and nocopy.**

Notice that there are two consistency checks to do:
- Const/nocopy arguments vs what plumed really does with the pointer
- Declared intent vs const/nocopy arguments

The first can only be done at runtime, and will raise an exception.

The second is now done at runtime. You would suggest to do the second check at compile time. This would be very useful. In C++, it's already done at compile time by using overloads (I mean: the compiler sends const-ness information to PLUMED based on const-ness of the argument). In Fortran I think you cannot resolve overloads based on intent, so it looks that what you suggested is the only way to go. I could do it as follows:

- `cmd_val(...)` with `intent(in)`, calls `cmd(...,nocopy=.true.,const=.true.)`
- `cmd_ref(...)` with `intent(inout)`, calls `cmd(...,nocopy=.true.)`
- `cmd_ptr(...)` with `pointer, contiguous, intent(inout)` calls `cmd(...)`

and add the fourth case:
- `cmd_cptr(...)` with `pointer, contiguous, intent(in)` calls `cmd(...,const=.true.)`

Would this work out?

I am not sure about the `asynchronous` thing. Should it be there?

Thanks again!

aradi:
> In Fortran I think you cannot resolve overloads based on intent, so it looks that what you suggested is the only way to go. I could do it as follows:
> 
>     * `cmd_val(...)` with `intent(in)`, calls `cmd(...,nocopy=.true.,const=.true.)`
> 
>     * `cmd_ref(...)` with `intent(inout)`, calls `cmd(...,nocopy=.true.)`
> 
>     * `cmd_ptr(...)` with `pointer, contiguous, intent(inout)` calls `cmd(...)`
> 
> 
> and add the fourth case:
> 
>     * `cmd_cptr(...)` with `pointer, contiguous, intent(in)` calls `cmd(...,const=.true.)`
> 
> 
> Would this work out?

Oh, that's a really nice solution. That means, by choosing the given form of the `cmd_*` call, the compiler makes sure that you only pass arguments, which are compatible with the choice. And at runtime, Plumed would check, whether the chosen `cmd_*` call form was the correct one for the given command.

What is actually the use case for the 4th one? Would Plumed store the pointer and read it later but not modify the content?

Note, that on the Fortran side, the interfaces for `cmd_ptr` and `cmd_cptr` would be identical, as Fortran does not have the concept of pointer constantness. Attribute `pointer, intent(in)` means, that you do not change the target (the address the pointer points to), which you don't do in any of the cases, It does not mean, that you do not change the content of the pointer target. If you specify `pointer, intent(inout)`, it would allow the routine to change the pointer association (let it point to a different target or getting deallocated), which you don't do (and should not do). Additionally, it would make the users life more complicated, as they would not be allowed to pass a variable with a `target` attribute any more (that only works with `intent(in)`), but they would have to pass real pointers (which they should not have to deal with).

Consequently, if the 4th case is really needed (is it really used?), you could in theory merge it with the 3rd and use the optional argument `const` to distinguish between the two use cases, since you can not distinguish between them at compile-time anyway. But,  it's a matter of taste, one can live with both solutions. Only, if you keep it as a separate call (in the hope, that in the far future Fortran introduces const-pointers, and then the signatures could indeed become distinguishable), you may consider to choose a different suffix, as , as I immediately associated `cptr` with `c_ptr` and thought the routine would expect a `type(c_ptr)`  argument...

As for the `asynchrous` attribute: It tells the compiler, that the content of the variable may change unexpectedly. Changing the content in a call, where the variable is passed as argument, is for sure not unexpected. If the content of the variable changes otherwise, e.g. `%cmd("calc",...)` changes the content of a variable to which a pointer was passed earlier, that would be a possible case for `asynchronous`. But then, it should be defined in the caller scope as `asynchronous`, not in the interface:
```
real(dp), allocatable, target, asynchronous :: derivs(:,:)
...
call plumedcalc%cmd_ptr("setForces", derivs)
...
! At this point, value of derivs suddendly changes, although it was not passed,
! so probably it should have the asynchronous attribute in this scope
call plumedcalc%cmd_val("calc", 0)
...
```
So, I think you can leave away it from your wrappers, and users should specify it in the caller scope, if they want to be pedantic and 100% sure, the compiler does not do anything funny with their arrays between the calls.

aradi:
@GiovanniBussi Let me know, when you are finished with all the changes (especially the implementation of the various `cmd_*` call forms). As that would be then already very close to our current implementation, I could give a quick try, how it works in DFTB+ with various compilers, before you merge it to the main branch.

fortran/plumed_f08.f90
```diff
@@ -2015,7 +2015,9 @@ end subroutine pl_cmd_ref_integer_0_0
     subroutine pl_cmd_ptr_integer_0_0(this,key,val,dummy,error)
       class(plumed),                 intent(inout) :: this ! inout to allow for initialization
       character(kind=c_char,len=*),  intent(in)    :: key
-      integer(KIND=c_int), pointer,  intent(in) :: val
+      ! workaround for bug in Intel 19 compiler (pointer => target)
+      ! https://github.com/dftbplus/dftbplus/commit/a9329e72791a224b005df70e7fe59dfb26b92a75
+      integer(KIND=c_int), target,  intent(in) :: val
```
aradi, line 2020: A few remarks on this:
* Indeed, unfortunately ifort 19 has problems with the original (though correct) formulation
* If you change `pointer` to `target`, you also must change `intent(in)` into `intent(inout)` as you change the value of the object (This was the typo, I mentioned earlier, which I had in my plumed interface in DFTB+)
* Although, the workaround is probably the closest you can get to the original formulation, they are not equivalent! The workaround does not force the caller to declare the `target`  attribute in the caller scope for the actual argument for `val` any more. Buy forgetting to do so, there is no warranty, that the compiler does not make a copy-in/copy-out when entering the `pl_cmd_ptr` routine. So, in theory, it could happen, that plumed obtains and stores a pointer to the address of a temporary object, which does not exist any more, when plumed tries to read or write it later. I am not aware of any compilers, which would do so (and it would be quite illogical), still it would be standard conforming to do so. :disappointed: 

I think, we need to add a few warnings/recommendation to the Fortran-interface somewhere in the docs. One for the `asynchronous` and one for the `target` attribute for the variables in the caller scope, when `pl_cmd_ptr` or `pl_cmd_const_ptr` is invoked.

GiovanniBussi: @aradi is there a standard preprocessor macro that can be used to have specific code for intel compiler? Was this bug submitted to intel already? Notice that on GH Actions we are using the latest intel compiler, so this is not just intel 19.

aradi: No, the preprocessor is agnostic and does not know anything about the compiler which processes its output. The build framework should pass an appropriate define (e.g. `-DINTEL_FIX`) to the preprocessor, if it uses an Intel compiler, and then within the code one could do in `#:if`:
```fortran
#:if defined('INTEL_FIX')
  ...  ! target version
#:else
  ... ! pointer version
#:endif
```

As for reporting to intel, we did not report so far. Actually, I've just tried now to set up a minimal working example, and that actually works with Intel! I'll try a bit more, whether I can distill the problem. In case, you want to play around with it, my toy example was the following:
```
module test_target
  implicit none

  interface subptr
    module procedure subptr_integer, subptr_real
  end interface subptr

contains

  subroutine main()

    integer, target :: intval
    real, target :: realval

    intval = 42
    realval = 4E2

    call subptr(intval)

  end subroutine main


  subroutine subptr_integer(val)
    integer, pointer, intent(in) :: val

    print *, "INTEGER PTR obtained", val

  end subroutine subptr_integer


  subroutine subptr_real(val)
    real, pointer, intent(in) :: val

    print *, "REAL PTR obtained", val

  end subroutine subptr_real

end module test_target


program test_target_progam
  use test_target
  implicit none

  call main()

end program test_target_progam
```

GiovanniBussi: >If you change pointer to target, you also must change intent(in) into intent(inout) as you change the value of the object (This was the typo, I mentioned earlier, which I had in my plumed interface in DFTB+)

I didn't make this, and still the code is working. Is this expected?

You can see a sample code in the [regtest](https://github.com/plumed/plumed2/blob/576aca10897c539cbf7b4fc3a561c1770a65bf0a/regtest/basic/rt-make-fortran08/testfull.f90)

GiovanniBussi: And: shall I use `intent(in)` for `cmd_const_ptr` and `intent(inout)` for `cmd_ptr`?

GiovanniBussi:
@aradi I think the code is basically working (except for the intel bug above). A few things that could be made, but none of them is needed at this stage:
- Cleanup the code. There are currently repetitions that might be avoided using more fypp.
- Implement character arrays. Something that would make sense is using arrays to increase the rank, that is a `character(len=:), allocatable :: a(:)` would be rank 2. Not needed now, as there's no use case within plumed.
- Add `cmd_ref` `cmd_ptr` and `cmd_const_ptr` for characters. Not needed now, as there's no use case within plumed.

So, if you want to give it a try within DFTBplus I will then merge it to PLUMED master branch.

Thanks again!

fortran/plumed_f08.f90
aradi: OK, I managed to create a minimal working example, and sent a [bug report](https://community.intel.com/t5/Intel-Fortran-Compiler/Compiler-bug-target-array-can-not-be-passed-for-pointer-argument/m-p/1327248#M158240) to the Intel forum.

aradi: Yes, the code works irrespective of the `intent`. Since you pass the address of the object to something else, it does not really matter. Actually, now that I think once more about it, since you do not change the value of the object within the call itself, probably `intent(in)` is more appropriate, so leave it that way. What is more important, that the caller sets the `asynchronous` attribute in its scope in the non-constant case.

GiovanniBussi: > you do not change the value of the object within the call itself
This is not guaranteed in the `cmd_ptr` case. So I will use `intent(inout)` for that.

I mean: one should be able to use `cmd_ptr` in any place where it is possible to use any other `cmd_*` call, since it is the least restrictive one on the PLUMED side. So we cannot assume that PLUMED will not change the target.

aradi: I see, that makes sense, indeed.

However, if possible, I would recommend to force the user to choose the most restrictive calling form. Calling `cmd_ptr` and `cmd_const_ptr` should **only** be done (and possible), if plumed stores a pointer and postpones the reading or writing. The caller is supposed to make a lot of extra gimics int those cases (`asynchronous`, `target`, etc.), which we can neither enforce nor check (either due to compiler bug or due to the lacking support in the Fortran standard). Therefore, the resulting code could in theory fail, and finding the bug would be a nightmare (try to find out that the segfault was due to the access to a non-existing temporary array...) In all other cases, when the action on the argument from Plumed happens only within the call, one should not allow the `cmd*ptr` calls, but force the user (via your run-time checking mechanism) to use `cmd_val` or `cmd_ref` instead. (By checking somewhere, whether the `nocopy` and eventually the `const` attributes were passed, and stop if they were not -- if that is possible on the C++ side. Then one given command had exactly one possible call).

aradi:
@GiovanniBussi I updated DFTB+. Currently I stumbled upon, that `setPositions`, which I use to pass the atom positions to Plumed, does not work with `cmd_val`. Is it supposed to be like that. Can I look it up somewhere, which keyword requires which kind of call / constantness?

Also a suggestions. Calling the new f08 interface module `plumed_f08_module` would feel much more natural as `plumed_module_f08` and would follow a convention seen in several projects (adding the `_module` suffix to module names).

Finally, a remark: the fact that the plumed_f08 module is not compiled during installation (or did I just mess it up?), but it must be added to the users code and compiled there, could be very critical, if the C-interface ever changes. Projects will just distribute this Fortran file as part of their own source. If a code the uses an old `plumed_f08` module source file but links to a the new library with the changed C-interface, it will make the wrong assumption about the C-interface and possible crash unexpectedly. The solution would be to compile the Fortran file and install the resulting mod-file, which projects then can directly use. (It would of course require a separate installation for each Fortran compiler, but that is the case with every modern Fortran project anyway.)

aradi:
OK, I've tried to modify my calls based on the example in the repository. I have now:
```fortran
    call plumedCalc%cmd_val("setStep", iGeoStep)
    call plumedCalc%cmd_ptr("setForces", derivs)
    call plumedCalc%cmd_val("setEnergy", energy)
    call plumedCalc%cmd_const_ptr("setPositions", coord0)
    call plumedCalc%cmd_const_ptr("setMasses", mass)
    if (tPeriodic) then
      call plumedCalc%cmd_const_ptr("setBox", latVecs)
    end if
    call plumedCalc%cmd_val("calc", 0)
```
When I run it, I get a segfault. Before I try to find it out why, do you see anything spurious in the definitions above? (It works if I use the calls of our Plumed interface, so probably one of the calls is wrong?)

fortran/plumed_f08.f90
GiovanniBussi: This would not be possible, and would not be compatible with the spirit of the original type-check implementation. The idea is to reproduce the behavior of type conversions in C++. If a given cmd requires a `const` reference, which means it promises that will not modify the buffer, a user should be allowed to pass a non `const` data.

> Therefore, the resulting code could in theory fail, and finding the bug would be a nightmare

This will not happen, unless users use `cmd_ptr` without `target`, which is not detectable by the compiler but explained in the documentation of the new interface. And, in any case, forcing the user to use `cmd_val` when PLUMED does not need the address will not help in this case.

GiovanniBussi:
>@GiovanniBussi I updated DFTB+. Currently I stumbled upon, that setPositions, which I use to pass the atom positions to Plumed, does not work with cmd_val. Is it supposed to be like that? Can I look it up somewhere which keyword requires which kind of call / constantness?

Currently this is not documented anywhere, but I plan to add this information. Anyway, positions are only transfered later when "calc" is called, so it is expected.

> Also a suggestions. Calling the new f08 interface module plumed_f08_module would feel much more natural as plumed_module_f08 and would follow a convention seen in several projects (adding the _module suffix to module names).

I will fix this.

> Finally, a remark: the fact that the plumed_f08 module is not compiled during installation (or did I just mess it up?), but it must be added to the users code and compiled there, could be very critical, if the C-interface ever changes.

The C interface called from the module is frozen, so this is not an issue. Notice that it is designed to be extensible without ABI changes (that's for instance the reason for the `opt` argument [here](https://github.com/plumed/plumed2/blob/473c3283e23c120b8aafaab17e58129e7d68805e/fortran/plumed_f08.f90.template#L159-L167)).

>When I run it, I get a segfault. Before I try to find it out why, do you see anything spurious in the definitions above? (It works if I use the calls of our Plumed interface, so probably one of the calls is wrong?)

This is not expected... if there is a mistake in the calls you should have an exception and not a segfault.  Are you using the version with fixed intent(inout) 473c3283e23c120b8aafaab17e58129e7d68805e ? Are all arguments of `cmd_*ptr` functions declared as target?

aradi:
OK, the problem is on PLUMED's side, as it trashes the memory during the call. The segfault at some other point is just the consequence of it. I have a good guess, where it comes from, I'll try to fix it and test and send you a patch.

aradi:
OK, have a look at #753, the implementation and the comments, and then lets discuss about it. According to my current knowledge about Fortran, the solution proposed there is waterproof: Plumed will always get pointers to contiguous arrays, and it will be warranted (by the standard), that the pointers will not point to temporary arrays created in the interface when a pointer or a const pointer is needed. It would also handle strides (e.g. `array(1:10:2)`) correctly or stop with a meaningful error message if not possible (for pointers).

PR #753, opened 2021-11-07
Increase F08 interface robustness, simplify implementation

aradi:
Makes the Fortran 2008 interface more robust and simplifies it.

* Ensures that arrays are contiguous when the C-pointer to them is created. Note, that passing something like `array(:, :2:2)` would work now correctly. Either it would be copied into a contiguous array (`_ref` and `_val`), or a run-time error would be raised (`_ptr`, `_const_ptr`). In earlier versions it just created a spurious segfault without much clues why this ocured.
* Creates C-pointer immediately to exclude any copy-in/copy-out happening in further calls within the interface.
* Requires pointer actual argument for `cmd_ptr` and `cmd_const_ptr` to be consistent with naming as well as ensuring the the objects can be safely pointed to.

Notes:
* Workaround for Intel & Co: Caller defines manually pointers to the arrays, and passes those pointers. That's the only safe way to ensure, that Plumed gets something which can be safely pointed to. Note, that this is a [compiler bug](https://community.intel.com/t5/Intel-Fortran-Compiler/Compiler-bug-target-array-can-not-be-passed-for-pointer-argument/m-p/1327580#M158270).
* I've simplified a lot by making `plumed_safe_ptr` and `plumed_f_cmd_*` rank agnostic. As far as I see, this is without side effects. If I overlooked something, let me know, I can add back the rank-specific version.
* I don't understand, why `nelem = product(pass_shape)`, which always evaluates to zero, but left it as it is, in case, it is on purpuse.
* One should make the interface generation for `long_double` optional (I did not do so), as at least one compiler I've tried (NAG) does not support that data type.
* The contiguosity of the passed arguments for `_ref` and `_val` is enforces by the compiler (with eventual copy-in/copy-out)
* The contiguosity of the passed pointers can not be enforced (as copy-in/copy-out should be avoided) and is, therefore, checked at run-time. I am calling `error stop` if not fulfilled, maybe you can turn it into a proper error reporting.
* I've compiled and succesfully tested the thing with Intel2021, NAG 7 and GNU 10 so far. The corresponding DFTB+ implementation is in the [plumed08 branch](https://github.com/aradi/dftbplus/tree/plumed08)

### DIFF
--- a/fortran/plumed_f08.f90.template
+++ b/fortran/plumed_f08.f90.template
@@ -60,7 +60,7 @@ module plumed_f08_module
   type plumed
     type(cplumed), private :: handle
     logical,       private :: initialized = .false.
-  contains 
+  contains
     private
 
     generic, public :: cmd => pl_cmd
@@ -137,7 +137,7 @@ module plumed_f08_module
       integer(kind=c_int) :: cplumed_installed
     end function cplumed_installed
   end interface
-    
+
   interface
     function cplumed_valid(p) bind(C,name="plumed_valid")
       import
@@ -170,7 +170,7 @@ module plumed_f08_module
   interface
     function plumed_f_safeptr_char(val,nelem,pass_shape,flags,opt) bind(C)
       import
-      character(kind=c_char)                :: val(*)
+      type(c_ptr),                    value :: val
       integer(kind=c_size_t),         value :: nelem
       integer(kind=c_size_t)                :: pass_shape(*)
       integer(kind=c_size_t),         value :: flags
@@ -182,20 +182,9 @@ module plumed_f08_module
 #: for t in types
 #: for k in kinds[t]
   interface
-    function plumed_f_safeptr_${cname(k)}$_scalar(val,nelem,pass_shape,flags,opt) bind(C)
-      import
-      ${t}$(kind=${k}$)                     :: val
-      integer(kind=c_size_t),         value :: nelem
-      integer(kind=c_size_t)                :: pass_shape(*)
-      integer(kind=c_size_t),         value :: flags
-      type(c_ptr),                    value :: opt
-      type(cplumed_safeptr)                 :: plumed_f_safeptr_${cname(k)}$_scalar
-    end function plumed_f_safeptr_${cname(k)}$_scalar
-  end interface
-  interface
     function plumed_f_safeptr_${cname(k)}$(val,nelem,pass_shape,flags,opt) bind(C)
       import
-      ${t}$(kind=${k}$)                     :: val(*)
+      type(c_ptr),                    value :: val
       integer(kind=c_size_t),         value :: nelem
       integer(kind=c_size_t)                :: pass_shape(*)
       integer(kind=c_size_t),         value :: flags
@@ -237,32 +226,23 @@ module plumed_f08_module
 
      ! we then define all the functions needed for overloading
 
-     subroutine plumed_f_cmd_ptr(p,key,val,dummy,error,const,nocopy)
+     subroutine plumed_f_cmd_ptr(p,key,val,const,nocopy,error)
        type(cplumed),                 intent(in)    :: p
        character(kind=c_char,len=*),  intent(in)    :: key
-       type(c_ptr),                     value       :: val
-       type(dummy_type),   optional                 :: dummy
+       type(c_ptr),                   intent(in)    :: val
+       logical,                       intent(in)    :: const
+       logical,                       intent(in)    :: nocopy
        type(plumed_error), optional,  intent(out)   :: error
-       logical,            optional,  intent(in)    :: const
-       logical,            optional,  intent(in)    :: nocopy
        type(plumed_error), target :: myerror
        integer(kind=c_size_t) :: pass_shape(1)
        integer(kind=c_size_t) :: flags
        type(cplumed_nothrow_handler) :: nothrow
        integer(kind=c_size_t) :: nelem
-       pass_shape=(/0/)
+       pass_shape=[0]
        nelem=product(pass_shape)
        flags=flags_ptr
-       if(present(const)) then
-         if(const) then
-           flags=flags_const_ptr
-         endif
-       endif
-       if(present(nocopy)) then
-         if(nocopy) then
-           flags=flags+flags_nocopy
-         endif
-       endif
+       if (const) flags=flags_const_ptr
+       if (nocopy) flags=flags+flags_nocopy
        nothrow%ptr=c_loc(myerror)
        if(present(error)) then
          nothrow%handler=c_funloc(plumed_f_f08_eh)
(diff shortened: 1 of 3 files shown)


aradi:
Just one more note: I've removed the `type(dummy)` "argument separator" in the internal routines to make the code more compact. In the public ones, they are still present. I'd, nevertheless, recommend to reconsider them there as well. I understand the intention (forcing the dumb caller to specify all optional arguments as keyword arguments to keep) and it make a lot of sense. However, if the caller ever wanted to create a procedure pointer to that routine (or pass it as an argument to an other routine), he would fail, as part of the API (`type(dummy)`) is not public... (If you really insist, I can put them back even in the internal ones...)

fortran/plumed_f08.f90.template
```diff
@@ -181,21 +181,10 @@ module plumed_f08_module
 
 #: for t in types
 #: for k in kinds[t]
-  interface
```
GiovanniBussi, line 184: I agree with removing this. It was there since it is needed in gfortran 4 (and I originally copied this from the plumed.f90 interface, compatible with gfortran 4), but it can be removed here

fortran/plumed_f08.f90.template
```diff
@@ -237,32 +226,23 @@ module plumed_f08_module
 
      ! we then define all the functions needed for overloading
 
-     subroutine plumed_f_cmd_ptr(p,key,val,dummy,error,const,nocopy)
```
GiovanniBussi, line 240: Also here, there's not reason to keep the dummy argument in this internal call, so good idea removing it

src/wrapper/Plumed.h
```diff
@@ -3405,7 +3405,7 @@ __PLUMED_IMPLEMENT_FORTRAN(plumed_f_use_count,PLUMED_F_USE_COUNT,(char*c,int*i),
 /* note: flags & (~0x1ffffff) removes bits that are set here (code and size) */
 
 #define __PLUMED_IMPLEMENT_F_SAFEPTR_INNER(type,type_,size,code,suffix) \
-plumed_safeptr plumed_f_safeptr_ ## type_ ## suffix(type*val,__PLUMED_WRAPPER_STD size_t nelem,__PLUMED_WRAPPER_STD size_t*shape,__PLUMED_WRAPPER_STD size_t flags,void*opt) {\
+plumed_safeptr plumed_f_safeptr_ ## type_ ## suffix(void*val,__PLUMED_WRAPPER_STD size_t nelem,__PLUMED_WRAPPER_STD size_t*shape,__PLUMED_WRAPPER_STD size_t flags,void*opt) {\
```
GiovanniBussi, line 3408: I guess this change has no effect (the function is only called by Fortran), but it's ok. I also guess it is not a problem to continue defining it as with
```
${t}$(kind=${k}$)                     :: val
```
in the plumed.f90 (old) interface.

GiovanniBussi:
>I don't understand, why nelem = product(pass_shape), which always evaluates to zero, but left it as it is, in case, it is on purpuse.

It's a mistake, sorry for that... Should have been the product until the second-to-the-last component. The last one (0) is only added to make the array zero terminated (that's how it is defined in the C interface).

However, notice that if you pass zero the check will be done correctly using the passed shape. The reason to allow passing shape and/or nelem was to allow the user passing them as additional optional arguments, but in the end I didn't enable this. This means that the feature could be removed. In short, there are two options that would make sense:
- pass 0 (replacing e.g. `plumed_f_safeptr_char(val,nelem,pass_shape,flags,c_null_ptr),nothrow)` with `plumed_f_safeptr_char(val,0,pass_shape,flags,c_null_ptr),nothrow)`, and removing the variable nelem).
- fix the product to something like `nelem=product(pass_shape(1:size(pass_shape)-1))`

> One should make the interface generation for long_double optional (I did not do so), as at least one compiler I've tried (NAG) does not support that data type.

I didn't know this was not in the standard. It can be removed as we are not using it now, it was just added for completeness.

> The contiguosity of the passed arguments for _ref and _val is enforces by the compiler (with eventual copy-in/copy-out)
The contiguosity of the passed pointers can not be enforced (as copy-in/copy-out should be avoided) and is, therefore, checked at run-time. I am calling error stop if not fulfilled, maybe you can turn it into a proper error reporting.

Very good. I would rather do the equivalent of "raising an exception", that is: if `error` was passed, populate it, otherwise stop. I can include this after merging your PR.

> Just one more note: I've removed the type(dummy) "argument separator" in the internal routines to make the code more compact

Thinking about it, I think it could be removed also from the external methods. The reason for this was to make sure explicit keywords were used for `const` and `nocopy`, which are logical. But since `error` is already a dedicated type, there's no risk of ambiguity. So I would remove it also from the external methods. Also here, I can do it after merging your PR.

> I've compiled and succesfully tested the thing with Intel2021, NAG 7 and GNU 10 so far. The corresponding DFTB+ implementation is in the plumed08 branch

This is all very good. Many thanks for your help!!! I will double check things tomorrow then merge.

GiovanniBussi:
Hopefully 20bf0dd81b9fae91369e3c7fe73f7e3d90826167 on https://github.com/plumed/plumed2/tree/fortran-exceptions-fix fixes the regtest failure with Intel compiler

src/wrapper/Plumed.h
aradi: Yes, the difference is only "syntactic sugar". The old Fortran interfaces should also work without problems.

aradi:
> However, notice that if you pass zero the check will be done correctly using the passed shape. The reason to allow passing shape and/or nelem was to allow the user passing them as additional optional arguments, but in the end I didn't enable this. This means that the feature could be removed. In short, there are two options that would make sense:
> 
>     * pass 0 (replacing e.g. `plumed_f_safeptr_char(val,nelem,pass_shape,flags,c_null_ptr),nothrow)` with `plumed_f_safeptr_char(val,0,pass_shape,flags,c_null_ptr),nothrow)`, and removing the variable nelem).
> 
>     * fix the product to something like `nelem=product(pass_shape(1:size(pass_shape)-1))`
> 

I'd lobby for the first approach in order to avoid passing redundant information. If you pass the shape, the callee can calculated the nr. of element itself if needed.
> 
> > One should make the interface generation for long_double optional (I did not do so), as at least one compiler I've tried (NAG) does not support that data type.
> 
> I didn't know this was not in the standard. It can be removed as we are not using it now, it was just added for completeness.

Oh well, the Fortran standard is very generous about things like this in order to ensure, that the language stays "independent" of the hardware it is used on. Compilers are allowed to support various different numerical representations, but they don't have to. (All compilers I know fortunately support double precision..)

Actually, one other thing: the `testfull.f90` example passes an allocated, but uninitialized `masses` array  via `const_ptr` to Plumed. I guess, the value of the masses is irrelevant for the specific test, but then one should either not pass it  or at least fill it with definite values. The nice thing of the new interface is, that by seeing that the array is uninitialized and that it is passed via `const_ptr`, one immediately knows, that it can not be correct. :wink:

GiovanniBussi:
I implemented all your suggestions [here](https://github.com/plumed/plumed2/tree/fortran-exceptions-fix) (just force pushed). If tests are successful on GHActions I will merge.

Thanks a lot!!

Giovanni

aradi:
Cool, you're welcome! Actually, do you plan to publish something about the new interface in some form? I think, our current C++/Fortran interface is really nice (I may be biased, of course :wink:) and the concept could be of great interest for both communities.

GiovanniBussi:
Merged indirectly in 81892a946574063484e9ee13407edcdb07745903

PR #759, opened 2021-11-18
Various additions and changes in the VES module

valsson:
##### Description

Various additions and changes in the VES module. Almost all of the work was done by @bpampel. 

Including:  
  - New localized basis functions: Wavelets (BF_WAVELETS), Gaussians (BF_GAUSSIANS), and cubic splines (BF_CUBIC_B_SPLINES). 
  - New optimizer based on Adam (OPT_ADAM). 
  - New optimizer based on classical Robbins Monro stochastic gradient descent (OPT_ROBBINS_MONRO_SGD). 
  - Fixed a bug in VES_LINEAR_EXPANSION for multidimensional bias potential if one (or more) of the CVs is outside the range of the bias potential. 
  - Updated copyright header.
  - Other minor changes to fix compiler warnings. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests

Tests are currently running on GH Actions and with the latest changes, I expect them to pass. But I will make changes if needed. 


valsson:
I have some issues with the numerics of the regtests for the wavelet generation. The derivate values seem to be sensitive to the compiler and lapack used. But I have fixed that by reducing the printout and making some other changes. However, now the failure is due to the fact that the fixzeros.sh script does not correctly handle values of the form `-0.00`. PR #761 should fix this and with that PR, this PR should pass the CI checks.

valsson:
The LGTM errors ("Multiplication result may overflow 'int' before it is converted to 'size_type'.")
are related to things like this:
`std::vector<double> A(N*N)` 
where it complains that `N` is an `int`. This is the same error as in Matrix.h (line 271) and seems to be present elsewhere in the code. 
We could fix it by using `A( static_cast<size_t>(N) * static_cast<size_t>(N) )`. 
Should I do this?

GiovanniBussi:
Don't worry about the lgtm warning. Let's see if the numerical issue is solved on GHA now (I guess you have the merge master in your branch and push again)

valsson:
It passed on GH Action after the fixzeros.sh fix, but then I noticed in the coverage that the regtests were missing some parts of the code, so I added further tests. Once that has passed, this can be merged as I am concerned.

GiovanniBussi:
Thanks!!

PR #761, opened 2021-11-20
Change to fixzeros.sh script in regtests

valsson:
##### Description

Change to fixzeros.sh script in regtests to make it more general. For example, now values of the form `-0.00` are correctly handled. This change is needed to make the PR #759 pass the CI tests.

This passes all CI tests, see https://github.com/ves-code/plumed2-ves/actions/runs/1482785956 

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/regtest/scripts/fixzeroes.sh
+++ b/regtest/scripts/fixzeroes.sh
@@ -2,6 +2,7 @@
 
 file=$1
 
-sed "s/-0.000/ 0.000/g" $file > $file.$$.tmp
+sed "s/-\(0\.0*0 \)/ \1/g; 
+     s/-\(0\.0*0$\)/ \1/g" $file > $file.$$.tmp
 mv $file.$$.tmp $file
 


GiovanniBussi:
Good point. These are true zeros right? (Any number of zeros followed by a space or new line)

valsson:
Yes, it should capture anything of the form `0.0....0`, there is perhaps some other way to do this

valsson:
For example, 

`echo "-0.0000000 -0.00000  -0.00 -0.0 -0.000000 -0.0000010000" | sed "s/-\(0\.0*0 \)/ \1/g;  s/-\(0\.0*0$\)/ \1/g"`

will generate 

` 0.0000000  0.00000   0.00  0.0  0.000000 -0.0000010000` 
(the white space is not correct due to some formating of the comment, but it is correctly preserved in the terminal)

valsson:
The current script would change something like `-0.00034` to `0.00034`, I recall I had some issues with this in the past but I solved it differently.

GiovanniBussi:
Yes it makes sense thanks for fixing it!

PR #762, opened 2021-11-20
Fix OPES_METAD html doc

invemichele:
##### Description

The doc page for OPES_METAD is displaying only up to this line `Ftg: BIASVALUE ARG=Ftg_func`, see https://www.plumed.org/doc-master/user-doc/html/_o_p_e_s__m_e_t_a_d.html
When clicking on `Ftg:` one can see that the rest of the page is wrongly included among the components of `BIASVALUE`.

I fixed the problem simply by renaming `Ftg_func` to something without an underscore, but this might happen also somewhere else in the doc, and probably the function "click to see the output components" should be made more robust.

##### Target release

I would like my code to appear in release __2.8__

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/opes/OPESmetad.cpp
+++ b/src/opes/OPESmetad.cpp
@@ -121,8 +121,8 @@ Our target distribution is a Gaussian centered there, thus the target free energ
 
 \plumedfile
 phi: TORSION ATOMS=5,7,9,15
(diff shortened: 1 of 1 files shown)


invemichele:
This can be merged, as far as I am concerned

GiovanniBussi:
Thanks @invemichele !

PR #767, opened 2021-11-27
--enable-rpath also use to PREFIX/lib

EricDeveaud:
##### Description

when `configure --enable-rpath --prefix=/non/standard/location`  build does not set RPATH to PREFIX/lib 
that brings the following error while installed
```
rpm_maker:plumed/plumed-2.7.2 > plumed -h 
plumed: error while loading shared libraries: libplumedKernel.so: cannot open shared object file: No such file or directory
```
the follwoing diff add PREFIX/lib to the rpath

##### Target release
plumed-2.7.2

regards 
Eric


GiovanniBussi:
Thanks! Notice that the right path is `$libdir`. 

I fixed this in master and not in 2.7 since I am not sure it is really a bug fix, rather a preference. If you are aware of some other code following this convention I am happy to backport it.

PR #768, opened 2021-11-29
S2 contact model

valsson:
##### Description

NMR S2 contact model CV used in https://pubs.acs.org/doi/abs/10.1021/acs.jpclett.7b01770. 

I still need to add documentation, I will do that as soon as possible. But the code can be added now. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Is there any reason for this to be added in colvar and not in a separate module?

valsson:
Not really, it is just one file so I didn't think about putting it into a separate module. I could move it to the VES code module, though it is not related to that. Should I do that? If I do that, I would still have it appear under the CV in the manual.

GiovanniBussi:
@valsson I would suggest to have a module with a single action, as we have several already.

valsson:
I have moved the code to a separate module (s2cm)

GiovanniBussi:
Thanks @valsson ! I am afraid you merged in some master-only change. Can you rebase this so that it is only based on v2.8 branch?

valsson:
Ah, sorry, I have a mess of it. I am confused about what is the best way to do the rebase. How would you recommend to do it?

valsson:
I have done it with cherry-picking, I will close this PR and open another one with the correct branch

valsson:
Replaced by #768

PR #771, opened 2021-11-29
Fix for S2CM manual generation

valsson:
##### Description

Fixes some issues with incorrect manual generation for the S2CM module. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/s2cm/COPYRIGHT
+++ b/src/s2cm/COPYRIGHT
@@ -1,4 +1,4 @@
-   Copyright (c) 2021 Omar Valsson
+   Copyright (c) 2017-2021 Omar Valsson
 
    This file is part of S2 contact model module 
 
--- a/src/s2cm/S2ContactModel.cpp
+++ b/src/s2cm/S2ContactModel.cpp
@@ -1,7 +1,7 @@
 /* +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-   Copyright (c) 2021 Omar Valsson
+   Copyright (c) 2017-2021 Omar Valsson
 
-   This file is part of S2 contact model module
+   This file is part of S2 contact model module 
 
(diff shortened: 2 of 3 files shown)


valsson:
However, this only partially fixes the issue, it does not fully work correctly. 
It seems that because there is a numerical character in `S2CMMOD_COLVAR`, the script to replace `@S2CMMOD_COLVAR@` when going from `S2CMMOD.md` to `S2CMMODPP.md` does not work correctly. If I change `S2CMMOD` to `SSCMMOD` everywhere in the module documentation it works fine. 

So using a numerical character in the name of the module documentation (i.e., in X1XX_COLVAR`), breaks the script to do the replacement in the `PP.md` files. 

I could fix this by renaming the module documation. But maybe it would be better to solve this in the manual generation script.

GiovanniBussi:
@valsson is it due to these two lines perhaps:
```
user-doc/extract:keywords=`grep "@[A-Z_]*@" *.md | awk -F"@" '{print $2}' | sort | uniq`
user-doc/extract:grep -Ev '^ *@[A-Z:_]*@ *$' ${file}PP.$extension > tmp;
```

Maybe the search should be replaced with something like `@[A-Z_][A-Z0-9_]*@`. Can you check if this solves the problem?

GiovanniBussi:
@valsson after 4478523 will be pushed to https://www.plumed.org/doc-v2.8/user-doc/html/index.html we will see if it works on Github Actions. Then let me know if I should merge this anyway. Thanks!

valsson:
Thanks @GiovanniBussi! 

This PR should be merged regardless as it fixes wrong names used in `PLUMEDOC S2CMMOD_COLVAR S2CM` that is independent of the issue with numerical character in the module names.

PR #790, opened 2022-02-10
Some changes to OPES_METAD

invemichele:
##### Description

Changed the syntax of EXTRA_BIAS_ARG, since some users told me it was confusing.
Added the keyword EXCLUDED_REGION for defining a region where kernels are not deposited.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Are you using a Mac? I think that option is not supported

invemichele:
ok, I found the problematic line, is [this one](https://github.com/plumed/plumed2/pull/790/files#diff-ac69602d16ad1cdaf73dc3789a3c7492541316e7b51c3f2fd91a7f45de7a2803R531). If I comment it all tests pass successfully, but I still don't understand why.

Here is what I am doing.
I want to read the output value of some other action, besides those in ARG. To make sure that this output is calculated, I call the `requestArguments` function, on a vector containing all the ARG values plus the extra ones. This works fine in all of my tests, because `requestArguments` clears all dependencies before adding the new ones.
However, in the **CI / linux (-debug-mpi-)** run this runtime error appears:
`Error: attempt to subscript container with out-of-bounds index 1, but container only holds 1 elements.`
when 1 CV is passed through ARG, and 
`Error: attempt to subscript container with out-of-bounds index 2, but container only holds 2 elements.`
when 2 CVs are passed through ARG.

If I remove the call to `requestArguments` this error does not appear, but then the value of the extra argument might not be calculated by plumed if it is not used somewhere else.

what am I misunderstanding? any suggestion?

invemichele:
on my side this pull request is ready for merge

GiovanniBussi:
Thanks @invemichele !

I checked the changes to the core, and they look good to me. I am merging then.

PR #796, opened 2022-02-27
Masterclass 22.3

invemichele:
##### Description
Here is my masterclass for tomorrow.

I also unified the numbering of previous 2022 masterclasses, so that they will appear in the proper order both on the website and in the source:
- 22-0x -> name of the file and web page
- 22.x   -> name in the index and title of the page

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Thanks @invemichele !! Good that you fixed the file names. I think we have to fix the links in the plumed.org/masterclass as well

invemichele:
Actually, now that I check, in the end I only changed the file name, and not the web page name. So the links are safe.

If we want to have more consistency in the web pages naming, the one of funnel metadynamics should be changed from `masterclass-22-1.html` to `masterclass-22-01.html`
Note that this is different from the masterclass of 2021, but those were less than 10...

spiwokv:
Sorry, it was my mistake to introduce "0" to the file name and to chapter IDs. I will unify everything with the first masterclass of 2022.

GiovanniBussi:
Hi! I think the 0 is fine and will be needed to have the correct order in the end

PR #799, opened 2022-03-02
QuantumEspresso 7.0 patch

ralf-meyer:
##### Description

I have recently been contacted by several people about an updated version of the QuantumEspresso patch. While I personally do not currently use the combination of PLUMED + QE this updated version might still be of interest to the plumed community (see e.g.: https://lists.quantum-espresso.org/pipermail/developers/2019-December/002137.html).

##### Type of contribution

##### Copyright

Not applicable?

##### Tests

I would be happy to contribute an automated test provided some input of how to implement a regtest for patches.


GiovanniBussi:
@ralf-meyer this is very useful thanks!!!

We don't have any official way to test patches because the patching procedure is something that we are trying to phase out (encouraging developers of MD codes to include directly the code in their source). I am afraid anyway that this will not happen for QE. It would be nice then to have some infrastructure to download the MD code, install PLUMED, do the patch, and perform some very basic calculation, ideally using GitHub Actions. This would not be done within the PLUMED regression test suites, but rather on a separate GitHub repository. I think @gtribello was trying to do [something similar with LAMMPS](https://github.com/plumed/test-lammps-and-plumed) (in this case though there is no patch, as the calls are already included in LAMMPS), but I think it's a bit outdated (I see it was not yet migrated to GitHub Actions).

Given this, I am happy to merge your request 'as is' if you tested it. But I would open an issue so as to add to our TODO list the creation of such a repository to test this patch, and will ask your help later. If you can provide a simple list of commands that can be used to (a) install QE from scratch on ubuntu (b) run a basic calculation with QE+PLUMED and (c) include a reference output (ideally a text file) I can start from there.

ralf-meyer:
Ok, this does indeed sound like a longer term project. 

Regarding the current status: I have already received feedback from two of the people that reached out to me confirming that they were able use the new patch and reproduce results they obtain with QE6.3 and the old patch.

Regarding a possible workflow:
(a) Setup
```
git clone https://gitlab.com/QEF/q-e.git
cd q-e
git checkout qe-7.0
./configure
plumed patch --engine qespresso-7.0 --patch
make pw
```   


(b) I will first look into the existing test files in https://github.com/plumed/plumed2/tree/master/test/qespresso/PW and see if those are still functional. Alternatively I have a minimal h2 example that I used for testing.

(c) Which output files would you ideally like to compare against (COLVAR, HILLS)?

GiovanniBussi:
Great!

Regarding the script, does it require (and find) MPI compilers? Should we install MPI first right? Or can it work both with or without (in which case I would just have two builds)?

Regarding the output, I would keep the thing to minimum. Something like:
- DUMPATOMS to make sure atoms are passed correctly
- RESTRAINT to add some force on some arbitrary variable, making sure that it affects the trajectory

We can even just check the file resulting from the DUMPATOMS command.

GiovanniBussi:
@ralf-meyer I merged your request and created a new repository [here](https://github.com/plumed/test-plumed-qe).  You should be able to see builds in the [GitHub Actions tab](https://github.com/plumed/test-plumed-qe/actions). And [here](https://github.com/plumed/test-plumed-qe/blob/main/.github/workflows/ci.yml)'s the corresponding workflow.

I think at this point we can add tests there. I will then make sure that the repository is routinely tested. Thanks!

Issue #678, opened 2021-03-17
Add configure support for libtorch


valsson:
I see there is a branch called libtorch with commit 684b5d7 trying to add support for libtorch to the configure script. @GiovanniBussi, what is the status of this? Is it still a work in progress? 

It would be useful to add support for libtorch as it is started to be used in some plumed modules. 

I am happy to help with this as I can, including testing and such

luigibonati:
Hi @valsson!
I am using Libtorch with PLUMED,  and I also wrote a simple interface to load serialized models trained in python as CVs (see e.g. https://www.plumed-nest.org/eggs/20/004/). 

It is relatively easy to link a specific version of libtorch, here you can find some instructions for libtorch v. 1.4 (in the last section of the notebook)
https://colab.research.google.com/drive/1dG0ohT75R-UZAFMf_cbYPNQwBaOsVaAA 

I tried with @GiovanniBussi to see whether it was possible to detect Libtorch automatically in the configure script. However, I believe that even if it is possible it is not a good option for now, as libtorch is not yet mature enough to be included in a stable version. Some issues are the following:

- The standard way to use libtorch is to link the precompiled version which works fine in most circumstances. However, this makes it harder to link PLUMED against it, as they both look for the same libraries. For instance, the precompiled version of libtorch comes with a version of lapack that is not complete, so when PLUMED looks for the latter it complains. A workaround is to use internal blas/lapack with PLUMED.

- The Pytorch C++ APIs are in beta phase, with continuous major breaking changes. In the last two years, they changed two or three times the way in which they serialize the model from Python, with no backward compatibility. This means that the interface to load models trained in Python inside PLUMED possibly needs to be updated.

- They recommend using CMake to build applications, in order to set the right library flags which might otherwise change from one version to another. 

So, while it is relatively easy to link a specific version of libtorch for development purposes, I think it is not easy to integrate into PLUMED, at least until they release a more stable version.

I'm working on writing updated instructions on how to configure PLUMED with newer versions of libtorch, I'll let you know as soon as I do.

Luigi

valsson:
Thank you @luigibonati for the information! 

@jakryd and I are also relying heavily on libtorch in our [MRSE](https://arxiv.org/abs/2007.06377) method. We have been using our own compiled version of libtorch (not the precompiles ones) that is linked to PLUMED. 

[...]

PR #800, opened 2022-03-08
Add Pytorch module

luigibonati:
##### Description

This pull request adds a module called `pytorch` which contains an interface between the Pytorch machine learning library and PLUMED. This allows one to define a model in Python (e.g. a neural network), compile it with TorchScript (Pytorch jit), and use it as a CV or as input of other functions in PLUMED.

This module relies on the PyTorch C++ APIs (LibTorch). I documented the steps required for the installation in the documentation of the module. 

I modified the configure script by adding a flag `--enable-libtorch` which tells PLUMED to look for these libraries during configure. This also solves #678.
The pytorch interface is then only compiled if libtorch is found (via a preprocessor directive `__PLUMED_HAS_LIBTORCH`)

I updated the Github Action workflows to download the pre-built version of LibTorch and link it to plumed. I tested this for ubuntu and it works, however not for all queues:
- debug: since the library was not compiled with GLIBCXX_DEBUG
- intel: apparently, when linking libtorch several regtests of other actions fail.

There are also a couple of drawbacks related to linking the pre-built library:
- Due to a conflict with the BLAS/LAPACK libraries it is necessary to disable the search for external libraries. The reason is that (part of) these libraries are contained as dependencies in the LibTorch library but they are not complete, so PLUMED complains about missing dependencies.
- It contains unique global symbols, so I had to disable nmcheck as done for boost. Not sure about how big is this issue

Note that downloading the pre-built library is the only suggested way to install LibTorch, as they provide very little documentation on how to compile it.
Despite these shortcomings, I tried to make the integration as robust as possible, in particular:
- In the testing workflows I link the LTS version of LibTorch rather than the last one, to avoid breaking changes in LibTorch
- Thanks to the preprocessor condition if the linking is not successful no errors are raised, only in this case the pytorch module would not work

Let me know what you think! 

Thanks, 
Luigi

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@luigibonati thanks a lot this is very useful!!! @maxbonomi will be happy.

I think the way you modified the GitHub Actions workflow is perfect. Related to your comments:

> Due to a conflict with the BLAS/LAPACK libraries it is necessary to disable the search for external libraries. The reason is that (part of) these libraries are contained as dependencies in the LibTorch library but they are not complete, so PLUMED complains about missing dependencies.

Is there a way to solve this? Did you check what happens if you link both the libtorch lapack/blas and the system ones? If the problem is that linking libtorch the `./configure` script thinks blas are present but then linking plumed fails, maybe it is sufficient to use in the `./configure` test one of the lapack/blas functions that you know are missing

> It contains unique global symbols, so I had to disable nmcheck as done for boost. Not sure about how big is this issue

This is not a problem (also boost libs unfortunately do this).

Then an additional possible improvement. Now you pass a lot of variables to configure. Whereas I see that passing paths is normal, perhaps these libraries can be automatically added by configure: `-ltorch -lc10 -ltorch_cpu`? Notice that this should only be done if `test "${libsearch}" == true`, since we typically use `--disable-libsearch` when packaging (conda/macports) so as to make sure that only explicitly added libraries are linked. But for a end user it might be convenient if these libraries are added automatically. Or maybe there is a reason to leave the flexibility e.g. to link `torch_gpu` instead of `torch_cpu`?

luigibonati:
Hi @GiovanniBussi, thanks for the feedback! 

> Is there a way to solve this? Did you check what happens if you link both the libtorch lapack/blas and the system ones? If the problem is that linking libtorch the ./configure script thinks blas are present but then linking plumed fails, maybe it is sufficient to use in the ./configure test one of the lapack/blas functions that you know are missing

If I do a standard installation the config log looks like this:

```
./configure --enable-modules=all

checking whether dgemv can be linked with no library... no
checking whether dgemv_ can be linked with no library... no
checking for library containing dgemv... no
checking for library containing dgemv_... -lblas
checking for library containing dsyevr_... -llapack
```

while if I ask for libtorch:

```
./configure --enable-cxx=14  --enable-libtorch LIBS="-ltorch -lc10 -ltorch_cpu"  --enable-modules=all

checking whether dgemv can be linked with no library... yes
checking for library containing dsyevr... no
configure: WARNING: using internal lapack, could be inefficient
```

it appears that it finds blas but not lapack. I checked and libtorch_cpu.so contains dgemv (and dgemv_) but no lapack symbols. From what I understand from the configure, in this case it should use the external blas but internal lapack. However when linking it throws a lot of undefined references to lapack routines:

```
 /usr/bin/ld: ../lapack/lapack.o: in function `PLMD::lapack::dgetf2(int*, int*, double*, int*, int*, int*)':
lapack.cpp:(.text+0x105): undefined reference to `idamax'
/usr/bin/ld: lapack.cpp:(.text+0x1c4): undefined reference to `dger'
/usr/bin/ld: ../lapack/lapack.o: in function `PLMD::lapack::dlarf(char const*, int*, int*, double*, int*, double*, double*, int*, double*)':
lapack.cpp:(.text+0x3c03): undefined reference to `dger'
/usr/bin/ld: lapack.cpp:(.text+0x3c80): undefined reference to `dger'
/usr/bin/ld: ../lapack/lapack.o: in function `PLMD::lapack::dlarfb(char const*, char const*, char const*, char const*, int*, int*, int*, double*, int*, double*, int*, double*, int*, double*, int*)':
lapack.cpp:(.text+0x3f39): undefined reference to `dtrmm'
/usr/bin/ld: lapack.cpp:(.text+0x3f8a): undefined reference to `dtrmm'
/usr/bin/ld: ../lapack/lapack.o:lapack.cpp:(.text+0x42a4): more undefined references to `dtrmm' follow
/usr/bin/ld: ../lapack/lapack.o: in function `PLMD::lapack::dlarft(char const*, char const*, int*, int*, double*, int*, double*, double*, int*)':
lapack.cpp:(.text+0x8539): undefined reference to `dtrmv'
/usr/bin/ld: lapack.cpp:(.text+0x8746): undefined reference to `dtrmv'
/usr/bin/ld: ../lapack/lapack.o: in function `PLMD::lapack::dlasd7(int*, int*, int*, int*, int*, double*, double*, double*, double*, double*, double*, double*, double*, double*, double*, int*, int*, int*, int*, int*, int*, int*, double*, int*, double*, double*, int*)':
lapack.cpp:(.text+0x11e2a): undefined reference to `drot'
/usr/bin/ld: lapack.cpp:(.text+0x11e63): undefined reference to `drot'
```

Do you think this issue might be solved if instead of passing the libtorch libraries in LIBS we add them in the configure, as you suggested in the second part of your answer? If I understand correctly in this way they should be linked only when looking for libtorch, and not for every other configure test as it happens now with LIBS.

> Then an additional possible improvement. Now you pass a lot of variables to configure. Whereas I see that passing paths is normal, perhaps these libraries can be automatically added by configure: -ltorch -lc10 -ltorch_cpu? Notice that this should only be done if test "${libsearch}" == true, since we typically use --disable-libsearch when packaging (conda/macports) so as to make sure that only explicitly added libraries are linked. But for a end user it might be convenient if these libraries are added automatically. Or maybe there is a reason to leave the flexibility e.g. to link torch_gpu instead of torch_cpu?

In general there are pros and cons. The name of the libraries to be linked have changed from one version to the next, so
keeping this flexibility would allow linking also different Pytorch versions. The best would be to have a default that works with the LTS version together with the possibility to change it at configure time. Do you think this is possible?

Regarding its implementation, I tried to add at the fourth argument of PLUMED_CHECK_CXX_PACKAGE the list of libraries that need to be linked, but I have some troubles (e.g. they are used one at a time rather than simultaneously). What is the syntax for executing the test program while linking multiple libraries at once?

Thank you very much!

GiovanniBussi:
@luigibonati I would proceed in steps.

First, I think we should find out if it is possible to link simultaneously torch and system blas/lapack, then we can figure out how to make the flags to be discovered automatically.

1. Check that both lapack and blas and installed on your system
2. Try to configure with `--disable-libsearch LIBS="-ltorch -lc10 -ltorch_cpu -llapack -lblas"`

`--disable-libsearch` makes sure that no other library is added (it will then fail in finding e.g. gsl and fftw, but this is irrelevant for this test). This should link external blas and lapack. Then try to remove `-llapack` to see if it can work with internal lapack and external blas. And, finally, remove also `-lblas`. Here it will think that blas are present and thus enable external blas with internal lapack (as you have seen already), then failing the link.

If this is what happens, then we have just to change the search for the blas function so that it looks for something that is not in torch lib. Currently we search for `dgemv`, we could search for `dtrmm`.

Once this is done, we can try to automatize the construction of LIBS (if it's worth) or even leave it as is. If this does not work I will try on my machine as well.

Notice that autoconf always prepends libraries. So, if you let autoconf add the `-llapack -lblas` flags, they will be added *before* your supplied `LIBS`. On the other hand, if we add the flags with autoconf, they will be in the order above.

@carlocamilloni 
>very nice! also another options could be to automatically disable external bias and lapack when using --enable-libtorch

I think this is ok, but I would try to solve the problem since with RMSD based variables I think our lapack are measurably slower.

luigibonati:
``` 
./configure --enable-cxx=14  --enable-libtorch --disable-libsearch LIBS="-llapack -lblas" --enable-modules=all

checking whether dgemv can be linked with no library... no
checking whether dgemv_ can be linked with no library... yes
checking for library containing dsyevr_... none required
```

```

**Comment by luigibonati** [comment]
Thanks for the very detailed instructions. Below is what I found:

I am on Ubuntu 20.04 with blas/lapack installed via `sudo apt install libblas-dev liblapack-dev`. 

```
./configure --enable-cxx=14 --disable-libsearch LIBS="-llapack -lblas" --enable-modules=all

checking whether dgemv can be linked with no library... no
checking whether dgemv_ can be linked with no library... yes
checking for library containing dsyevr_... none required
```

``` 
./configure --enable-cxx=14  --enable-libtorch --disable-libsearch LIBS="-ltorch -lc10 -ltorch_cpu -llapack -lblas" -enable-modules=all

checking whether dgemv can be linked with no library... yes
checking for library containing dsyevr... no
configure: WARNING: using internal lapack, could be inefficient
```

The issue is the following: libtorch contains dgemv, so the configure looks for lapack without underscore. However, the system blas / lapack are with the underscore so they are not found in this way. 

If I change the order in which the libraries are searched for in the configure (first with underscore and then without) I am able to both configure and compile without linking problems:
```
./configure --enable-cxx=14  --enable-libtorch --disable-libsearch LIBS="-llapack -lblas -ltorch -lc10 -ltorch_cpu" -enable-modules=all

checking whether dgemv_ can be linked with no library... yes
checking for library containing dsyevr_... none required
```

I tried it on my workstation, I will commit it now to trigger the actions and see if it works. Do you think is it a good solution? Should we check what happens in the case in which the system blas/lapack are without underscore?

**Comment by GiovanniBussi** [comment]
Great!!

I was also thinking about the underscore problem. I agree it should be reversed. Historically, blas and lapack are written in Fortran and thus have an additional underscore. So it makes sense to search first the version with underscore.

If I understand correctly, with this fix things work using `--disable-libsearch` and explicitly adding lapack and blas. Still, I guess they would fail if you let autoconf search for lapack and blas, because it will think that blas are available just based on libtorch. To solve this, I think we have to search for `dtrmm`. Then, everything should work at least if you provide explicitly the libtorch libraries.

If this is confirmed we can check how to make autoconf discover them automatically.

**Comment by luigibonati** [comment]
> If understand correctly, with this fix things work using --disable-libsearch and explicitly adding lapack and blas. Still, I guess they would fail if you let autoconf search for lapack and blas, because it will think that blas are available just based on libtorch. To solve this, I think we have to search for dtrmm. Then, everything should work at least if you provide explicitly the libtorch libraries.

Apparently, the issue was not with blas but just with lapack (which was not contained in libtorch), indeed also `dtrmm` is a lapack routine.  By changing the search order in the plumed configure it works well, even if I remove --disable-libsearch. In the latter case, it selects blas from libtorch and lapack from -llapack:

```
./configure --enable-cxx=14  --enable-libtorch LIBS="-ltorch -lc10 -ltorch_cpu" -enable-modules=all

checking whether dgemv_ can be linked with no library... yes
checking for library containing dsyevr_... -llapack
```

with no linking errors. This means that the blas library contained in libtorch is complete, and can coexist with the system lapack.

Now we have to see what happens if we make autoconf look automatically for the libtorch libraries, without explicitly adding them in LIBS. It would also be nice to automatically enable c++14 if libtorch is requested.

**Comment by GiovanniBussi** [comment]
I see thanks! I cannot test on my workstation (I have Centos7 and apparently I would have to recompile torch..). Anyway, I confirm that libtorch has both versions:
```
nm -C libtorch_cpu.so | grep " dgemv"
0000000004e223c0 T dgemv
0000000004e223c0 T dgemv_
```

Notice however that the set of blas libraries in libtorch is not complete. On my machine, libblas contains `drot`:
```
nm -D /usr/lib64/libblas.so | grep " drot"
0000000000010230 T drot_
0000000000010380 T drotg_
0000000000010d40 T drotm_
0000000000010930 T drotmg_
```
and liblapack uses it
```
nm -D /usr/lib64/liblapack.so | grep " drot"
                 U drot_
                 U drotm_
```

However, libtorch does not contain it:
```
nm -D libtorch_cpu.so | grep " drot"
# no match
```

Can you check if you are linking the shared lapack library on your workstation?

What could be happening now is that:
- you are linking the static version of lapack (liblapack.a). When linking static libraries, only the objects within the library actually used are pulled in
- we currently do not use any lapack function calling `drot` (or some other missing blas).
- if later we start using a lapack function calling `drot`, suddenly linking will fail

For this reason, I think it would be better to avoid using the incomplete blas from libtorch. If we manage to have the torch libraries added by autoconf, this should not be a problem because autoconf will first search for a real blas library (using `-lblas`) and only later add the libtorch libraries.

**Comment by luigibonati** [comment]
> Can you check if you are linking the shared lapack library on your workstation?

From what I see it is actually using the shared version.

```
ldd -r src/lib/plumed
	linux-vdso.so.1 (0x00007ffd241d7000)
	libplumedKernel.so => /home/lbonati@iit.local/software/plumed2/src/lib/libplumedKernel.so (0x00007f0c8948b000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0c8946c000)
	libmpi.so.40 => /usr/local/lib/libmpi.so.40 (0x00007f0c89348000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f0c89166000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f0c8914b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f0c88f57000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f0c88f3b000)
-->	liblapack.so.3 => /lib/x86_64-linux-gnu/liblapack.so.3 (0x00007f0c88897000)
	libc10.so => /home/lbonati@iit.local/software/libtorch-1.8.1-cu111//lib/libc10.so (0x00007f0c8860b000)
	libtorch_cpu.so => /home/lbonati@iit.local/software/libtorch-1.8.1-cu111//lib/libtorch_cpu.so (0x00007f0c75d20000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f0c75bd1000)
	libgomp.so.1 => /lib/x86_64-linux-gnu/libgomp.so.1 (0x00007f0c75b8d000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f0c75b6a000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0c8a1d5000)
	libopen-rte.so.40 => /usr/local/lib/libopen-rte.so.40 (0x00007f0c75aad000)
	libopen-pal.so.40 => /usr/local/lib/libopen-pal.so.40 (0x00007f0c7599e000)
-->	libblas.so.3 => /lib/x86_64-linux-gnu/libblas.so.3 (0x00007f0c75931000)
	libgfortran.so.5 => /lib/x86_64-linux-gnu/libgfortran.so.5 (0x00007f0c75667000)
	libgomp-75eea7e8.so.1 => /home/lbonati@iit.local/software/libtorch-1.8.1-cu111//lib/libgomp-75eea7e8.so.1 (0x00007f0c75442000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f0c75437000)
	libcudart-6d56b25a.so.11.0 => /home/lbonati@iit.local/software/libtorch-1.8.1-cu111//lib/libcudart-6d56b25a.so.11.0 (0x00007f0c751ae000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f0c751a9000)
	libquadmath.so.0 => /lib/x86_64-linux-gnu/libquadmath.so.0 (0x00007f0c7515d000)
```

Surprisingly, in this list there is also the system version of blas. Or maybe is it possible that this is an indirect dependency (e.g. called from lapack)? Indeed if I look in the output variables in the config.log (which I also attached), I find that only `-llapack` is added. So at this point it is not clear to me which library is being used. 

```
LIBS=' -lstdc++ -lz -ldl -llapack -ltorch -lc10 -ltorch_cpu'
```

> For this reason, I think it would be better to avoid using the incomplete blas from libtorch. If we manage to have the torch libraries added by autoconf, this should not be a problem because autoconf will first search for a real blas library (using -lblas) and only later add the libtorch libraries. 

I agree with you that if the libtorch blas is incomplete we should try not to rely on it to avoid having broken links in the future. If  adding the libraries via autoconf solves this issue I would go for it.

**Comment by luigibonati** [comment]
I modified the configure script to allow multiple libraries to be linked in the test program. I did it by adding to the `PLUMED_CHECK_CXX_PACKAGE` function an extra argument, which if present and equal to true uses all the libraries specified together rather than as alternatives one at a time.
Actually, they are added one at the time rather than all together. If you think there are better ways of handling this (e.g. by changing the `PLUMED_CHECK_CXX_PACKAGE` to take a bash array in the `library` argument and looping over it), let me know! The problem is that the for loop was made based on spaces, so it was not suited for linking multiple libraries at once.

In this way the blas/lapack libraries are both linked to PLUMED explicitly. @GiovanniBussi  do you think that this solves the compatibility issues you were mentioning before?
```
./configure --enable-libtorch --enable-modules=all 

checking whether dgemv_ can be linked with no library... no
checking whether dgemv_ can be linked with no library... no
checking for library containing dgemv_... -lblas
checking for library containing dsyevr_... -llapack
```

I also modified the configure so that if --enable-libtorch is requested the c++14 support is enabled automatically. 

The result of these changes is that, provided that the location of the libtorch include/libraries is exported in the relevant envs, one can just do:
```
./configure --enable-libtorch --enable-modules=pytorch
```
which is much cleaner for the final user.

GiovanniBussi:
I think it's perfect thanks a lot @luigibonati !!! I will just wait for the CI build then merge

luigibonati:
@GiovanniBussi I double-checked a couple of things and I think it's fine now! you can proceed with the merge!

PR #803, opened 2022-03-16
MEMFUSIONP

lautarodibartolo:
##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Hi @lautarodibartolo I think you added by mistake a lot of files (all those in subdirectories of `src/membranefusion/`). These files are generated automatically. Could you please fix this in the pull request?

lautarodibartolo:
Hello @GiovanniBussi , you are right. I forgot to `make clean` before submitting. I will resubmit a new pull request (I can't delete all the files).

lautarodibartolo:
Done!

src/membranefusion/MemFusionP.cpp
```diff
@@ -0,0 +1,583 @@
...
+   plumed is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with plumed.  If not, see <http://www.gnu.org/licenses/>.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ */
+#include "Colvar.h"
+#include "ActionRegister.h"
```
GiovanniBussi, line 23: @lautarodibartolo could you please replace these with:
```
#include "colvar/Colvar.h"
#include "core/ActionRegister.h"
```
and remove the `Colvar.h` and `ActionRegister.h` files from your directory?

lautarodibartolo: Yes!

GiovanniBussi:
@lautarodibartolo for some reason GitHub Actions is not running... can you:
- add `colvar` to `src/membranefusion/Makefile` `USE=`
- push again and we can see if GHA runs

lautarodibartolo:
@GiovanniBussi done. 

Yesterday I forgot to push the `user-doc/bibliography.bib` file which have two new references (that I cite in the CV documentation).

GiovanniBussi:
There are a few issues on GHA. I am not sure you can see them, so I paste them here:

on centos7:
```
MemFusionP.cpp:278:108: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, ‘||’, ‘min’ or ‘max’ before ‘vec_double_plus’
MemFusionP.cpp:446:303: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, ‘||’, ‘min’ or ‘max’ before ‘vec_double_plus’
```

If you need this for more recent compilers, you can add some precompiler `#if` to check the openmp version

on codecheck:
```
[membranefusion/MemFusionP.cpp:0] (error) :plmd_astyle: astyle not satisfied
[membranefusion/MemFusionP.cpp:583] (error) :plmd_missing_namespace_module: missing membranefusion namespace
[global_check] (error) :plmd_action_without_examples: action MEMFUSIONP at membranefusion/MemFusionP.cpp:31 has no example
```

Notice that examples are needed for the man page to be built correctly (and are very useful anyway).

To apply astyle, you can compile it (`cd astyle; make`) then apply it with make astyle

Also please add the `membranefusion` namespace

lautarodibartolo:
@GiovanniBussi 

## centos7
I declared a custom omp reduction to be able to use reduction with vectors inside the MemFusionP.cpp file (lines 274 to 276).

`#pragma omp declare reduction(vec_double_plus : std::vector<double> : \
                              std::transform(omp_out.begin(), omp_out.end(), omp_in.begin(), omp_out.begin(), std::plus<double>())) \
                    initializer(omp_priv = decltype(omp_orig)(omp_orig.size()))`

And then used it in this for example:
`#pragma omp parallel for private(ZTailDistance, PositionS_Mem, TailPosition, x, aux) reduction(vec_double_plus:Fs_Mem, sx_Mem, sy_Mem, cx_Mem, cy_Mem)`

Maybe centos7 verification is not taking this into account?

## astyle
Done!

## namespace
Done!

## Example
The error of the example is pointing to the starting point of the documentation section.

`action MEMFUSIONP at membranefusion/MemFusionP.cpp:31 has no example`

In the lines 47 to 68 I wrote an example and linked a repository with an example.

`\par Example`
...
`You can test this CV with an example in this <a href="https://github.com/lautarodibartolo/MemFusion/tree/main/ExampleParallel">GitHub folder</a>.`

GiovanniBussi:
Centos7 has gcc 4.8.5, which does not support OpenMP 4. To keep the code portable I would suggest to make this optional. I think you should enclose the OpenMP 4 pragmas in something like:

```
#ifdef _OPENMP
#if _OPENMP >= 201307
...
#endif
#endif
```
Maybe the outer condition is not necessary. See definitions [here](https://github.com/Kitware/CMake/blob/21e60da5bb0715e09f2feba5930ea2c8bc8180af/Modules/FindOpenMP.cmake#L384-L388)

GiovanniBussi:
And this is still failing as well:
```
[global_check] (error) :plmd_action_without_examples: action MEMFUSIONP at membranefusion/MemFusionP.cpp:31 has no example
```

lautarodibartolo:
I think I was able to fix the examples problem (I was missing the S from `\par ExampleS`)  and the Centos7 issue with your comment. I also added a check at the beginning of the #include <omp.h>.

GiovanniBussi:
@lautarodibartolo thanks! Code is perfect now. Before merging, could you please also consider these points? Thanks!

- is it possible to use a smaller xtc file for the regtest? This is binary, so it will increase by 20 Mb every single repository clone. Once this is solved, you should rebase and force push (or open a new pull request), otherwise the large file will stay forever in our history
- could you please revert the changes you did on the bibliography? It seems you edited a lot of records with a script, please only add the new stuff

lautarodibartolo:
@GiovanniBussi I will make a new pull request with both of the points that you marked solved. From 20Mb to 800Kb the size of the .xtc file, and the fixed .bib 👍

PR #805, opened 2022-03-21
support lazy copy while retrieving atom positions

shazj99:
##### Description

While running Gromacs with Plumed, we noticed that `Sharing data` phrase is very time consuming which took `44.639684` cycles:
```
1934 PLUMED:                                               Cycles        Total      Average      Minimum      Maximum
1935 PLUMED:                                                    1   107.355771   107.355771   107.355771   107.355771
1936 PLUMED: 1 Prepare dependencies                        100001     0.103680     0.000001     0.000001     0.000188
1937 PLUMED: 2 Sharing data                                100001    44.639684     0.000446     0.000376     0.019799
1938 PLUMED: 3 Waiting for data                            100001     0.022178     0.000000     0.000000     0.000186
1939 PLUMED: 4 Calculating (forward loop)                  100001    45.147029     0.000451     0.000426     0.011148
```
Most of the cycles are spending on copying atom positions. 
In our simulation, there are about 120,000 atoms but only ~3200 atoms are used in plumed. Actually, we do not need to copy all atoms' positions in `Sharing data`, but leave it to be done on action calculating which will only copy requested atoms.
In our tests, this way can save a lot which reduce to `0.047627` cycle:

```
1935 PLUMED:                                               Cycles        Total      Average      Minimum      Maximum
1936 PLUMED:                                                    1    61.439300    61.439300    61.439300    61.439300
1937 PLUMED: 1 Prepare dependencies                        100001     0.105186     0.000001     0.000001     0.000199
1938 PLUMED: 2 Sharing data                                100001     0.047627     0.000000     0.000000     0.019528
1939 PLUMED: 3 Waiting for data                            100001     0.017113     0.000000     0.000000     0.000181
1940 PLUMED: 4 Calculating (forward loop)                  100001    46.277080     0.000463     0.000442     0.011822
```
Notice, the lazy copy can only be used in single walker simulation since in multi-simulation it needs to exchange atom positions among different domains before calculating.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/ActionAtomistic.cpp
+++ b/src/core/ActionAtomistic.cpp
@@ -228,7 +228,11 @@ void ActionAtomistic::retrieveAtoms() {
   const std::vector<Vector> & p(atoms.positions);
   const std::vector<double> & c(atoms.charges);
   const std::vector<double> & m(atoms.masses);
-  for(unsigned j=0; j<indexes.size(); j++) positions[j]=p[indexes[j].index()];
+  if (plumed.isLazyCopy()) {
+    atoms.lazyCopyPos(positions, indexes);
+  } else {
+    for(unsigned j=0; j<indexes.size(); j++) positions[j]=p[indexes[j].index()];
+  }
   for(unsigned j=0; j<indexes.size(); j++) charges[j]=c[indexes[j].index()];
   for(unsigned j=0; j<indexes.size(); j++) masses[j]=m[indexes[j].index()];
 }
--- a/src/core/Atoms.cpp
+++ b/src/core/Atoms.cpp
@@ -178,16 +178,27 @@ void Atoms::shareAll() {
   share(unique);
 }
 
+void Atoms::lazyCopyPos(std::vector<Vector>&action_positions, const std::vector<AtomNumber>&indexes) {
+  mdatoms->getPositionsLazy(indexes,natoms,action_positions,positions,forces);
+}
+
 void Atoms::share(const std::set<AtomNumber>& unique) {
   plumed_assert( positionsHaveBeenSet==3 && massesHaveBeenSet );
 
   virial.zero();
-  if(zeroallforces || int(gatindex.size())==natoms) {
-    for(int i=0; i<natoms; i++) forces[i].zero();
-  } else {
-    for(const auto & p : unique) forces[p.index()].zero();
+  // only allow lazy copy for single simulation
+  if(!(int(gatindex.size())==natoms && shuffledAtoms==0))
+      plumed.disableLazyCopy();
+
+  if (!plumed.isLazyCopy()) {
+    // force.zero is time consuming, also do it lazily
+    if (zeroallforces || int(gatindex.size()) == natoms) {
+      for (int i = 0; i < natoms; i++) forces[i].zero();
+    } else {
+      for (const auto &p : unique) forces[p.index()].zero();
+    }
+    for (unsigned i = getNatoms(); i < positions.size(); i++) forces[i].zero(); // virtual atoms
   }
-  for(unsigned i=getNatoms(); i<positions.size(); i++) forces[i].zero(); // virtual atoms
   forceOnEnergy=0.0;
   mdatoms->getBox(box);
 
@@ -196,7 +207,8 @@ void Atoms::share(const std::set<AtomNumber>& unique) {
 
   if(int(gatindex.size())==natoms && shuffledAtoms==0) {
 // faster version, which retrieves all atoms
-    mdatoms->getPositions(0,natoms,positions);
+    if(!plumed.isLazyCopy())
+      mdatoms->getPositions(0,natoms,positions);
   } else {
     uniq_index.clear();
     uniq_index.reserve(unique.size());
--- a/src/core/Atoms.h
+++ b/src/core/Atoms.h
@@ -154,6 +154,8 @@ public:
   void wait();
   void updateForces();
 
+  void lazyCopyPos(std::vector<Vector>&, const std::vector<AtomNumber>&);
+
   void setRealPrecision(int);
   int  getRealPrecision()const;
 
--- a/src/core/MDAtoms.cpp
+++ b/src/core/MDAtoms.cpp
@@ -123,6 +123,8 @@ public:
   void getPositions(const std::vector<int>&index,std::vector<Vector>&positions) const override;
   void getPositions(const std::set<AtomNumber>&index,const std::vector<unsigned>&i,std::vector<Vector>&positions) const override;
   void getPositions(unsigned j,unsigned k,std::vector<Vector>&positions) const override;
+  void getPositionsLazy(const std::vector<AtomNumber>&indexes,int natoms,std::vector<Vector>&action_positions,
+	std::vector<Vector>&positions,std::vector<Vector>&forces) const override;
   void getLocalPositions(std::vector<Vector>&p) const override;
   void getMasses(const std::vector<int>&index,std::vector<double>&) const override;
   void getCharges(const std::vector<int>&index,std::vector<double>&) const override;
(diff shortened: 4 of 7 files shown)


GiovanniBussi:
@shazj99 thanks for contributing this. I am checking the code and I have some small doubt:

- You say "can only be used in single walker simulation" but the check is on `gatindex.size()`. This check: `(int(gatindex.size())==natoms && shuffledAtoms==0)` indeed implies that your code is disabled when (a) there is more than one domain or (b) atoms are not stored in order by the MD code (`shuffledAtoms`). Is this expected?

- I am trying to understand the logic of the code. The current code copies all atom coordinates (possibly scaling them for units) before starting to process the collective variables. Your code copies the needed atoms (possibly scaling them for units) every time an action is processed. This means that, if two actions use the same (large) subset of atoms, the atoms will be copied twice. Is this the desired behavior?

- The current version of the code copies all atoms *except* for the case where there are multiple domains or shuffled atoms. In the latter case, a unique list of used atoms (that thus do not suffer the duplication issue above) is constructed and used both for sending atoms with MPI and for getting positions in place. According to the [comments in the code](https://github.com/plumed/plumed2/blob/388dd76b17cd44af6af5a55440f2e165c4327f22/src/core/Atoms.cpp#L198), I guess that copying all atoms was faster than copying only the used ones. Can you try to reproduce your optimization by enabling this instead of adding new code? We could then use a temporary hidden flag as you are doing now to switch between one behavior and the other, or use some heuristic based on the number of atoms

Let me know if this is clear. Thanks!

shazj99:
> @shazj99 thanks for contributing this. I am checking the code and I have some small doubt:
> 
> * You say "can only be used in single walker simulation" but the check is on `gatindex.size()`. This check: `(int(gatindex.size())==natoms && shuffledAtoms==0)` indeed implies that your code is disabled when (a) there is more than one domain or (b) atoms are not stored in order by the MD code (`shuffledAtoms`). Is this expected?

Yes, the check `(int(gatindex.size())==natoms && shuffledAtoms==0)` is consistent with [here](https://github.com/plumed/plumed2/blob/master/src/core/Atoms.cpp#L197) and [here](https://github.com/plumed/plumed2/blob/master/src/core/Atoms.cpp#L220). I say "single walker simulation" since I'm running gromacs, may it is not preciseness.

> * I am trying to understand the logic of the code. The current code copies all atom coordinates (possibly scaling them for units) before starting to process the collective variables. Your code copies the needed atoms (possibly scaling them for units) every time an action is processed. This means that, if two actions use the same (large) subset of atoms, the atoms will be copied twice. Is this the desired behavior?

Yes, you are right, atoms may be copied multi times, so it is determined by users themselves to turn on lazy copy or not. But in general, these redundant copies is few and do not take long(in our simulations).

> * The current version of the code copies all atoms _except_ for the case where there are multiple domains or shuffled atoms. In the latter case, a unique list of used atoms (that thus do not suffer the duplication issue above) is constructed and used both for sending atoms with MPI and for getting positions in place. According to the [comments in the code](https://github.com/plumed/plumed2/blob/388dd76b17cd44af6af5a55440f2e165c4327f22/src/core/Atoms.cpp#L198), I guess that copying all atoms was faster than copying only the used ones. Can you try to reproduce your optimization by enabling this instead of adding new code? We could then use a temporary hidden flag as you are doing now to switch between one behavior and the other, or use some heuristic based on the number of atoms

Before this, I have already tested the case of `multiple domains or shuffled atoms` in current code version. I notice the `Sharing data` spend a lot on [operating `unique`](https://github.com/plumed/plumed2/blob/master/src/core/Atoms.cpp#L151).  So it may be true that copying all atoms is faster than copying only the used ones which will use `unique` and `unique_index`(I can reproduce it later). 
In my patch, there're 2 reasons why it saves a lot:
1) Only copy used atoms
2) Only zero the `forces` of used atoms([here](https://github.com/plumed/plumed2/blob/72867d534d99944974eb3ae82a1555200e8f96c6/src/core/Atoms.cpp#L194)). I've no idea why `forces.zero()` is so slow, I just do it [together with lazy copy.](https://github.com/plumed/plumed2/blob/72867d534d99944974eb3ae82a1555200e8f96c6/src/core/MDAtoms.cpp#L236)

GiovanniBussi:
I see. Can you confirm that the additional cost that you would like to remove is that of creating the `unique` set? To be quantitative: if you run using unique and then remove the cost of creating unique, you would get the same speedup (except for the speedup in forces.zero())?

Related to the speed of `forces.zero()`, can you try an ad hoc memset? It might be that the compiler has troubled in optimizing the unrolled loops.

shazj99:
> I see. Can you confirm that the additional cost that you would like to remove is that of creating the `unique` set? To be quantitative: if you run using unique and then remove the cost of creating unique, you would get the same speedup (except for the speedup in forces.zero())?
> 
> Related to the speed of `forces.zero()`, can you try an ad hoc memset? It might be that the compiler has troubled in optimizing the unrolled loops.

Hi GiovanniBussi, sorry for the late response. It takes me some time to change the codes using unique. Following is test number:
1. baseline(no changes)
```
1941 PLUMED: 2 Sharing data                                100001    43.375635     0.000434     0.000369     0.019004
```
2. only create unique, but not use it actually
```
1940 PLUMED: 2 Sharing data                                100001    69.417382     0.000694     0.000632     0.029695
```
3. create unique and use it (mdatoms->getPositions & mdatoms->updateForces should be modified)
```
1954 PLUMED: 2 Sharing data                                100001   191.532108     0.001915     0.001694     0.023750
```
4. only create unique,but not use it actually, and use `memset` instead of force.zero()
```
1944 PLUMED: 2 Sharing data                                100001    63.861091     0.000639     0.000596     0.020033
```

From the numbers, we can see that using unique (case 3) will degrade the performance notably, only creating unique but not use it(case 2) will slow down 60%(43.375635->69.417382).

GiovanniBussi:
Thanks this is very interesting. What I understand from this is that iterating through a std:set in order to pick elements from an array (as we do with unique) is much less efficient than doing the same on a std::vector. Hence: unique should be reimplemented as a std::vector. I already made some test with functions to merge sorted std::vectors, and it looks this is also faster than merging std::sets (to my surprise).

What I propose then is that:

- I reimplement all the "unique" machinery with vectors rather then sets, starting from the containers held in ActionAtomistic. If we keep them sorted, merging them is relatively fast using a std::priority_queue.
- Then you can test if the new implementation, used in the "non parallel" case, is as fast as to yours. It should, in theory, and potentially could be faster if you have multiple CVs asking for many atoms.
- If so, we can decide how to enable the use of the new std::vector "unique": always enabled, in parallel, perhaps with a keyword (if its not always convenient) in serial runs. And we can discard your pull request.
- Otherwise we include this pull request.

What do you think?

I don't have time to do this now, I will likely in a couple of weeks at most, but please keep this PR open as a reminder.

shazj99:
> Thanks this is very interesting. What I understand from this is that iterating through a std:set in order to pick elements from an array (as we do with unique) is much less efficient than doing the same on a std::vector. Hence: unique should be reimplemented as a std::vector. I already made some test with functions to merge sorted std::vectors, and it looks this is also faster than merging std::sets (to my surprise).
> 
> What I propose then is that:
> 
> * I reimplement all the "unique" machinery with vectors rather then sets, starting from the containers held in ActionAtomistic. If we keep them sorted, merging them is relatively fast using a std::priority_queue.
> * Then you can test if the new implementation, used in the "non parallel" case, is as fast as to yours. It should, in theory, and potentially could be faster if you have multiple CVs asking for many atoms.
> * If so, we can decide how to enable the use of the new std::vector "unique": always enabled, in parallel, perhaps with a keyword (if its not always convenient) in serial runs. And we can discard your pull request.
> * Otherwise we include this pull request.
> 
> What do you think?
> 
> I don't have time to do this now, I will likely in a couple of weeks at most, but please keep this PR open as a reminder.

Great. My approach can only handle single walker simulation, and as for "parallel" case the `unique` should be reworked anyway. I can test it after you have completed and please let me know if i can be of any help.

Issue #817, opened 2022-04-21
Pass MPI communicator from Python interface


hmcezar:
The Python interface currently can not receive a MPI communicator with `setMPIComm`.
This can be done because `mpi4py` is a cython wrapper itself, which gives access to the communicator with the `.ob_mpi` attribute.

hmcezar:
Fixed in #818

PR #818, opened 2022-04-21
Support for passing mpi4py comm through cython.

hmcezar:
##### Description

As mentioned in #817, currently it's not possible to `setMPIComm` using the Python interface.

##### Target release

I would like my code to appear in the next point releases.

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,6 +132,7 @@ jobs:
         echo "OMPI_MCA_btl_base_warn_component_unused=0" >> $GITHUB_ENV
         echo "OMPI_MCA_btl_base_verbose=0" >> $GITHUB_ENV
         echo "MPIEXEC=mpirun --oversubscribe" >> $GITHUB_ENV
+        pip install mpi4py
     - name: Build PLUMED
       run: |
         ccache -s -M 100M
--- a/conda/py-plumed/meta.yaml
+++ b/conda/py-plumed/meta.yaml
@@ -21,7 +21,6 @@ requirements:
     - python
     - pip
     - cython
-    - mpi4py
   run:
     - plumed
     - python
--- a/conda/py-plumed/meta.yaml
+++ b/conda/py-plumed/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - python
     - pip
     - cython
+    - mpi4py
   run:
     - plumed
     - python
--- a/docker/centos7
+++ b/docker/centos7
@@ -9,7 +9,7 @@ RUN yum -y update \
                    python3 python3-devel python3-pip \
  && pip3 install pillow==8.3.2 \
  && pip3 install numpy==1.19 \
- && pip3 install cython numpy pandas mdtraj mpi4py \
+ && pip3 install cython numpy pandas mdtraj \
  && pip3 install MDAnalysis==1.0.0
 
 RUN useradd -ms /bin/bash plumed
--- a/docker/centos7
+++ b/docker/centos7
@@ -9,7 +9,7 @@ RUN yum -y update \
                    python3 python3-devel python3-pip \
  && pip3 install pillow==8.3.2 \
  && pip3 install numpy==1.19 \
- && pip3 install cython numpy pandas mdtraj \
+ && pip3 install cython numpy pandas mdtraj mpi4py \
  && pip3 install MDAnalysis==1.0.0
 
 RUN useradd -ms /bin/bash plumed
--- a/docker/fedora35
+++ b/docker/fedora35
@@ -8,7 +8,7 @@ RUN yum -y update \
                    lapack-devel blas-devel zlib-devel gsl-devel fftw-devel openmpi-devel boost-devel \
                    python3 python3-devel python3-pip \
  && pip3 install numpy \
- && pip3 install cython numpy pandas mdtraj \
+ && pip3 install cython numpy pandas mdtraj mpi4py \
  && pip3 install MDAnalysis
 
 RUN useradd -ms /bin/bash plumed
--- a/docker/fedora35
+++ b/docker/fedora35
@@ -8,7 +8,7 @@ RUN yum -y update \
                    lapack-devel blas-devel zlib-devel gsl-devel fftw-devel openmpi-devel boost-devel \
                    python3 python3-devel python3-pip \
  && pip3 install numpy \
- && pip3 install cython numpy pandas mdtraj mpi4py \
+ && pip3 install cython numpy pandas mdtraj \
  && pip3 install MDAnalysis
 
 RUN useradd -ms /bin/bash plumed
--- a/docker/rocky8
+++ b/docker/rocky8
@@ -11,7 +11,7 @@ RUN yum -y update \
                    python3 python3-devel python3-pip \
  && pip3 install pillow==8.3.2 \
  && pip3 install numpy==1.19 \
- && pip3 install cython numpy pandas mdtraj \
+ && pip3 install cython numpy pandas mdtraj mpi4py \
  && pip3 install MDAnalysis==1.0.0
 
 RUN useradd -ms /bin/bash plumed
--- a/docker/rocky8
+++ b/docker/rocky8
@@ -11,7 +11,7 @@ RUN yum -y update \
                    python3 python3-devel python3-pip \
  && pip3 install pillow==8.3.2 \
  && pip3 install numpy==1.19 \
- && pip3 install cython numpy pandas mdtraj mpi4py \
+ && pip3 install cython numpy pandas mdtraj \
  && pip3 install MDAnalysis==1.0.0
 
 RUN useradd -ms /bin/bash plumed
--- a/macports/PortfilePython.in
+++ b/macports/PortfilePython.in
@@ -48,7 +48,8 @@ if {${name} ne ${subport}} {
 
     depends_test-append port:py${python.version}-pytest \
                         port:py${python.version}-numpy \
-                        port:py${python.version}-pandas
+                        port:py${python.version}-pandas \
+                        port:py${python.version}-mpi4py
 
     test.cmd            pytest-${python.branch}
     test.target         -v
--- a/macports/PortfilePython.in
+++ b/macports/PortfilePython.in
@@ -48,8 +48,7 @@ if {${name} ne ${subport}} {
 
     depends_test-append port:py${python.version}-pytest \
                         port:py${python.version}-numpy \
-                        port:py${python.version}-pandas \
-                        port:py${python.version}-mpi4py
+                        port:py${python.version}-pandas
 
     test.cmd            pytest-${python.branch}
     test.target         -v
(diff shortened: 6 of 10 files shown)


GiovanniBussi:
@hmcezar thanks for your contribution!

Could you please check if it's possible to add a regression test? You could just add one extra test in [this directory](https://github.com/plumed/plumed2/tree/master/python/test). I think that in our master branch they are run with [pytest](https://github.com/plumed/plumed2/blob/7299737be9cd9c15a60aa188c91f2a2a47327e9b/.github/workflows/ci.yml#L153-L156), so you could be able to just do something like
```
cd python
pytest -v
```

Then, for the test to work on GitHub actions, you should install the `mpi4py` package there as well.

A possible strategy would be to make sure that the test is skipped when the package is not available (by properly adding a try import in the new test you are going to add) and then install the `mpi4py` package in the GHActions jobs where MPI is installed. This can be done by adding after [this line](https://github.com/plumed/plumed2/blob/7299737be9cd9c15a60aa188c91f2a2a47327e9b/.github/workflows/ci.yml#L134) something like
```
pip install mpi4py
```

hmcezar:
@GiovanniBussi I know the test is too simple, but I'm not sure what could be done to really test the implementation.

GiovanniBussi:
@hmcezar the test is fine for start.

I am a bit skeptical in including mpi4py as a dependency in conda. The more dependencies you add, the more you have problems with upgrades on conda.

An ideal solution would be one where:

- having mpi4py installed at compile time is irrelevant
- if mpi4py is installed at run time, the py-plumed package can use it
- otherwise, it can't


For this to work, at least the cython compilation should work without mpi4py. Notice that we have type checkings in the python interface (that's why points are set to, e.g. `<double*>`, [here](https://github.com/plumed/plumed2/blob/7299737be9cd9c15a60aa188c91f2a2a47327e9b/python/plumed.pyx#L97)). However, the type of a MPI communicator is not checked (it might be a typedef for int or a pointer depending on the implementation). Thus, if the only problem is with the conversion, you can use a `<void*>` cast. Would that fix the compilation without the need to install mpi4py before compiling plumed?

hmcezar:
@GiovanniBussi No problem, I reverted the commit.

The CI problem seems to be [in this line](https://github.com/plumed/plumed2/blob/66c00b9b9c3d750325d1dde53fd3607f66b91954/python/plumed.pyx#L172) when I check if the object is MPI.Comm.
Any suggestions on how to make the cython compilation work in this case? Since it's in the Python part, it's not possible to cast it to something else, and the check is needed to make sure the right `cmd` is called.

GiovanniBussi:
Did you try to use a normal `import` [here](https://github.com/plumed/plumed2/blob/66c00b9b9c3d750325d1dde53fd3607f66b91954/python/plumed.pyx#L51) instead of a `cimport`?

In addition, [here](https://github.com/plumed/plumed2/blob/66c00b9b9c3d750325d1dde53fd3607f66b91954/python/plumed.pyx#L138) you could just convert to `<void>`.

hmcezar:
Indeed, I think using `import` instead of `cimport` may solve the issue. Let's see what the CI thinks.

hmcezar:
Well, after implementing the suggested changes I still get the CI errors below.

The `plumed.pyx:136:30: 'MPI' is not a cimported **module**` could be solved by not specifying the `MPI.Comm` of val in the cdef declaration, and that's ok, but the `plumed.pyx:137:14: 'MPI_Comm' is not a type identifier` I don't know how I can solve. I need to declare the buffer as the proper type before passing it as `<void>`, correct?


```bash
[1/1] Cythonizing plumed.pyx

Error compiling Cython file:
------------------------------------------------------------
...
except ImportError:
     HAS_NUMPY=False

try:
     import mpi4py.MPI as MPI
     cimport mpi4py.libmpi as libmpi
            ^
------------------------------------------------------------

plumed.pyx:52:13: 'mpi4py/libmpi.pxd' not found

Error compiling Cython file:
------------------------------------------------------------
...
         self.c_plumed.cmd( ckey, <long*>&abuffer[0], len(abuffer))
     cdef cmd_float(self, ckey, double val ):
         self.c_plumed.cmd_float( ckey, val)
     cdef cmd_int(self, ckey, int val):
         self.c_plumed.cmd_int( ckey, val)
     cdef cmd_mpi(self, ckey, MPI.Comm val):
                             ^
------------------------------------------------------------

plumed.pyx:136:30: 'MPI' is not a cimported module

Error compiling Cython file:
------------------------------------------------------------
...
     cdef cmd_float(self, ckey, double val ):
         self.c_plumed.cmd_float( ckey, val)
     cdef cmd_int(self, ckey, int val):
         self.c_plumed.cmd_int( ckey, val)
     cdef cmd_mpi(self, ckey, MPI.Comm val):
         cdef libmpi.MPI_Comm buffer = val.ob_mpi
             ^
------------------------------------------------------------

plumed.pyx:137:14: 'MPI_Comm' is not a type identifier
```

hmcezar:
Ok, so digging deeper into this issue, it looks like cython supports [conditional compilation](https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#conditional-compilation).

The solution probably should go in this direction. Detect mpi4py prior to the actual cython compilation and `DEF` a variable based on that. Then, a simple `IF` would do the trick.

GiovanniBussi:
@hmcezar this is a possible solution, but I think it would be suboptimal.

In practice, if you do so, one should compile the python extension with a MPI compiler. This is not done by default when you install plumed, and it is not done in the py-plumed conda and Macports packages.

Notice that at the level of the python extension there should be no need to include the `<mpi.h>` header. Indeed, since we are passing MPI communicators by pointer, it is not necessary to know their size. It would be ideal if there was a way to find the address of the MPI communicator so that we can pass it as `void*`, and this should be feasible even if we do not know its `sizeof` (which is the information contained in the `<mpi.h>` file.

GiovanniBussi:
I am not sure how portable this would be, but you might use `id()` to get the address. It looks like the `cdef class Comm` in py4mpi only contains the `ob_mpi` object. So, it is possible (though I didn't test) that by using `id(mpi4py.MPI.COMM_WORLD)` you can get an integer representation of a pointer that could be passed to plumed.

hmcezar:
@GiovanniBussi The problem is that even though I understand that and it should be possible in the sense that we are just passing pointers, we still have to make some checks before knowing which `cfunction` we should call.
The `isinstance` check should be done against a `MPI.comm`, which is what the Python codes will pass to the interface.
To perform this check, we will need `mpi4py`, otherwise we get stuff like:
```bash
Error compiling Cython file:
------------------------------------------------------------
...
     HAS_NUMPY=True
except ImportError:
     HAS_NUMPY=False

try:
     cimport mpi4py.MPI as MPI
            ^
------------------------------------------------------------

plumed.pyx:51:13: 'mpi4py/MPI.pxd' not found

Error compiling Cython file:
------------------------------------------------------------
...
               raise ValueError("ndarrays should be double (size=8), int, or long")
         elif isinstance(val, str ) :
              py_bytes = val.encode()
              cval = py_bytes
              self.c_plumed.cmd( ckey, <const char*>cval )
         elif HAS_MPI4PY and isinstance(val, MPI.Comm):
                                               ^
------------------------------------------------------------

plumed.pyx:171:48: cimported module has no attribute 'Comm'

```

GiovanniBussi:
I think you can use a plain `import` (no `cimport`) to this aim:

```
[bussi@giorgione mpi4py]$ python3
Python 3.9.10 (main, Feb  2 2022, 22:47:17) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mpi4py.MPI as MPI
[giorgione.phys.sissa.it:16911] mca_base_component_repository_open: unable to open mca_oob_ud: libosmcomp.so.3: cannot open shared object file: No such file or directory (ignored)
[giorgione.phys.sissa.it:16907] mca_base_component_repository_open: unable to open mca_oob_ud: libosmcomp.so.3: cannot open shared object file: No such file or directory (ignored)
[giorgione.phys.sissa.it:16907] mca_base_component_repository_open: unable to open mca_btl_openib: libosmcomp.so.3: cannot open shared object file: No such file or directory (ignored)
>>> MPI.Comm
<class 'mpi4py.MPI.Comm'>
```

The only problem that I've seen is that the resulting `MPI.Comm` object do not have a visible `ob_mpi` member.

One thing that I don't like very much is that it looks like just importing mpi4py is slowing down things by initializing MPI. We could thing about options such as only importing that only when reaching a failure in the `isinstance` test [here](https://github.com/plumed/plumed2/blob/99bcb04abc0d2be515ea8eea56eff0a3f1d88e69/python/plumed.pyx#L162). In this way, however, at the first mistake, one would have MPI imported for no reason.

Another option could be to have a specific python method such as `plumed.enable_mpi()`.

Finally, we could try to import mpi4py when the cmd string contains an `MPI` substring (then I would add some check within plumed that we do not add more `...MPI...` commands).

In any case, this can be solved if you find a way to pass a pointer to the `ob_mpi` member without `cimport`ing mpi4py.

hmcezar:
So, I found [this issue](https://bitbucket.org/mpi4py/mpi4py/issues/13/import-failure-on-cython-extensions-that) where the main `mpi4py` developer gives a suggestion on how to get the address of the communicator.

I've implemented it, and things seem to work:
 - If I don't have mpi4py at compile time, it's fine
 - If I have mpi4py at runtime then I can use `setMPIcomm`.

GiovanniBussi:
Thanks!

Regarding slowdown: the problematic case is a user with `mpi4py` installed that however does not want to use it. When plumed module is loaded, `mpi4py` will be loaded as well.

I tried to fix it here: 19cdea3862e0f5561c027fa2c975ba048ef09f78 This is the same thing that we do with `MDAnalysis` below in the same `plumed.pyx` file. Can you double check if things are working for you?

In this branch https://github.com/plumed/plumed2/tree/python-mpi I also added a couple more little change.

If you are ok with this, I would merge it to master. I think it's better not to merge this in v2.7 or v2.8 since it's to be considered a new feature.

GiovanniBussi:
>About merging it to master, in the end, it is up to you guys, but if possible, since I'm using this implementation in [HyMD](https://github.com/Cascella-Group-UiO/HyMD) it would be nice to have it in 2.7 and 2.8, so the code would be compatible with more versions of Plumed.
IMHO, it is not a new feature, just a fix to a functionality that was broken in the Python interface.

Notice that  you can install the plumed python wrappers from the master branch and then load (setting `PLUMED_KERNEL`) an older PLUMED version.

Anyway, I will check if this can be backported easily to 2.8. v2.7 is more difficult since the plumed.pyx file was extensively changed.

GiovanniBussi:
>Thanks for your help in implementing this!
Thanks to you, you actually made most of the job!

hmcezar:
Great, just let me know if you need anything else for merging.

hmcezar:
Thanks!

PR #834, opened 2022-06-29
Made some non-necessary modules default-on rather than always on

GiovanniBussi:
##### Description

I used the following script to analyze module dependences:
```
# run from plumed2/src
for type in always default-on default-off ; do
echo Type $type
for mod in $(grep "$type" */module.type  | sed "s|/.*||") ; do
used_in=$(git grep 'include *"'$mod'/' | sed "s|/.*||" | sort | uniq)
echo $mod: $used_in
done | grep -v ":$"
echo "###"
done
```

The result is the following:
```
Type always
asmjit: lepton
blas: lapack
core: adjmat analysis bias cltools colvar crystallization dimred drr eds fisst function funnel generic gridtools isdb logmfd manyrestraints mapping maze membranefusion multicolvar opes pamm piv pytorch reference s2cm sasa secondarystructure setup tools vatom ves vesselbase
gridtools: analysis dimred multicolvar
lapack: tools
lepton: core function tools ves
reference: analysis colvar dimred function mapping secondarystructure
tools: adjmat analysis bias cltools colvar core crystallization dimred drr eds fisst function funnel generic gridtools isdb mapping maze multicolvar opes pamm piv reference s2cm setup tools vatom ves vesselbase
vesselbase: adjmat analysis crystallization gridtools manyrestraints mapping multicolvar secondarystructure
###
Type default-on
analysis: dimred vesselbase
bias: analysis drr eds fisst funnel isdb logmfd maze opes ves vesselbase
cltools: drr mapping ves
colvar: funnel isdb maze membranefusion piv s2cm
function: annfunc isdb pytorch
molfile: cltools
multicolvar: adjmat analysis crystallization pamm
xdrfile: cltools generic
###
Type default-off
adjmat: pamm
###
```

In short:
- tools and core are used basically everywhere, so they should be always on
- asmjit blas lapack and lepton are, directly or indirectly, used in tools and core, so they should be always on as well
- gridtools, reference, and vesselbase can be made default-on. For comparison, also `bias` is now default-on, and is used in many other modules

I would then proceed with merging this in master.

See also https://github.com/spack/spack/pull/29743#issuecomment-1169326528

##### Target release

### DIFF
--- a/src/gridtools/module.type
+++ b/src/gridtools/module.type
@@ -1,1 +1,1 @@
-always
+default-on
--- a/src/reference/module.type
+++ b/src/reference/module.type
@@ -1,1 +1,1 @@
-always
+default-on
--- a/src/vesselbase/module.type
+++ b/src/vesselbase/module.type
@@ -1,1 +1,1 @@
-always
+default-on


marcodelapierre:
Hi @GiovanniBussi ,

adding comments here coming from the Spack PR https://github.com/spack/spack/pull/29743.

To provide you with some extra context and recap, in that PR I have added more fine grained support to enable/disable specific modules for Plumed.
Right now these are the options that I implemented, using a Spack multi-valued variant:
- all
- reset (i.e. defaults)
- none
- custom list of modules

To build the list of modules for the last case, I started from the list of modules that can be found in the Plumed user docs, for each minor version, e.g. for 2.8.x: 
https://www.plumed.org/doc-v2.8/user-doc/html/mymodules.html

I analysed the lists for all Spack supported minor versions of Plumed, i.e. 2.4.x to 2.8.x.
I then added version conflicts for each module accordingly.

In my opinion, if there are any additional properties/logics/dependencies that need to be declared for modules (e.g. dependencies as you describe above), it would be good to have them declared in a visible location, e.g. the Plumed user docs, so that non-Plumed developers like me do not have to dig in the source code.  In this way the inspection is quicker and less error prone.

Long story short, I am happy to get more information on the logics of the Plumed modules ,so that I can implement it correctly in the Spack recipe.
One question is...for how many versions shall we add support for this? If I had to do it, I would suggest to focus on 2.7 and 2.8, to keep it manageable.

One final comment on this topic.
I have noticed that `--optional-modules=none` results in a broken build. I am not sure about the root cause.. maybe some modules need to be always built?

Keen on hearing your feedback and thoughts.

Thanks, Marco

GiovanniBussi:
I am not sure it is relevant disabling all modules **in spack**. Who is willing to build PLUMED without any CV? This is harmless, but certainly among those modules that are on by default (e.g., colvar or bias) there are many non trivial dependencies. Tracking them is complicated.

Personally, I would aim at something like:
- base build with all default modules
- possibility to add optional modules one by one or in combination

There is a single dependency to track then: `adjmat: pamm`.

If you think it's relevant, we can add to the manual a page with this dependency, computed with the script above (so as to make sure that we detect if other dependencies appear later). Or even simpler write in the manual (where we explain how to enable and disable modules) that one could use a command such as `make moduledeps` (to be added) to retrieve the dependencies for the current version.

Regarding failures to build with `--enable-modules=none`, I just tried and it works for me. This is not to say that this build is useful for a final user, but that I don't find any problem.

Thanks!

marcodelapierre:
On the `none` option I agree, and so I have updated the Spack PR to remove it as a legal value for the variant `optional_modules`; this makes the variant compliant with your suggestion above, plus also the `reset` value, that only builds the default values.

https://github.com/spack/spack/pull/29743

Interesting that you can build with `none`, but I cannot. I might not be able to investigate any time soon though, also considering we agree the `none` option is not relevant to users.

On the dependencies:
> There is a single dependency to track then: adjmat: pamm.

I agree, it would be certainly good to declare the dependency in the manual, e.g. in the page
https://www.plumed.org/doc-v2.8/user-doc/html/mymodules.html

Can you clarify whether it is `adjmat` that depends on `pamm`, or viceversa?
I think I should be able to enforce this in the Spack recipe.

GiovanniBussi:
Thanks @marcodelapierre !

Actually pamm depends on adjmat (I realized the way I wrote it means the opposite in Makefile syntax).

marcodelapierre:
Done! added dependency of `pamm` on `adjmat` in https://github.com/spack/spack/pull/29743

PR #841, opened 2022-07-08
configure: avoid non-portable test == operator

drhpc:
The single = works as well and also with shells other than bash
and friends.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


drhpc:
Also … I got a change in the pipeline where I need to ensure that the RPATH setting reaches the plumed main binary. It touches configure, too, so it would be nice if this simple one here could be merged first.

GiovanniBussi:
Thanks! This looks like a portability fix, so I plan to backport it (manually) to version 2.7. Does it make sense?

Regarding [this point](https://github.com/plumed/plumed2/pull/841#issuecomment-1179352149), can you check if this is already fixed? I think I fixed it in v2.8 branch, so that it will end up in v2.8.1 (not released yet, though).

PR #832, opened 2022-06-22
Update to EnvironmentSimilarity

PabloPiaggi:
##### Description

Update the EnvironmentSimilarity multicolvar that is a part of the crystallization module. Changes:
- Allow the specification of chemical species in reference environments and in the simulation box through pdb files. This is useful to compare environments with more than one chemical species. As a result of this modification one can target complex crystal structures with any number of chemical species.
- A regtest for the new feature has been added
- Bug fix - The code now checks if environment pdb files exist before reading them. Thanks to @sigbjobo for mentioning that I should fix this.

This feature is **not** required for the upcoming Masterclass.

@gtribello do you have any suggestions regarding functionality or input syntax? Thank you.

##### Target release

I would like my code to appear in the master branch

##### Type of contribution

- [x] changes to code or doc authored by PLUMED developers, or additions of code in the core or within the default modules
- [x] changes to a module not authored by you

##### Copyright

- [x] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

##### Tests

- [x] I added a new regtest or modified an existing regtest to validate my changes.
- [x] I verified that all regtests are passed successfully on [GitHub Actions](https://github.com/plumed/plumed2/actions).

PR #843, opened 2022-07-11
Fix bug in Environment Similarity - check for pdb files before read 

PabloPiaggi:
##### Description

Existence of PDB files for reference environments was not checked before reading them. Thus, absence of reference environments would not raise an error but rather give a wrong result. This PR checks for PDB files before reading them and raises an error if they are not found.

In the master branch this bug is fixed by #832 .

##### Target release

##### Type of contribution

##### Copyright

##### Tests


PabloPiaggi:
@GiovanniBussi Should I open another PR to fix this in v2.7 ? Thanks.

GiovanniBussi:
@PabloPiaggi sorry I think I never replied to [this](https://github.com/plumed/plumed2/pull/843#issuecomment-1180616090).

If you want to have this fixed in v2.7 as well, you should do a new pull request addressing that branch. Thanks!

PR #845, opened 2022-07-16
Masterclass 22.12 docs

PabloPiaggi:
##### Description

These are the instructions for Masterclass 22.12

##### Target release

I would like my code to appear in the **master** branch

##### Tests

### DIFF
--- a/user-doc/tutorials/a-masterclass-22-12.txt
+++ b/user-doc/tutorials/a-masterclass-22-12.txt
@@ -35,7 +35,57 @@ git clone https://github.com/PabloPiaggi/masterclass-22-12
 
 \section masterclass-22-12-theory Summary of theory
 
-\section masterclass-22-12-system The system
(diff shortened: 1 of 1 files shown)


PabloPiaggi:
With this PR and the repo [in this link](https://github.com/PabloPiaggi/masterclass-22-12) we should be all set for Monday.

I believe right now a test is failing due to a missing astyle of a previous commit

invemichele:
@PabloPiaggi I just noticed that there is a visualization issue in the manual page https://www.plumed.org/doc-master/user-doc/html/masterclass-22-12.html 
Probably it is due to the script that changes the plumed input to the new convention (no repeated name at the end of action, and `mylabel:` instead of `LABEL=mylabel`) I suggest you manually do these changes, this will probably fix the page

PabloPiaggi:
Thanks @invemichele for pointing out what the issue might be! I will try to correct it now.

PR #848, opened 2022-08-10
Fixed division by zero when SIGMA_MEAN0=0 in Metainference

### LINKED ISSUE #847: SIGMA_MEAN0=0.0 in OUTLIERS noisetype for Metainference
When performing metainference simulations, it is possible to set `SIGMA_MEAN0=0.0` while using `NOISETYPE = OUTLIERS/MOUTLIERS`.

However, since there's a division by `sigma_mean2_`, there might be divisions by zero, resulting in `NaNs`:
https://github.com/plumed/plumed2/blob/beaf4e60640bd179418d0ae8430ac71a3a7d7a99/src/isdb/Metainference.cpp#L1337-L1357

I think there should be a check for `SIGMA_MEAN0=0.0` or, even better, maybe work in the limiting case where `t -> 0` in the above snippet? Since `sigma_mean2_` is a square, we can assume a limit from the right and get `dit/sm2 -> 0`.

hmcezar:
##### Description

This PR fixes #847.

I'm not happy with the `if`s inside the loops, and fixing this would be easy for the `SP` versions, but for `SPE` there might be zeros for just some of the replicas, so I'm not sure what would be a good approach in this case.

Let me know if you have any better ideas.

##### Target release

I would like my code to appear in release 2.7

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/isdb/Metainference.cpp
+++ b/src/isdb/Metainference.cpp
@@ -934,7 +934,12 @@ double Metainference::getEnergySP(const std::vector<double> &mean, const std::ve
     for(unsigned i=0; i<narg; ++i) {
       const double dev = scale*mean[i]-parameters[i]+offset;
       const double a2 = 0.5*dev*dev + ss2;
-      ene += std::log(2.0*a2/(1.0-std::exp(-a2/sm2)));
+      if(sm2 > 0.0) {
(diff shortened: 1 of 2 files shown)


carlocamilloni:
Thanks!

hmcezar:
No problem!
If possible, just backport it to 2.7 and 2.8 please :smile:

PR #856, opened 2022-08-30
`long long unsigned` for step in VES MD

invemichele:
##### Description

Since we are here...
This PR changes the step in  [VES PES MD](https://www.plumed.org/doc-v2.7/user-doc/html/ves_md_linearexpansion.html) from `unsigned` to `long long unsigned`, to allow for longer simulations.

@valsson we had to do this change to perform some of the simulations in this paper https://arxiv.org/abs/2208.04942, but then we never pushed it to the official repo. (cc @dhimanray, @valeriorizzi, @narjesansari)

A similar change could be done also in `src/cltools/pesmd.cpp` that now uses an `int` to count steps.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Yes it's a random error... thanks!

PR #868, opened 2022-09-15
Fix grid spacing in MetaD/PBMetaD/sum_hills

GiovanniBussi:
##### Description

I think there is a mistake in how grid spacing is used now. For instance, in METAD

```
GRID_MIN=0 GRID_MAX=10 GRID_SPACING=1
```
results incorrectly in 11 bins. The reason is that to compute the number of bins we now use this formula:
```
unsigned n=((b-a)/gspacing[i])+1;
```
but I think the correct one is the following formula:
```
unsigned n=std::ceil(((b-a)/gspacing[i]));
```

Notice that the two formulas are different only if the grid spacing exactly divides `max-min`, in which case the second formula returns `(max-min)/spacing` whereas the first formula returns `(max-min)/spacing+1`. In all other cases they should be identical.

I think this is a bug fix, so I would like to submit it to v2.7, even though it changes numerically the behavior of some regtests due to the fact that the chosen grid is different. I would apply this to:
- MetaD
- PBMetaD
- sum_hills

Merging to v2.8 should then be done manually due to same numerical change we made in the shape of Gaussians.

##### Target release


carlocamilloni:
I would apply it only to master, is a small change and I don’t think it makes sense to change the behavior of former versions at this point

GiovanniBussi:
Yes probably it's better. On one hand, this is a bug affecting very few people (those using MetaD on integer variables like lambda dynamics). On the other hand, fixing it would change numerical reproducibility for everyone else using grids. I will bring it forward to master only

GiovanniBussi:
Merged directly to master (v2.9) bfc648c42e5aff78ac51b640b8deab5caef7d3c7

PR #887, opened 2022-11-24
New SAXS implementation

riccardocapelli:
##### Description

We added a new kind of coarse-grained calculation of SAXS spectra, where the protein structure is represented by one bead per residue. In such representation we add the possibility to add a solvent exposure correction to SAXS intensity; the evaluation of the SASA is based on the LCPO module written by @andrea-arsiccio, and minimally modified here.

The current implementation has been tested with both CPU and GPU (arrayfire) and is retro-compatible with old input files for PARAMETER, MARTINI, and ATOMISTIC SAXS calculations.

##### Target release

2.9

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@carlocamilloni just one comment: this is "copy and paste" of @andrea-arsiccio 's code, or are you basing on that code?

Maybe, if there are shared parts, and if Andrea agrees, we could think about moving them to the core of the code os that both your module and his module can use them. To avoid duplication.

andrea-arsiccio:
Hi all,
all works fine for me, I am very happy if my portion of code can be useful
for other modules and I have nothing against moving parts of my scipt to
the core of the code.

Thanks a lot!

Best,

Andrea

carlocamilloni:
@GiovanniBussi @riccardocapelli @andrea-arsiccio at the moment is a stripped down version minimal version, but yes in principle we can think of having a base module on which to build.

Issue #651, opened 2020-11-23
[Gromacs 2020.2] Issue when trying to run SAXS on GPU


RemiLacroix-IDRIS:
Hello,

One of our users ran into some issues when trying to run Gromacs 2020.2 patched with Plumed 2.6.1 while configuring SAXS to run on GPU (using the ArrayFire library).

You can find a small test case here: https://filesender.renater.fr/?s=download&token=f3a932e9-e74e-4ec4-b0ed-d768551aec5b (link valid until December 23rd).

The error is as follow:
```
Program:     gmx mdrun, version 2020.2-MODIFIED
Source file: src/gromacs/gpu_utils/devicebuffer.cuh (line 207)
Function:    clearDeviceBufferAsync(ValueType**, size_t, size_t, CommandStream) [with ValueType = float; DeviceBuffer<ValueType> = float*; size_t = long unsigned int; CommandStream = CUstream_st*]::<lambda()>
MPI rank:    1 (out of 4)

Assertion failed:
Condition: stat == cudaSuccess
Couldn't clear the device buffer
```
The full log is attached: [output.log](https://github.com/plumed/plumed2/files/5585224/output.log).

There is no error if SAXS runs on CPU (i.e. when removing the `GPU` keyword from `Rep*/plumed-saxsCG.dat` files).

There is no error if I set the `CUDA_VISIBLE_DEVICES` environment variable so that each of the 4 MPI tasks only see one distinct GPU (note that in this case I have to modify the `Rep*/plumed-saxsCG.dat` files so that `DEVICEID=0` is used everywhere).

I hope that you can shed some light on this issue. Let me know if you need more information.

Best regards,
Rémi

carlocamilloni:
Dear Rémi

I don’t know, but my guess is that if all the replica run on the same node, still they try to use all the same GPU (there isn’t an automatic distribution of the GPU at the moment, but it should be possible to learn from GROMACS which GPU is used and use the same (I noted this down for future development). A possibility in alternative to use CUDA_VISIBLE_DEVICES is to sed different DEVICEID to the different replicas using the replica syntax in plumed (I am not sure, but it should work) something like 

DEVIDEID=@replicas:{0 1 2 3}

Best 

Carlo

RemiLacroix-IDRIS:
Dear Carlo,

Thanks for your answer!

> my guess is that if all the replica run on the same node, still they try to use all the same GPU

[...]

PR #893, opened 2023-01-12
Added the option to set the gpuDeviceId from the MD engine

carlocamilloni:


carlocamilloni:
This should fix #651 , we have the same problem on Marconi100 because the rank=0 of gromacs assign itself the deviceId=3 while the default assignment we had was 0 thus resulting in a race condition due to hardware binding

PR #894, opened 2023-01-13
fix(pathtools): Fix suggested lambda as per masterclass example

rebelot:
##### Description

Fix suggested lambda as per masterclass example: lambda = 2.3/MSD

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/mapping/PathTools.cpp
+++ b/src/mapping/PathTools.cpp
@@ -31,6 +31,7 @@
 #include <string>
 #include <vector>
 #include <iostream>
+#include <cmath>
 
 namespace PLMD {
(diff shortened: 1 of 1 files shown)


carlocamilloni:
@gtribello @GiovanniBussi anything wrong with this?

GiovanniBussi:
Perhaps @gtribello wants to double check.. Anyway, the fix seems right to me. The original code had two problems:
1. It computes the distance (not squared). It should be squared because lambda is multiplied times the squared distance
2. There's a missing parenthesis in taking the average.

So, I think this is right and should even be backported

PR #903, opened 2023-02-14
Made PIV implementation thread safe

GiovanniBussi:
##### Description

This is meant to solve the non-thread-safety of PIV variables (see [this comment](https://github.com/plumed/plumed2/pull/370#issuecomment-419094411)). Specifically:
- Before this fix, a number of objects are allocated as `static` and shared between all instances of the PIV object.
- After this fix, these objects are only shared between instances of the PIV object created within the same PlumedMain object

As a result, it should be possible to allocate multiple PlumedMain objects without interference (e.g., for multithreading).

To minimize code changes, I defined a [struct](https://github.com/plumed/plumed2/blob/d927afee7f3427c1559939e55e437d71d5ba26a1/src/piv/PIV.cpp#L213-L220) containing all the required data structures. The first PIV instance [allocates an owning pointer](https://github.com/plumed/plumed2/blob/d927afee7f3427c1559939e55e437d71d5ba26a1/src/piv/PIV.cpp#L312) (unique_ptr) to this structure. The following PIV instances use a [raw pointer](https://github.com/plumed/plumed2/blob/d927afee7f3427c1559939e55e437d71d5ba26a1/src/piv/PIV.cpp#L317). [References](https://github.com/plumed/plumed2/blob/d927afee7f3427c1559939e55e437d71d5ba26a1/src/piv/PIV.cpp#L775-L781) are then used to refer to the elements on the `struct` so that the original code is largely unchanged. 

@SilvioPipolo please let me know if there is any objections, or I will merge this in a couple of weeks.

##### Target release

I would like my code to appear in release __v2.9__

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/piv/PIV.cpp
+++ b/src/piv/PIV.cpp
@@ -23,6 +23,7 @@ along with plumed.  If not, see <http://www.gnu.org/licenses/>.
 #include "tools/PDB.h"
 #include "tools/Pbc.h"
 #include "tools/Stopwatch.h"
+#include "core/ActionSet.h"
 
 #include <string>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


SilvioPipolo:
Thank you Giovanni, green light from my side,
Best,
Silvio

Issue #83, opened 2014-02-19
Compute collective variables in MD engine


GiovanniBussi:
This is something we discussed long ago at the first developer meeting, I add it here even if I think the priority is presently low.

We could add the possibility to compute CVs directly in MD engine. They could be passed  as:
plumed_cmd("setCV pippo",&pippo); // this will be passed to plumed
plumed_cmd("setCVforce pippo",&pippoforce); // this will be retrieve from plumed
plumed_cmd("calc");
The MD code would be then responsible to propagate the force to the atoms. This would open to these possibilities:
- Calculation of variables depending on electronic density in DFT codes.
- Calculation of very expensive variables inside MD engine taking advantage of e.g. MD neighbor list or optimized routines.
- Implementation of a dummy MD code to test free-energy methods with simple low dimensional models (e.g. Langevin models).

yuanchaohu:
That would be good to generalize Plumed well over various research fields

PR #909, opened 2023-02-28
Added cmd("isExtraCVNeeded")

GiovanniBussi:
##### Description

This is to complete the implementation of extra CVs (computed in the MD code), see #83 .

What I add here is an extra plumed cmd to allow the MD code to know if an extra CV is needed at the step or not.

This can be used to enable:
- multiple time stepping
- monitoring with large STRIDE
 
possibly useful for expensive CVs.

##### Target release

### DIFF
--- a/regtest/basic/rt-make-extracv/main.cpp
+++ b/regtest/basic/rt-make-extracv/main.cpp
@@ -22,10 +22,12 @@ int main(){
   plumed->cmd("readInputLine","d: DISTANCE ATOMS=1,2");
   plumed->cmd("readInputLine","e: EXTRACV NAME=extra");
   plumed->cmd("readInputLine","e2: EXTRACV NAME=extra2");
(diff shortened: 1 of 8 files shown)


thangckt:
Dear Prof. @GiovanniBussi 

Can you give an example of using this feature with an existing parameter computed in LAMMPS?
Thank you so much

GiovanniBussi:
This feature is implemented in Plumed, but I don't think it is used by the current LAMMPS PLUMED interface.

To give you an example, this feature is used in the GROMACS PLUMED interface to pass lambda (alchemical variable) to a plumed so as to use it as a collective variable

thangckt:
Dear Prof. @GiovanniBussi

Thank you so much for the information.
I am looking for if there is a plan to update LAMMPS PLUMED interface.

Issue #897, opened 2023-01-30
Centos 7 and rt-make-c11


carlocamilloni:
There is a build issue on centos7 @GiovanniBussi do you have any clue?

plumed/wrapper/Plumed.h:1531:5: error: expected expression before ‘const’
and other

GiovanniBussi:
Looks like the testing system does not understand anymore that Centos7 does not support C11.

It is weird that 2 weeks ago it was working. At [this build](https://github.com/plumed/plumed2/actions/runs/3932699625/jobs/6725570096):
```
make[3]: Entering directory `/home/plumed/plumed2/regtest/basic/rt-make-c11'
Mon Jan 16 17:53:25 UTC 2023
Running regtest in /home/plumed/plumed2/regtest/basic/rt-make-c11
++ Test type: make
++ Arguments: 
++ Processors: 0
/home/plumed/plumed2/regtest/basic/rt-make-c11/tmp
NOT_APPLICABLE (CC does not support c11)
```

Do you know when it started failing?

carlocamilloni:
yes I think it happen when I regenerated the configure con my Mac

PR #910, opened 2023-02-28
Pytorch documentation fix and update

luigibonati:
##### Description

Small bug fixes and updates related to the Pytorch module documentation.

- I noticed that there was a bug in the documentation of the pytorch module. Due to recent changes in section headers, only the first two sections were shown.
- I also updated some instructions on how to verify that the configuration with libtorch is succesful. 
- I updated the module doc and the CI instructions with the link to the latest libtorch version (1.13.1). Before the Long Term Support version (1.8.2) was recommended and used in the CI, but it has been [discontinued](https://pytorch.org/blog/pytorch-enterprise-support-update/).
- I updated the behavior of the PLUMED_CHECK_CXX_PACKAGE function in configure  when linking against multiple libraries. Now the check is done against all the libraries at once, rather than trying by adding one at a time (which was also creating confusion in the config.log)

##### Target release

I would like my code to appear in release 2.9.

##### Type of contribution

##### Copyright

##### Tests

- [] I verified that all regtests are passed successfully on [GitHub Actions](https://github.com/plumed/plumed2/actions).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,8 +81,8 @@ jobs:
         # added by luigibonati
         mkdir -p $HOME/opt/
         cd $HOME/opt/ # GB: install in $HOME/opt/libtorch
-        wget --no-check-certificate https://download.pytorch.org/libtorch/lts/1.8/cpu/libtorch-cxx11-abi-shared-with-deps-1.8.2%2Bcpu.zip ;
-        unzip libtorch-cxx11-abi-shared-with-deps-1.8.2+cpu.zip ;
+        wget --no-check-certificate https://download.pytorch.org/libtorch/cpu/libtorch-cxx11-abi-shared-with-deps-1.13.1%2Bcpu.zip ;
+        unzip libtorch-cxx11-abi-shared-with-deps-1.13.1+cpu.zip ;
         echo "LIBRARY_PATH=$PWD/libtorch/lib:$LIBRARY_PATH" >> $GITHUB_ENV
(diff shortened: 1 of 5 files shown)


GiovanniBussi:
@luigibonati thanks! A couple of points before merging:
- Can you confirm that the change to configure.ac only affects the tests done for libtorch? I think that all other calls to `PLUMED_CHECK_CXX_PACKAGE` are missing the fifth argument (`multiple_libraries`), but I wanted to double check that I understood correctly.
- I noticed you are also using a different autoconf version (I guess the ubuntu ones). I can fix this after merge. There was a problem with another autoconf version recently (#897) so I try to stick to the current one.

luigibonati:
> @luigibonati thanks! A couple of points before merging:
> 
> * Can you confirm that the change to configure.ac only affects the tests done for libtorch? I think that all other calls to `PLUMED_CHECK_CXX_PACKAGE` are missing the fifth argument (`multiple_libraries`), but I wanted to double check that I understood correctly.

Yes, exactly. I had added the optional argument multiple_libraries to allow linking to more than one library at the same time, instead of trying each one individually as was the case with the other libraries. So this change will only involve libtorch. In any case, the behavior is the same, I just simplified it by having the test run with all the specified libraries, instead of first trying only the first library, then the first and second, and so on.

> * I noticed you are also using a different autoconf version (I guess the ubuntu ones). I can fix this after merge. There was a problem with another autoconf version recently ([Centos 7 and rt-make-c11 #897](https://github.com/plumed/plumed2/issues/897)) so I try to stick to the current one.

Sounds good! yes I used the ubuntu one

PR #931, opened 2023-04-27
New warnings for METAD  adn WALKERS_MPI

### LINKED ISSUE #926: Need an informative errore when a user uses METAD WALKERS_MPI but plumed was compiled without MPI
As title, this issue has been addressed [in this conversation](https://groups.google.com/g/plumed-users/c/Q1TJ1A4aG_Y/m/pOewzRn2AQAJ).
I'm opening a branch to add the warning.

Iximiel:
##### Description

This should resolve #926: 

METAD now throws if the user ask for WALKERS_MPI without a MPI installation and also if the MPI routines are not set up.

I wanted the changes testable so:

- I added a static variable to the Communicator to have the user access to the information "plumed is compiled with MPI" at runtime.
- I created a small non standard reg test to test the error thrown without using external tools and that tests exactly the needed messages
- I added the thrown exceptions with the messages set up to pass the new test

This is a draft because I wanted to show the solution that I have found, and I have not added the new test to the github workflow

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -877,7 +877,13 @@ MetaD::MetaD(const ActionOptions& ao):
 
   // MPI version
   parseFlag("WALKERS_MPI",walkers_mpi_);
-
+  
+  ///TODO: set up a test that simulates an external program that trigger this error when MPI is not initialized
+  //If this Action is not compiled with MPI the user is informed and we exit gracefully
+  if(walkers_mpi_)
+    plumed_assert(Communicator::initialized()) << "Invalid walkers configuration: WALKERS_MPI"
+  " flag requires MPI compilation and the correct initialization of the communicator.";
+  
   // Flying Gaussian
   parseFlag("FLYING_GAUSSIAN", flying_);
 
--- a/src/bias/MetaD.cpp
+++ b/src/bias/MetaD.cpp
@@ -878,11 +878,11 @@ MetaD::MetaD(const ActionOptions& ao):
   // MPI version
   parseFlag("WALKERS_MPI",walkers_mpi_);
   
-  ///TODO: set up a test that simulates an external program that trigger this error when MPI is not initialized
   //If this Action is not compiled with MPI the user is informed and we exit gracefully
-  if(walkers_mpi_)
-    plumed_assert(Communicator::initialized()) << "Invalid walkers configuration: WALKERS_MPI"
-  " flag requires MPI compilation and the correct initialization of the communicator.";
+  if(walkers_mpi_){
+    plumed_assert(Communicator::PlumedHasMPI) << "Invalid walkers configuration: WALKERS_MPI flag requires MPI compilation";
+    plumed_assert(Communicator::initialized()) << "Invalid walkers configuration: WALKERS_MPI needs the communicator correctly initialized.";
+  }
   
   // Flying Gaussian
   parseFlag("FLYING_GAUSSIAN", flying_);
--- a/src/tools/Communicator.cpp
+++ b/src/tools/Communicator.cpp
@@ -28,6 +28,13 @@
 
 namespace PLMD {
 
+const bool Communicator::PlumedHasMPI=
+#ifdef __PLUMED_HAS_MPI
+    true;
+#else
+     false;
+#endif
+
 Communicator::Communicator()
 #ifdef __PLUMED_HAS_MPI
   : communicator(MPI_COMM_SELF)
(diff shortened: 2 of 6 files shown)


GiovanniBussi:
Thanks @Iximiel ! A few comments:

- You can directly access the macro __PLUMED_HAS_MPI. Notice that any code including `Communicator.h` is currently forced to define all the relevant macros or it will not work.
 
- In general, I would avoid static member variables, since they can be modified by mistake with const_cast. I think using a function that returns a bool is more robust.

- An alternative way is to use directly Communicator::initialized() which also check if MPI is initialized. I think this is even better since you want to make sure that this is true

- Finally, in src we only keep source code and not tests. All tests are supposed to go in regtests, so that (a) they are executed on GitHub actions and (b) they are not installed

src/bias/MetaD.cpp
```diff
@@ -877,7 +877,13 @@ MetaD::MetaD(const ActionOptions& ao):
 
   // MPI version
   parseFlag("WALKERS_MPI",walkers_mpi_);
-
+  
+  //If this Action is not compiled with MPI the user is informed and we exit gracefully
+  if(walkers_mpi_){
+    plumed_assert(Communicator::PlumedHasMPI) << "Invalid walkers configuration: WALKERS_MPI flag requires MPI compilation";
+    plumed_assert(Communicator::initialized()) << "Invalid walkers configuration: WALKERS_MPI needs the communicator correctly initialized.";
```
GiovanniBussi, line 884: Here I think you could just remove the first test.

Beside this, the way to raise the warnings is correct!!

Iximiel: In my first draft i had a single error message with "MPI is not initialized or plumed is not compiled with MPI", but I think is more clear for the user to have two more specialized messages, and easier to test for the developers

GiovanniBussi: I think you are right

Iximiel:
> Thanks @Iximiel ! A few comments:
> 
>     * You can directly access the macro __PLUMED_HAS_MPI. Notice that any code including `Communicator.h` is currently forced to define all the relevant macros or it will not work.

Yes, but I set up the test as a completely external code, so when I compile the test I'm not seeing the compile-time definitions of PLUMED, or at least, I do not know how to access to them. 

> 
>     * In general, I would avoid static member variables, since they can be modified by mistake with const_cast. I think using a function that returns a bool is more robust.

My bad, now `Communicator::PlumedHasMPI()` is a function

> 
>     * An alternative way is to use directly Communicator::initialized() which also check if MPI is initialized. I think this is even better since you want to make sure that this is true

when `Communicator::initialized()` returns false I don't know if it is  because PLUMED is compiled without MPI or the MPI communication is not initialized

> 
>     * Finally, in src we only keep source code and not tests. All tests are supposed to go in regtests, so that (a) they are executed on GitHub actions and (b) they are not installed

the test is in not in src/, but in unittest. Since is very far from the standard you use in the regtest directory, I though it was a good idea to keep it separated before the full merge

carlocamilloni:
Hi @Iximiel, this is good with me! as a side note, the test that is failing now is a "style check" one. We enforce some basic styling in the code and you can get it fixed by running "make astyle" in src/ and commit again

GiovanniBussi:
@Iximiel regarding this point:
> Yes, but I set up the test as a completely external code, so when I compile the test I'm not seeing the compile-time definitions of PLUMED, or at least, I do not know how to access to them.

You are right. However, since we use some of the compile-time definitions of PLUMED in header files (actually very few, but `__PLUMED_HAS_MPI` is currently one of those), in order to properly compile source files including those header files you need to use the proper compile-time definitions. This is an (unsolved) problem for anyone willing to use plumed as an external library. Indeed, I am not aware of general solutions to this problem. A partial solution is that all those macros are included in the installed file `lib/pkgconfig/plumedInternals.pc`, but this is only for packages using pkgconfig. I guess for packages using cmake it should be straightforward.

Given this, once you bring your test within the regtest library, this is designed to make the macros available when compiling regtests.

So, I think it's fine using these macros in regtests. But it's also ok (and I think better) to write a small function that makes the bool available for outside usage.

GiovanniBussi:
> My bad, now Communicator::PlumedHasMPI() is a function

another stylistic convention that we have is: classes upper case, functions lower case. So perhaps it's more consistent to have `Communicator::plumedHasMPI()`

:-)

Iximiel:
I have moved the new test in the regtest folder and I did some addition to make it work via `make check` in the main directory and with `make` in the regtest directory with an environment ready to launch an installed plumed, moreover is compatible with `make reset`

I runned astyle on the src, I think this can exit the draft state

PR #935, opened 2023-05-12
Action register

Iximiel:
##### Description
 - added to ActionRegister.h a new template class ActionRegistration that works like the, PLUMED_REGISTER_ACTION without the need of the preprocessor.
 - changed PLUMED_REGISTER_ACTION in ActionRegister.h: now is a shortcut to a declaration: `#define PLUMED_REGISTER_ACTION(classname,directive) ActionRegistration<classname> directive##Registerer(#directive);` instead of a whole class definition+declaration 
 - changed PLUMED_REGISTER_ACTION in the sources with: `for a in */; do sed -r 's%(PLUMED_REGISTER_ACTION\(.*,.*)\"([A-Z_0-9]*)\"(\))%\1\2\3%g' "$a"/*cpp -i; done` the string constant is automatically stringified by the "#" operator in the preprocessor
   - it is a `sed` away to be completely exchanged by the explicit template tool
   - even if, I think that keeping the macro will keep the operation less verbose
In the case of multiple definitions, like XDistances and XYTorsion, the compiler will only declare one specialized ActionRegistration, but with multiple differently named instances without errors, for example
```C++
ActionRegistration<XDistances> xDistancesRegister("XDISTANCES");
ActionRegistration<XDistances> yDistancesRegister("YDISTANCES");
ActionRegistration<XDistances> zDistancesRegister("ZDISTANCES");
```

##### Target release

I would like my code to appear in the C++17 release (it uses a `inline constexpr` variable).

##### Type of contribution

##### Copyright

- [C] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

##### Tests


src/core/ActionRegister.h
```diff
@@ -29,7 +29,7 @@
 #include "tools/Keywords.h"
 #include <memory>
 
-namespace PLMD {
+namespace PLMD { //{
```
GiovanniBussi, line 32: @Iximiel why this additional open brace after the comment?

src/core/ActionRegister.h
```diff
@@ -100,14 +109,53 @@ std::ostream & operator<<(std::ostream &log,const ActionRegister&ar);
...
+  ::PLMD::PLMD::actionRegister()<classname>(directive);
+*/
+//in C++20 you we'll make this a concept a change typename in plumedRegisterer
+template<typename T>
+inline constexpr bool isActionType = std::is_base_of<::PLMD::Action, T>::value;
+//concept ActionType = std::is_base_of<::PLMD::Action, T>::value;
+//and the next template will be template<ActionType ActionType>
+
+template<typename ActionClass>
+class plumedRegisterer{
```
GiovanniBussi: We usually use capital letters for classes (PlumedRegisterer)

src/core/ActionRegister.h
Iximiel: I'm not a vim user, but is what I understood to do from [this](https://github.com/plumed/plumed2/pull/934#discussion_r1190056226)

src/core/ActionRegister.h
Iximiel: i corrected that in the fourth commit, is that not up?

GiovanniBussi: Maybe I was reviewing an older commit sorry for this

Iximiel:
Yes, I wanted to allow multiple names by using the wanted directive as a variable name and use the `#` to get the cstring for the directive

I suspected that removing the quotes was a too heavy change

If you prefer, I'll reverting that and substituting:
`#define PLUMED_REGISTER_ACTION(classname,directive) ActionRegistration<classname> directive##Registerer(#directive);`

with 
`#define PLUMED_REGISTER_ACTION(classname,directive) ActionRegistration<classname> classname##__LINE__##Registerer(directive);`
as I said is just a `sed` (or a `git rebase`) away
so there will be no virtual changes in the other files

GiovanniBussi:
Yes I think it's cleaner thanks!!!

src/core/ActionRegister.h
```diff
@@ -88,11 +88,34 @@ ActionRegister& actionRegister();
...
+    static_assert(isActionType<ActionClass>,"ActionRegistration accepts only class that inherit from Action");
+    actionRegister().add(directive.data(),create,ActionClass::registerKeywords);
+  }
+  ///On destruction deregister the ActionClass (useful when you unload a shared object)
+  ~ActionRegistration() {actionRegister().remove(create);}
+};
 
 #define PLUMED_CONCATENATE_DIRECT(s1, s2) s1##s2
 #define PLUMED_CONCATENATE(s1, s2) PLUMED_CONCATENATE_DIRECT(s1, s2)
-#define PLUMED_UNIQUENAME(str) PLUMED_CONCATENATE(str, __LINE__)
```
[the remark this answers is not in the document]
GiovanniBussi: I think it's correct. You are just using PLUMED_CONCATENATE below..

.gitignore
```diff
@@ -24,3 +24,5 @@ makefile.dep
 /config.*
 /autom4*
 /stamp-h
+/build
```
GiovanniBussi, line 27: maybe this can be a standalone commit?

Iximiel: removed from this PR

PR #944, opened 2023-05-23
OPES ECV stable algorithm instead of long double

invemichele:
##### Description

Using a numerically robust implementation instead of `long double` in the ExpansionCV class.
This makes all the calls to std::exp in the OPES module numerically stable and removes all occurrences of `long double` in the module.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@invemichele could you comment also on this in the changelog? I am sure you will summarize it better than me. Thanks!!!

invemichele:
These changes should have been part of the previous PR, I missed one file that had to be updated! 
That file used `long double` so my tests did not overflow, but still it could in principle overflow given large enough values. It is now safe.

So, I wouldn't add anything specific regarding this PR to che changelog.

PR #946, opened 2023-05-25
Json check

gtribello:
##### Description

This change adds a test on syntax.json after it is created to ensure that it is all written in valid json.

##### Target release

I think it is useful to include this as soon as possible as if there are problems in the syntax.json file it causes problems with the nest and the tutorials building.  I would put it in v2.9 as v2.9 is tested on the nest and on the tutorials pages.  

##### Type of contribution

##### Copyright

##### Tests


gtribello:
It is running on GitHub actions now.  Also the instruction to do the test is in the makefile for json so it runs every time you build the code including when it is built on Github actions.

PR #949, opened 2023-06-07
Proposal for mklib.sh modification

Iximiel:
##### Description

I'm changing the behaviour of mklib.sh in order to address the changes in #941, and make older codes compatible with the modifications in that PR. The script MUST be tested, and maybe reviewed, I'm still not quite convinced this is the best solution.

##### Target release

I would like my code to appear in the next release, it is related to some of the failing CI in the nest due to #941

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Can you check if the grep options you used are standard or if some new gnu version is needed?

In addition, you can add a regent. Look at the current regtest for the LOAD action, you can duplicate it or extend it

Iximiel:
> Can you check if the grep options you used are standard or if some new gnu version is needed?
> 
> In addition, you can add a regent. Look at the current regtest for the LOAD action, you can duplicate it or extend it

Now I'm using grep without -P, and the grep commands are changed to address only the legacy includes (before any typo would be autocorrected with a simple, easy to ignore warning).

I redirected the echoes of the warning to the stderr, so they can be tested with err.reference

Iximiel:
> I redirected the echoes of the warning to the stderr, so they can be tested with err.reference

Using the test features I injected a way to grep the warning line out of the stderr output

GiovanniBussi:
@Iximiel there's still an error in the mac build, please check the log.

Don't worry about the failure in conda-linux, it's unrelated

PR #951, opened 2023-06-14
Fixed nativeq (Q) Switching Function Analytical Derivatives

nm3787:
##### Description

This fixes the derivative for the nativeq (Q) switching function in tools/SwitchingFunction.cpp. The derivative of the Q switching function should be multiplied by beta according to the chain rule and not divided by r0. This would affect the forces acting on CVs using this switching function. This modification was discussed with @hockyg.

##### Type of contribution

##### Copyright

##### Tests


src/tools/SwitchingFunction.cpp
```diff
@@ -487,7 +487,7 @@ double SwitchingFunction::calculate(double distance,double&dfunc)const {
       dfunc=expression_deriv[t].evaluate();
     } else plumed_merror("Unknown switching function type");
 // this is for the chain rule:
-    dfunc*=invr0;
+    if(type!=nativeq) dfunc*=invr0;
```
carlocamilloni, line 490: Instead than this if I would divide for invr0 on line 454
if you can change this I will then merge it, thank!

carlocamilloni:
@nm3787 thanks for reporting and fixing, if you can change the code as suggested I will then merge it

hockyg:
> @nm3787 thanks for reporting and fixing, if you can change the code as suggested I will then merge it

Hi @carlocamilloni - I see why this makes sense from an efficiency point of view, but maybe you can add a code comment to warn people in case they add future switching functions?

nm3787:
> @nm3787 thanks for reporting and fixing, if you can change the code as suggested I will then merge it

Thank you @carlocamilloni for the great suggestion. I have implemented it and added a comment as @hockyg suggested.

PR #953, opened 2023-06-19
C++17

### LINKED ISSUE #930: Switch to C++17
In version 2.4 we switched to C++11. I guess in version 2.10 (=2.4 + 6) we might switch to C++17 (=C++11 + 6). I open this issue to collect suggestions on this topic.

**Portability**

Most C++17 is supported by gcc 8 (see [here](https://en.cppreference.com/w/cpp/17)). Historically, I was willing to preserve compatibility with Centos 7 / gcc 4.8 (used on machines at my institution). I recently learned they will switch to Rockylinux 8 / gcc 8, so I am happy to change.

Regarding tests:

- We could remove [centos7](https://github.com/plumed/plumed2/blob/master/docker/centos7)
- [Rockylinux 8](https://github.com/plumed/plumed2/blob/master/docker/rocky8) is already tested

I would instead continue to support older C++ versions in [our wrappers](https://github.com/plumed/plumed2/blob/master/src/wrapper/Plumed.h), mostly to make sure it is not necessary to use special compilation options when using PLUMED from an MD code.

**New features**

Incomplete set of possible improvements:

- We could remove 'dirent' dependence and switch to C++ [filesystem library](https://en.cppreference.com/w/cpp/header/filesystem).
- We could remove 'regex' dependence and switch to C++ regular expression (also in C++11 but not supported by gcc 4.8). Both these issues might help Windows portability (#855) in case we try again.
- [Aligned allocations](https://en.cppreference.com/w/cpp/memory/new/operator_new) might be helpful with OpenMP
- Maybe our formatting routines can be simplified using [from_chars](https://en.cppreference.com/w/cpp/utility/from_chars) and [to_chars](https://en.cppreference.com/w/cpp/utility/to_chars)
- Slightly simpler syntax for [exception safe locks](https://en.cppreference.com/w/cpp/thread/scoped_lock)
- any other idea?

GiovanniBussi:
##### Description

I think that in the future it will become more and more difficult to maintain the CI tests on centos7 as packages we depend on for testing are less and less supported (see this recent fix 762bca72b45e98c58e5d1a22c2ae1abb011dc72a). I would suggest we do the transition now and switch to C++17 in plumed 2.10.

Closes #930 .

NB: I have to take care of the conflicts now, should be easy.

##### Target release


GiovanniBussi:
And, there is a massive change now from `Tools::make_unique<Grid>` to `std::make_unique<Grid>`. I think it could be reverted by just reimplementing the former as a forward to the latter, to avoid massive failures on plumed nest.

GiovanniBussi:
Given the problems we have in #967 , I am more and more tempted to merge this and remove support for centos7.

If there are no objections, I will merge within tomorrow.

Issue #415, opened 2018-12-05
OpenMP regressions - more


carlocamilloni:
in multiple place in the code we make ad hoc estimates of the number of openMP threads like

  // Get number of threads for OpenMP
  unsigned nt=OpenMP::getNumThreads();
  if( nt*stride*10>nactive_tasks ) nt=nactive_tasks/stride/10;
  if( nt==0 || !threadSafe() ) nt=1;

This will likely have a bad effect on performances as noted recently.

carlocamilloni:
an easy fix is to modify it everywhere to
if( ntstride10>nactive_tasks ) nt=1
(already tested)

carlocamilloni:
@plumed/all I am fixing this from 2.3 on, the effect on performances can be huge when using many openmp threads (for example on KNL or with GPU). Since we fixed also the former openMP regression from 2.3 and since this is a big thing, I would suggest to release a one more version of plumed2.3 together with the next 2.4 and 2.5

carlocamilloni:
This can be close now

PR #955, opened 2023-06-26
small changes in the openMP settings

Iximiel:
##### Description

I checked out that the class `PLMD::OpenMP` was been used as a namespace: it only has static methods and no members
I think these small modifications may improve code readability: 
 - `PLMD::OpenMP` is now a namespace
 - the warning workaround in `PLMD::getGoodNumThreads` is now a commented input instead of a C-style cast, so there is no unused variable to throw a warning
 - in Tools/OpenMP.cpp `OpenMPVars` is now explicitly a singleton with a line of documentation and a private constructor

A question: reading `PLMD::getGoodNumThreads(T*,unsigned)`, it looks like it returns either `getNumThreads()` or `1`, is this the wanted behaviour? If yes I'm updating the documentation, there should be a unit test to check that the wanted  behaviour is present.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/regtest/.gitignore
+++ b/regtest/.gitignore
@@ -33,6 +33,7 @@
 !/s2cm
 !/pytorch
 !/membranefusion
+!/unittest
 # These files we just want to ignore completely
 tmp
 report.txt
--- a/regtest/unittest/Makefile
+++ b/regtest/unittest/Makefile
@@ -1,0 +1,2 @@
+include ../scripts/module.make
+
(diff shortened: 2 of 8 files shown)


GiovanniBussi:
@Iximiel good to me. Regarding your question:

> A question: reading PLMD::getGoodNumThreads(T*,unsigned), it looks like it returns either getNumThreads() or 1, is this the wanted behaviour? If yes I'm updating the documentation, there should be a unit test to check that the wanted behaviour is present.

There was some heavy slowdown reported at some point (#415) if a different number of threads is chosen for different loops (with the exception that it is always fine to choose one thread).

PR #962, opened 2023-07-07
Pw patch masses

Iximiel:
##### Description

I Think I solved the problem with quantum espresso's pw don't communicating the correct masses of the atoms. I need the CI to test the patch

##### Target release

I may need to rebase these commit on 2.8 or older version, since the code change is very small

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/patches/qespresso-6.2.diff/PW/src/plugin_ext_forces.f90
+++ b/patches/qespresso-6.2.diff/PW/src/plugin_ext_forces.f90
@@ -25,13 +25,14 @@ SUBROUTINE plugin_ext_forces()
   !
   IMPLICIT NONE
   !
-  INTEGER:: i,j
+  INTEGER:: i,j,ia
   REAL(DP) :: at_plumed(3,3)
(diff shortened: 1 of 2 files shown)


GiovanniBussi:
Great! Yes I think it can be backported. Can you test it [here](https://github.com/plumed/test-plumed-qe)?

I guess you should be able to do it by:
- making a fork
- in your fork, modify [this line](https://github.com/plumed/test-plumed-qe/blob/621be366fcbb4361aaf228c104a3a6fe71ea09e2/.github/workflows/ci.yml#L77) so as to install the modified plumed patch
- modify the test, which currently has only two Si atoms, or better add a new test

Iximiel:
The [PR](https://github.com/plumed/test-plumed-qe/pull/4#issue-1796406567) for the test is here,
I also discovered that I have updated the patch for qe 6.2 and not the one for qe 7.0.
Moreover I corrected a few Fortran errors that I wrote.

PR #967, opened 2023-08-02
Updated python version requested from the ci

Iximiel:
##### Description

Python 3.7 has finally come to its EOL,I'm starting a series of PR for trying to see if python 3.9 can be used without to many modifications

##### Copyright

##### Tests

### DIFF
--- a/docker/centos7
+++ b/docker/centos7
@@ -10,7 +10,7 @@ RUN yum -y update \
                    python3 python3-devel python3-pip \
  && pip3 install pillow==8.3.2 \
  && pip3 install numpy==1.19 \
- && pip3 install cython<3 numpy pandas mdtraj \
+ && pip3 install "cython<3" numpy pandas mdtraj \
  && pip3 install "gsd<3" \
  && pip3 install MDAnalysis==1.0.0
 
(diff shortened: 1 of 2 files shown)


GiovanniBussi:
Great!

Looks like with centos 7 this is [not trivial](https://discuss.python.org/t/is-there-any-alternative-way-to-install-python-3-9-in-centos-7-instead-of-using-source-code-tar-gz-tgz/22938/10).

However, I checked a [previously running job](https://github.com/plumed/plumed2/actions/runs/5566584934/job/15075142615) and it looks like the breaking change was the switch to cython 3.0.

Maybe it's sufficient to specify `cython<3` [here](https://github.com/plumed/plumed2/blob/0493ec1266ea10c365229c09313002f0042191c9/docker/centos7#L13)?

Iximiel:
> Maybe it's sufficient to specify `cython<3` [here](https://github.com/plumed/plumed2/blob/0493ec1266ea10c365229c09313002f0042191c9/docker/centos7#L13)?

If it works I think it can be safely merged into 2.9 and then we should carefully update the master, because of the change in the CI

docker/centos7
GiovanniBussi, line 1: I think you need quotes (as for `gsd` in the following line)

Iximiel: I saw and I added them

PR #972, opened 2023-08-29
Resolution function and PARAMETERS for SANS

hmcezar:
##### Description

This PR contains the implementation of a resolution function for the SAS spectra and enables using user-provided `PARAMETERS` for computing SANS.
I also implemented an option called `PARAMETERSFILE`, that can be used to provide a text file containing all the `PARAMETERS` instead of including them in the main input (when you have to specify the `PARAMETERS` for thousands of beads, this makes the input more readable).
To avoid increasing the computational cost too much when using the resolution function, splines are used to get the intensities and derivatives for `q` not provided by the user.
A description of the resolution function implementation and on how one can use the `PARAMETERS` in case of CG simulations can be seen in [this paper](https://pubs.acs.org/doi/full/10.1021/acs.jcim.3c00724).

A bit of context in the paper and implementation:
Last summer (before you pushed the SANS implementation in `SAXS.cpp` and reworked the code for the `ONEBEAD` implementation) I needed to perform metainference simulations biasing the SANS spectrum for a collaboration.
I implemented it in my [sans branch](https://github.com/hmcezar/plumed2/tree/sans) and have been using it ever since.
The difference between what is implemented in `SANS.cpp` in that branch and this PR is basically the normalization, which used to set `I(q_0) = 1`, with `q_0` being the lowest `q`, and now uses the same normalization you are using.

About the normalization, I performed a couple of tests, and it seems that in some cases, even when the spectra is flat for low q, the calculated curve is not set to 1 at `q=0`.
Instead, one has to use a reasonably large `SCALE` with metainference (around 25 for one of my systems) to get a good match between the experimental curve with `I(q=0)=1` and the computed curve.
This is true for both SAXS and SANS, so this is not something that comes from my implementation.
However, if you use a reasonable value for the `SCALE`, the curves match and everything works as it should.

Anyways, let me know if you need anything else regarding this PR.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/isdb/SAXS.cpp
+++ b/src/isdb/SAXS.cpp
@@ -252,7 +252,7 @@ private:
          BB_PO2,
          NONEBEAD
        };
-  struct SplineCoeffs{
+  struct SplineCoeffs {
     double a;
     double b;
     double c;
@@ -547,7 +547,7 @@ SAXS::SAXS(const ActionOptions&ao):
       Iq0 *= Iq0;
     }
     else { // SANS
-      std::vector<long double> parameter;    
+      std::vector<long double> parameter;
       parameter.resize(size);
       ntarget=0;
       for(unsigned i=0; i<size; ++i) {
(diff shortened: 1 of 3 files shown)


carlocamilloni:
merging! @hmcezar could you send me a working example with the Iq0 normalisation issue? something that can be run with plumed driver would be great

PR #980, opened 2023-10-16
Corrected a typo in mklib

Iximiel:
##### Description
Mklib now deletes correctly all the tmpfile:

I have changed back:
- `tmpfile=$(mktemp ${file%.cpp}.XXXXXX).cpp` that creates a tmpfile WITHOUT the .cpp extension, but saves `tmpfile` with the .cpp extension to be manipulated by the script
into:
- `tmpfile=$(mktemp ${file%.cpp}.XXXXXX.cpp)` creates a tmp .cpp file that will be correctly deleted after

This patch was though for the single file mklib, and we must note that do NOT patch an eventual header file with the old paths in it

##### Target release

I would like my code to appear in the next 2.9 patch /in 2.10

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/scripts/mklib.sh
+++ b/scripts/mklib.sh
@@ -44,7 +44,8 @@ do
   #adding a simple tmpfile, to preprocess "in place" the input file,
   #this assumes the user has write permission in the current directory
   #which should be true since we are going to compile and output something here
-  tmpfile=$(mktemp ${file%.cpp}.XXXXXX.cpp)
+  tmpfile=$(mktemp ${file%.cpp}.XXXXXX).cpp
+  mv "${tmpfile%.cpp}" "${tmpfile}"
   cp "${file}" "${tmpfile}"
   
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
On my computer, the file is named `pippo.XXXXXX.cpp` literally (XXXXXX is not replaced with a number. This is consistent with the documentation:
```
     The mktemp utility takes each of the given file name templates and overwrites a portion of it to create a file name.  This file name is unique and suitable
     for use by the application.  The template may be any file name with some number of ‘Xs’ appended to it, for example /tmp/temp.XXXX.  The trailing ‘Xs’ are
     replaced with the current process number and/or a unique letter combination.  The number of unique file names mktemp can return depends on the number of
```

GiovanniBussi:
More discussion [here](https://stackoverflow.com/questions/2419754/how-can-i-create-a-temp-file-with-a-specific-extension-in-bash)

Iximiel:
So as usual I fallen for the trap of gnu/linux that doing standard and are more permissive than the standard. I'm pushing the workaround in minutes

Issue #976, opened 2023-10-04
INTEL compiler not working on GH Actions


GiovanniBussi:
There seems to be an issue with the certificate from INTEL.

I will temporarily disable the test and leave this issue open as a reminder

Miniland1333:
Have you checked recently if this issue has fixed itself?

If this issue hasn't already resolved itself, one thing you may want to do is tweak the key process slightly so it always pulls the most recent key instead of the 2023-specific key. This would involve modifying the following lines
https://github.com/plumed/plumed2/blob/029d5211f4ac0cea22583dc203df44a5b3bd4da9/.github/workflows/ci.yml#L105-L107

to
```console
wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
rm GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
```

Lastly, since you are using the Intel OpenAPI would it be possible to please also test `icpx` as part of the intel test? The intel classic compilers (`icc` and `icpc`) are both deprecated and will be removed at the end of the year, so it would be good to check that `icpx` is working for plumed. Thanks!

Miniland1333:
For `icpx` specifically, there is a possibility you may need to add the CXXFLAGS `-qopenmp` or `-fiopenmp` instead of the traditional `-fopenmp` for it to link properly.

carlocamilloni:
I am testing this
```
wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
rm GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
```
and it works (branch test_intel), thanks!
I will now try to switch it to the `icpx` compiler.

carlocamilloni:
cf #990 where the new compilers work well to build plumed but we have multiple regtest failing not only for numerical issues.

Miniland1333:
Thank you all for addressing this so quickly!

PR #990, opened 2023-11-28
Test new latest intel compiler on GitHub Action

carlocamilloni:
This enables back the intel compilers testing and switches it to the new clang based version, cf. #976


carlocamilloni:
@valsson @invemichele multiple VES and OPES regtest fail using the new intel compilers, could you please check if there is anything relevant?

invemichele:
I just checked, for OPES it fails every time I rely on NaNs, it seems they are interpreted as zeros. In particular std::isnan seems to always return false. 
Maybe it's related to this? https://stackoverflow.com/questions/32195949/why-does-nan-nan-0-0-with-the-intel-c-compiler https://stackoverflow.com/questions/47703436/isnan-does-not-work-correctly-with-ofast-flags

GiovanniBussi:
Interesting. I am afraid @invemichele identified the point with NaN, which might affect other tests. Regarding the `rt-make-threads` I will try to check locally (it seems we have icpx 2021 installed, hopefully it can reproduce the bug).

@carlocamilloni was the problem present also with (old) icpc? Or maybe the new suites does not include icpc, and we are forced to switch to icpx?

carlocamilloni:
no we are not force to `icpx` yet but I think it is worth moving to it because `icpc` is deprecated. We can possibly overcome to `nan` issue by not using the default fast-math behaviour as mentioned in the post above, but at the same time I am not sure this is good because then people using the compiler will still get the error

Miniland1333:
Based on [this stackoverflow article](https://stackoverflow.com/questions/74401205/nans-are-getting-reported-as-0-while-compiling-with-icpx-and-not-as-nan), perhaps check if the issue also goes away if you use `-O2` instead of `O3`. If this is the case, the stackoverflow article recommends perhaps using `-fp-model=precise` with `-O3`.

invemichele:
On my side, using NaNs is a convenient way to do the parsing, but I could easily add a workaround for the case `isnan(nan)==false` so that the code is more robust

carlocamilloni:
I think that generally speaking is not worth trading performances to  parsing so it would be better to make the parsing more robust, so if a solution exist for the parsing I would say is better, @GiovanniBussi @maxbonomi @gtribello

GiovanniBussi:
I agree with latest @carlocamilloni 's comment.

Regarding the problem on `rt-make-threads`, it's misleading now. Due to an incorrect coding (that I am going to fix), the code crashes. The underlying error is just numeric. I guess it will be sufficient to change a little bit the trajectory in the test to make it work with optimizing icpx

valsson:
Hello, 

From what I can see, there are two failures in the VES regtests:
- ves/rt-bf-combined
- ves/rt-bf-legendre
and both seem to be due to numerics. Reducing the precision of the regtests should fix that. 

Changing `FORMAT_VALUES_DERIVS=%13.6f` to `FORMAT_VALUES_DERIVS=%13.5f` would reduce the precision of the relevant files. 

Will you do this change? Or should I do that via a PR on the master branch?

carlocamilloni:
@valsson thanks, I have just pushed the change into this pull request

valsson:
It seems that this commit has some unintended consequences on other files, so now it failing due to other issues. It is related to the `bf*.targetdist-averages.data.reference`. @carlocamilloni did you do that change on a Apple computer? I think it is related to issues with Apple clang compiler that I still needed to fix (see #950). I think that reverting all the `bf*.targetdist-averages.data.reference` files to the ones previous to this commit should fix the issue on GH action. 

If you revert this commit, I can do a PR on the master branch where I only commit the relevant files related to `FORMAT_VALUES_DERIVS=%13.5f`

carlocamilloni:
@valsson indeed, reverted, thanks

invemichele:
I make the relevant changes here #992
Notice that there are other parts of the code that rely on `std::isnan`, e.g. https://github.com/plumed/plumed2/blob/25abfd876b6260465593b7c277d83e9e74cfc722/src/vatom/Center.cpp#L221 probably it's worth checking for possible issues not caught by the regtests

carlocamilloni:
tests still failing: @invemichele @GiovanniBussi @valsson (this is to keep track of progresses)

+ ERROR in test basic/rt-ermsd/
+ ERROR in test basic/rt-lepton-asmjit/
+ ERROR in test basic/rt-lepton/
+ ERROR in test basic/rt-make-fortran/
+ ERROR in test basic/rt-make-threads/
+ ERROR in test basic/rt16/
+ ERROR in test crystallization/rt-averaged-q6-spAspB/
+ ERROR in test drr/rt-divergence/
+ ERROR in test drr/rt-restart/
+ ERROR in test opes/rt-ecv_linear-bis/
+ ERROR in test opes/rt-ecv_linear-tris/
+ ERROR in test opes/rt-ecv_multithermal_multibaric-bis/
+ ERROR in test opes/rt-ecv_multithermal_multibaric-custom-restart/
+ ERROR in test opes/rt-ecv_multithermal_multibaric-custom/
+ ERROR in test opes/rt-ecv_multithermal_multibaric-tris/
+ ERROR in test opes/rt-ecv_multithermal_multibaric/
+ ERROR in test opes/rt-opes_expanded-bis/
+ ERROR in test ves/rt-bf-combined/
+ ERROR in test ves/rt-bf-legendre/

invemichele:
for my code I switched to a much simpler solution that does not use NaNs and works the same in practice, see #993

carlocamilloni:
@gtribello could you check why this test `crystallization/rt-averaged-q6-spAspB` is broken using GitHub action with the intel compilers? is the only one left to be fixed but I do not really know this part of the code

PR #996, opened 2023-11-30
PyCV - calling python within plumed

Iximiel:
##### Description

The original work and ideas for PyCV are of @tonigi

This is the (draft) PR for PyCV.

PyCV gives the user the possibility to write a CV or a function in Python, with the possibility to use Python packages (like jax, sci-kit, etc..), while having access to some capability of plumed.

PyCV is a plug-in that can be LOADed in plumed.

Before setting this PR as ready to review some checks are needed:
- [ ] Are you ok with:
  - [ ] API: the names
  - [ ] API: the way the functions are called
  - [ ] API: what and how is returned
  - [ ] The two compilation methods
- [ ] Finalize the CI
- [ ] Correct(?) how I set up and updated the licenses

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
I have modified the CI to also run the PyCV tests, but only for the linux test:

I do not know if adding another identical compilation to the linux wf is a good idea, but was the fastest set up possible.

I tried to be the less invasive possible, but as you can see Rocky linux has been changed a bit:
yum insalls a python that cannot be linked with `pithon3-config --ldflags --embed` and so I modified the container recipe to install it.

Both for Fedora and Rocky I modified the the CI by adding a new Docker recipe that uses the previously compiled image to run pycv with the tests, this is more portable and the CI will be more clear if only PyCV fails

I see there are some conflict that are not possible to solve in the online editor, I will rebase this branch on master and then propose a non-draft PR.
Also, the merging https://github.com/plumed/plumed2/tree/autoconf-dladdr of makes the use of `LOAD GLOBAL` obsolete

PR #998, opened 2023-11-30
Drr fmt and regtest precision

carlocamilloni:
@HanatoK we need to decrease the precision of some drr regtest to avoid issues with GitHub Actions and intel compilers.
Here I have implemented some formatting options and decreased the precision in colvar_UI from 9 to 6. Is this fine with you? Do you see any possible issue?


HanatoK:
@carlocamilloni Thank you for making the changes! They looks good to me.

PR #1001, opened 2023-12-06
PyCV -- calling python within plumed

Iximiel:
##### Description

See #996 for a more complete description:
I have "rebased with cp" the pycv branch on the master branch

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@Iximiel I am trying to make this branch work on my laptop. Meanwhile, just looking at the scripts:
- you are using in many places `python3-config`. This should be customizable. Thinking about the possible use cases:
   -  By default, use the variable `python_bin` configured in `Makefile.conf`. It's not trivial to access to Makefile variables from a shell script. My suggestion is to use a Makefile that includes `Makefile.conf` and then, if you need a shell script, pass the value of the relevant Makefile variables as arguments to the shell script or as environment variables. You can also get some idea from what we do [here](https://github.com/plumed/plumed2/blob/5cbb885d7c389749059285cea82faf483ff89daf/python/Makefile#L15-L22).
   - To allow a user to just compile this plugin *after* plumed has been installed already somewhere, I guess you should allow the user to pass the name of the python interpreter likely as an environment variable. The variable could be `PYTHON_BIN`, consistently with what we use for `./configure`
- `--no-as-needed` is not available on MacOS. I guess it's not needed.

GiovanniBussi:
Some fixes [here](https://github.com/plumed/plumed2/tree/pycvMaster-gb).

A few things are missing. For the "simple" user, which would like to have everything installed automatically:
- find if `pybind11` is available in `./configure.ac`. If so, set a variable in `Makefile.conf`
- only if that variable is set, compile pycv during `make`. This is the same that we do for the python wrappers
- again, only if that variable is set, install pycv during `make install`. This is what we do [here](https://github.com/plumed/plumed2/blob/5cbb885d7c389749059285cea82faf483ff89daf/src/lib/Makefile#L314-L325) for python wrappers. We have to decide where, it should be a place that is easily accessible.

Next, we will have to find out how to add a new conda package to install only pycv.

Iximiel:
>     * you are using in many places `python3-config`. This should be customizable. Thinking about the possible use cases:
>       * By default, use the variable `python_bin` configured in `Makefile.conf`. It's not trivial to access to Makefile variables from a shell script. My suggestion is to use a Makefile that includes `Makefile.conf` and then, if you need a shell script, pass the value of the relevant Makefile variables as arguments to the shell script or as environment variables. You can also get some idea from what we do [here](https://github.com/plumed/plumed2/blob/5cbb885d7c389749059285cea82faf483ff89daf/python/Makefile#L15-L22).

It is possible to grep the python bin from plumed config, but the problem is that the 'official way' of getting the flags is calling `python3-config` and the hard thing is to get the right one, I found something about that.

>       * To allow a user to just compile this plugin _after_ plumed has been installed already somewhere, I guess you should allow the user to pass the name of the python interpreter likely as an environment variable. The variable could be `PYTHON_BIN`, consistently with what we use for `./configure`

In that case I would use the standaloneCompile.sh, and fish the pythonbin from `plumed config` or use the one that is in the active venv, I think that it is the less effort solution (for the user).

>     * `--no-as-needed` is not available on MacOS. I guess it's not needed.

On some system (for example on the default WSL, that should behave like ubuntu!!!) `gcc -shared` refuses to link libpython*.so to `PythonCVInterface.so`, I do not undestand why, since from manual `--no-as-needed` restores the "default behaviour", and it should be needed, since no `--as-needed` has been stated before, and there are calls for the Python lib symbols in ActionWithPython, and that should make the libpython to be linked even with `--as-needed`. (uses gcc 11.4.0).
I am confused by the CI passing that should be also using Ubuntu...

On mac you get an error? Or just a warning?

Iximiel:
@GiovanniBussi 

> find if pybind11 is available in ./configure.ac. If so, set a variable in Makefile.conf

Should I do something like [this](https://github.com/Iximiel/plumed2/pull/41/files)? (with autoconf2.69)

GiovanniBussi:
@Iximiel sorry for the delay. I was just thinking at checking if pybind11 is available in the same way as we [check if cython and setuptools are available](https://github.com/plumed/plumed2/blob/5cbb885d7c389749059285cea82faf483ff89daf/configure.ac#L850). In this way you can leave the logic of setting the linking flags in a single place.

Iximiel:
> @Iximiel sorry for the delay. I was just thinking at checking if pybind11 is available in the same way as we [check if cython and setuptools are available](https://github.com/plumed/plumed2/blob/5cbb885d7c389749059285cea82faf483ff89daf/configure.ac#L850). In this way you can leave the logic of setting the linking flags in a single place.

So like [this](https://github.com/Iximiel/plumed2/pull/41/files)? (with autoconf2.69)?
I added a few lines to the Makefile.conf, so that if at time of configure there is pybind11 in the python environment you get the ld and compile flags pre-baked there, if these are not present the sh will try to fetch them from the environment.

I think this solution should be refined, but I need some feedback to know in which direction:
Because as now, you can pycv (`canPyCV=@PLUMED_CAN_PYCV@` and related definition) even if there is no cython, even if with no cython there is no `PYTHON_BIN` set.

GiovanniBussi:
I think this is unnecessarily complicated and forces you to do the work twice. To me the simplest logic would be:

- at configure time:
  - check if `PYTHON_BIN` has been set, otherwise search for a python3 interpreter (without or without cython/pybind11, it does not matter).
  - then, check if cython is installed. If so, set a Makefile variable
  - then, check if pybind11 is installed. If so, set a Makefile variable
- at make time:
  - if cython was found, use it to build the wrappers
  - if pybind11 was found, use it to build pycv (HERE YOU FIND THE FLAGS)
- at make install time:
  - if cython was found, install the wrappers
  - if pybind11 was found, install pycv (**somewhere, I think we didn't decide yet where**)
 
Otherwise (this will be done in the conda builders):
- for python wrappers:
  - if the preprocessed plumed.c file is available use it (we provide it on pip so that the user does not need cython). otherwise use cython. Then use setuptools to build the wrappers
- for pycv:
  - search if plumed is available (you need to link against plumed here, which is not true for python wrappers)
  - if so, search **now** for pybind11 (it might not have been available at that time)
  - if available, use it to build pycv  (HERE YOU FIND THE FLAGS)

Does it make sense?

Iximiel:
I do not understand why I should not search the flags at configure time: if I do not set them there, why would I need to see if pybind11 is installed at configure time?

For the other things, I think is ok:
I'll have to to change the "plumed can python" statement that now is the `ifdef python_bin` in python/Makefile to something more flexible and remove the `PYTHON_BIN=` from the "failed to find cython" branch in the ac, if you are ok with this

GiovanniBussi:
>I do not understand why I should not search the flags at configure time: if I do not set them there, why would I need to see if pybind11 is installed at configure time?

Becausing searching the flags require some non trivial logic (e.g. to find the name of the proper `python-config` executable, as far as I understood). You do not want to implement this logic both in configure and in pycv build script. Since you need this logic in pycv build script (for people that want to build pycv on top of a preinstalled plumed), I would do it only there. Does it make sense? Or did I miss anything?

> I'll have to to change the "plumed can python" statement that now is the ifdef python_bin in python/Makefile to something more flexible and remove the PYTHON_BIN= from the "failed to find cython" branch in the ac, if you are ok with this

Yes. I guess there should be Makefile variables such as:
- `python_bin` (name)
- `plumed_python_has_cython` (`yes/no`), can only be `yes is `python_bin` is not empty.
- `plumed_python_has_pybind11` (`yes/no`), can only be `yes is `python_bin` is not empty.

or similar

GiovanniBussi:
> why would I need to see if pybind11 is installed at configure time
Regarding this, because you need to know if you should build the pycv module or not. 

I don't think this is the only possible way, but it is consistent with what we do in the python wrappers: we find in configure if cython is available or not.

Iximiel:
> Becausing searching the flags require some non trivial logic (e.g. to find the name of the proper `python-config` executable, as far as I understood). You do not want to implement this logic both in configure and in pycv build script. Since you need this logic in pycv build script (for people that want to build pycv on top of a preinstalled plumed), I would do it only there. Does it make sense? Or did I miss anything?

If you see the branch I linked, I think I found a simple workaround.
the cflags for pybind (it's header-only so no need for ld flags) can be found simply with the python executable (`-m pybind11 --includes`)

> Yes. I guess there should be Makefile variables such as:
> 
>     * `python_bin` (name)
> 
>     * `plumed_python_has_cython` (`yes/no`), can only be `yes is `python_bin` is not empty.
> 
>     * `plumed_python_has_pybind11` (`yes/no`), can only be `yes is `python_bin` is not empty.
> 
> 
> or similar

the checks for cython/pybind cannot be done without python_bin set, from the original ac, so I think we can pass from assuming that if python_bin is set, then cython have been found, to "if confugure has found cython/pybind then python_bin is set", and of course keeping python_bin into the Makefile.conf
I think that on this we are aligned

GiovanniBussi:
>If you see the branch I linked, I think I found a simple workaround.
the cflags for pybind (it's header-only so no need for ld flags) can be found simply with the python executable (-m pybind11 --includes)

OK, so it's fine to do it twice if you prefer. But do not add these flags to general INCLUDE flags, because we do not want pybind11 to be used in random places (for now).

Iximiel:
> > If you see the branch I linked, I think I found a simple workaround.
> > the cflags for pybind (it's header-only so no need for ld flags) can be found simply with the python executable (-m pybind11 --includes)
> 
> OK, so it's fine to do it twice if you prefer. But do not add these flags to general INCLUDE flags, because we do not want pybind11 to be used in random places (for now).
 as now is done like this:
```
canPyCV=@PLUMED_CAN_PYCV@
pybind11_cflags=@PYBIND11_CFLAGS@
python_cf_embedded=@PYTHON_CFLAGS@
python_ld_embedded=@PYTHON_LDFLAGS@
```

GiovanniBussi:
> it's header-only so no need for ld flags

If it can also be compiled with some `#define` that allows to bring it in a dedicated namespace (`PLMD::pybind11`) **and** licence allows it, we might even think of including the code as we do for `asmjit`. This namespacing stuff is necessary to avoid conflicts with MD codes using a different version of pybind11.

It might be worth if it's API is not very stable (and at some point we need to test for a specific pybind11 version). Otherwise, I think the current approach is perfect.

GiovanniBussi:
>I don't think it has many defines, but everything there is within their own pybind11 namespace and you do not have to fear about name conficts, because its (nearly) [mandatory](https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes) to compile with -fvisibility=hidden, exacltly for avoiding conflicts

Interesting

Iximiel:
I finally merged the autoconf modification, now if the ingredients (numpy and pybind11) are already present at plumed configuration, they will be automatically used for compiling PyCV, otherwise is up to the user to set up the correct variables, 

As now I did not rebased the branch onto the macOsS 12 commit

plugins/pycv/PythonFunction.cpp
```diff
@@ -0,0 +1,279 @@
...
+  }
+
+  log.printf("  with function : %s\n",calculateFunName.c_str());
+
+  log<<"  Bibliography "
+     <<plumed.cite(PYTHONCV_CITATION)
+     <<"\n";
+
+} catch (const py::error_already_set &e) {
+  plumed_merror(e.what());
```
GiovanniBussi, line 202: @Iximiel I would replace all these rethrow with something like:
```
plumed_error_nested();
```
or
```
plumed_error_nested() << "you can add more context here, for instance the label of the pycv, or anything that would help the user in understanding what happened";
```

plugins/pycv/PlumedPythonEmbeddedModule.cpp
```diff
@@ -0,0 +1,273 @@
...
+  .def("apply",[](const PLMD::Pbc* self, py::array_t<double>& deltas) -> py::array_t<double> {
+    //TODO:shape check
+    //TODO: this may be set up to modify the passed deltas, so no new intialization
+    //      ^this needs to be VERY clear to te user
+    auto accessor = deltas.unchecked<2>();
+    auto nat=deltas.shape(0);
+    py::array_t<double> toRet({nat,deltas.shape(1)});
+    auto retAccessor = toRet.mutable_unchecked<2>();
+    for (decltype(nat) i=0; i < nat; ++i) {
+      auto t= self->distance(PLMD::Vector(0.0,0.0,0.0),
```
GiovanniBussi, line 163: @Iximiel I think it might be useful to add this functionality to the Pbc class itself, I mean: the possibility to operate (probably in-place) on a `(PLMD::Vector*vectors,std::size_t size)` or on a `std::vector<PLMD::Vector>`. I guess  it would be more efficient (but it can be done later)

plugins/pycv/PythonFunction.cpp
Iximiel: `error(message)` is still ok for the ctors?

GiovanniBussi: Yes! However, the nested exceptions in theory allows one to retrieve more context. So I would use that one for new parts of the code (and I would suggest to change to this also in other parts of the code).

Iximiel: I added the nested error in the last commit

carlocamilloni:
@Iximiel @GiovanniBussi could we merge this and release a beta version of plumed 2.10? so that additional changes can be merged in new pull request?

GiovanniBussi:
If @Iximiel confirms I think we can merge. Conda build is failing for unrelated reasons

Iximiel:
I am ok with the merge, the thing on the pbc can be addressed in an ad hoc PR

PR #1002, opened 2023-12-13
Fix resolution function numerical issue

hmcezar:
##### Description

The way the resolution function was computed in the SAXS/SANS action could lead to some numerical issues due to computing the modified Bessel function with a large argument.
This could cause some `inf` and `NaN` depending on the sigmas and qs.

I have fixed this issue by implementing the [cephes](https://www.netlib.org/cephes/) function `i0e`, which computes `I_0(x) exp(-x)` and is less prone to overflow.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
@hmcezar sorry I merged and reverted it because you should target v2.9 if you want this to appear there

hmcezar:
@carlocamilloni I think it was my mistake. The current 2.9 does not have the "old" implementation of the resolution function and your most recent fixes. I think it's better if it goes to `master` to be consistent with what we already have. Should I create another PR?

carlocamilloni:
ok, I have reverted the revert, not the best but it is done

hmcezar:
Thank you! And sorry for the mess. My bad.

PR #1006, opened 2023-12-15
Link plumed tutes

gtribello:
##### Description

Replaced links for tutorials to new plumed tutorials page as discussed in Wednesday morning PLUMED meeting.

All the material for the old tutorials is still in the user-doc directory so folks can recreate their old tutorials on the new pages.  If they do they retain the ability to edit their tutorials and do not need us to check over their changes.  This is also an opportunity to refresh the tutorials and get rid of the old stuff that we think is no longer necessary.  I know that I have a number of tutorials that are now superseded but that I never got round to deleting.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@maxbonomi can you cross check as well? The idea is to transfer all tutorials to the new plumed-school page. I am asking you since you recently added one.

PR #1015, opened 2024-02-02
Optimize get words

### LINKED ISSUE #1011: Optimize getWords()
See [#933](https://github.com/plumed/plumed2/pull/933#issuecomment-1914776127)

GiovanniBussi:
I managed to optimize getWords.

@carlocamilloni can you also check the performance in this branch?

In short, there's a [new getWords](https://github.com/plumed/plumed2/blob/88e5e4a64ed84e9ce980b5b7755ac87c310f7313/src/tools/Tools.cpp#L211) written ad hoc for `plumed_cmd` strings, which is significantly faster for the following two reasons:
1. I removed the possibility to use complex syntax (parenthesis, tabs etc) in cmd string. I think this is fine, please let me know if anyone has concerns.
2. I used a `small_vector<std::string_view>` to avoid memory allocations. std::string_view is c++17, small_vector is an external library that I included [here](https://github.com/plumed/plumed2/commit/db959a39eb60d8bd653cfd29c3864e5938577a39), similarly to how we include lepton and other libraries. Notice that the small_vector library triggers a bug with gcc 8 that I fixed [here](https://github.com/plumed/plumed2/commit/7168aa1ca50e75eb5986e708018264d9c0af57db).

Before merging this branch I want to check with the author of the small_vector library if my fix is ok. To me, small_vector is a very useful tool that might allow optimizations in many places in PLUMED (basically, whenever we waste time allocating small temporary vectors). If I cannot sort this out, I can easily reimplement the optimization of getWords using a fixed size std::array (say with a maximum size of 4 words), that would work certainly in case of interpreting `plumed_cmd` strings.

This PR closes #1011


carlocamilloni:
@GiovanniBussi confirmed, with 11,500 ns/day vs 10,600 ns/day without this and 10,700 ns/day before htt.
Here the gain appears in the total time (~9 s vs 10.6 s) but not in the detailed timers.

carlocamilloni:
as a comparison without plumed we are 25,000 ns/day

GiovanniBussi:
>Here the gain appears in the total time (~9 s vs 10.6 s) but not in the detailed timers

This is expected. I think we are not including in the detailed timers what happens between when we call `plumed_cmd` and we start doing something. We always assumed this part of the code is not impacting performances.

I think this mostly relevant for very small (and irrelevant :-)) systems, but still it's a good exercise in hunting for bottlenecks.

carlocamilloni:
i agree I tested a 110,000 atom using saxs (the one for which we had 2.5% loss of performances by moving to htt) and the loss of performance is still exactly 2.5% irrespectively of all the optimisations done from that time. The main difference is in the "waiting data" (left is today, middle is before htt, with the htt at that time)
Screenshot 2024-02-02 at 18 27 41

Issue #1018, opened 2024-02-09
Possible bug in EDS and pesmd


GiovanniBussi:
While updating cppcheck I discovered two calls that might be bugs in pesmd and EDS, see [this commit](https://github.com/plumed/plumed2/commit/aaafacd174991c177436fe4827d01ba13ffdebcb)

@gtribello @whitead @hockyg 

Specifically, the second argument of `keys.addFlag` should be a `bool` but is actually a literal string, silently converted to `true`. I temporarily silenced the warning but I think it should be fixes. If you suggest the proper fix I would apply it back in PLUMED 2.8 and 2.9 as well.

BTW, @gtribello I am still confused how one would be willing to set the default value for a flag to `true`. Shouldn't this argument be removed and always set to `false`?

GiovanniBussi:
Closed [here](https://github.com/plumed/plumed2/pull/1021#issuecomment-1940647664) (fixed in v2.8 and later versions)

PR #1021, opened 2024-02-10
Setting flags to default true makes no sense so I added a plumed_merr…

gtribello:
##### Description

Addressing @GiovanniBussi issue #1018. 

I think you are correct @GiovanniBussi that it is not possible to set flags default true.  I thus added a plumed_merror in keywords if anyone does set the second argument to addFlag to true.  The better fix would be to get rid of the second argument but that is a massive change and I can't be bothered. 

##### Target release

I don't particularly mind

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/cltools/pesmd.cpp
+++ b/src/cltools/pesmd.cpp
@@ -111,7 +111,7 @@ public:
     keys.add("compulsory","idum","0","The random number seed");
     // temporary suppression, this is likely a bug that should be fixed!
     // cppcheck-suppress incorrectStringBooleanError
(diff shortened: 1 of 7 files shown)


GiovanniBussi:
OK it makes sense. If I understand properly, this does not affect the way the code behaves, but anyway it fixes the documentation (currently in the manual some flags are shown as "default on", incorrectly). I will try to backport this so as to have the manual fixed.

GiovanniBussi:
Backported in to v2.8 in 4736354aafd4ee7c0109a48a57224c3ac6132091, then merged to v2.9 and master

Issue #1025, opened 2024-02-14
Possible slow-down in htt (setting values in MPI code)


GiovanniBussi:
@gtribello this is another possible slow-down. For this I don't have timings, but I can produce them.

I am confused by the following code in DomainDecomposition:

      dd.Allgatherv(&dd.indexToBeSent[0],count,&dd.indexToBeReceived[0],&counts[0],&displ[0]);
      dd.Allgatherv(&dd.positionsToBeSent[0],ndata*count,&dd.positionsToBeReceived[0],&counts5[0],&displ5[0]);
      int tot=displ[n-1]+counts[n-1];
      for(int i=0; i<tot; i++) {
        int dpoint=0;
        for(unsigned j=0; j<values_to_get.size(); ++j) {
          values_to_get[j]->set( dd.indexToBeReceived[i], dd.positionsToBeReceived[ndata*i+dpoint] );
          dpoint++;
        }
      }

It looks like there’s a large loop (over all shared atoms) in which we call a quite expensive function (`void set(const std::size_t& n, const double& v ); `) which is not inlined and has quite a few if’s:

```
void Value::set(const std::size_t& n, const double& v ) {
  value_set=true;
  if( getRank()==0 ) { plumed_assert( n==0 ); data[n]=v; applyPeriodicity(n); }
  else if( !hasDeriv ) { plumed_dbg_massert( n<data.size(), "failing in " + getName() ); data[n]=v; applyPeriodicity(n); }
  else { data[n*(1+action->getNumberOfDerivatives())] = v; }
} 
```

I think this is suboptimal (and might affect timings with MPI, that we stopped doing at some point). Is there a way to avoid this sequence of calls?

I also do not understand where's the equivalent call for the serial version. For comparison, in the old code what we were doing was something like this:
```
        positions[dd.indexToBeReceived[i]][0]=dd.positionsToBeReceived[ndata*i+0];
        positions[dd.indexToBeReceived[i]][1]=dd.positionsToBeReceived[ndata*i+1];
        positions[dd.indexToBeReceived[i]][2]=dd.positionsToBeReceived[ndata*i+2];
```
which is just accessing the elements of an array.

GiovanniBussi:
@gtribello I found where this is done in the serial code. Basically, in the serial code we set the positions using a standard C++ loop in `(ip->mydata)->share_data`.

So, if I understand correctly:
- We first copy the local atoms quickly with those shared_data methods
- Then if we are using domain decomposition:
[...]

PR #1026, opened 2024-02-15
Removed call to set that is used in DomainDecomposition.

gtribello:
This addresses issues #1025 @GiovanniBussi

##### Description

This commit removes the call to set that is discussed in issue #1025.  I don't see that this making a noticeable difference to performance when compared with master but I am possibly not running on the correct setup.

timings

##### Target release

##### Type of contribution

- [x ] changes to code or doc authored by PLUMED developers, or additions of code in the core or within the default modules

##### Copyright

- [x ] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

##### Tests

### DIFF
--- a/src/core/DomainDecomposition.cpp
+++ b/src/core/DomainDecomposition.cpp
@@ -349,8 +349,7 @@ void DomainDecomposition::share(const std::vector<AtomNumber>& unique) {
       for(int i=0; i<tot; i++) {
         int dpoint=0;
         for(unsigned j=0; j<values_to_get.size(); ++j) {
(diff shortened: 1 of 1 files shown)


carlocamilloni:
@gtribello @gbussi on my big system (where I use 4 mpi) this is actually slower than master
this: PLUMED: 3 Waiting for data                              2001     1.161034 
master: PLUMED: 3 Waiting for data                              2001     0.713679

gtribello:
The fact that it is slower is very strange.

carlocamilloni:
yes you are right, with mpi in comparison to openMP (that is how I usually ran this test) the variability is much larger on the 2,000 steps that i use for this test. In this respect the branch does not make any significant difference

PR #1027, opened 2024-02-15
Removed second build of active atom list to address issue #1025

gtribello:
##### Description

@GiovanniBussi I think this change should give your speed up on the input you discussed in #1025.  This also ensures that the code still works when you run with DomainDecomposition, however.

##### Target release

##### Type of contribution

- [ x] changes to code or doc authored by PLUMED developers, or additions of code in the core or within the default modules

##### Copyright

##### Tests


GiovanniBussi:
@gtribello makes sense. It's however going to impact the MPI version, because in that case the function will be called.

I would merge this, then I'll be back when I will have done MPI tests :-)

PR #1028, opened 2024-02-22
An example of using cuda within plumed: the COORDINATION

Iximiel:
##### Description

I implemented a basic version of the Coordination with cuda.
It is not perfect, but showcase an approach to the problem using cuda.

I do not know if it is possible to run any GPU code on the nodes, as now there is no CI integration.

The plug in is set up like the pycv one.

I am adding a comment some graphs with the performance difference between the CPU and the cuda versions (run on Leonardo).

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
These images show how the time scales with nat= 108, 1944, 2916, 5832, 11664, 31104, 40500 atoms in the system.
The box is orthorombic and the atoms are arranged by tiling in xyz the space with the 6x6 box of 108 atoms that is used in the various coordination regexps.

The time axis is logarithmic.

### single group (self)
self
### between two groups
AB
### between two groups, but with the PAIR flag
pair

GiovanniBussi:
@Iximiel I see that this is fully in the plugin directory, so merging should nor hurt anything.

As for python stuff, I see some potential redundancy in the way you arranged regtests. Probably the best is to merge this as well and then check if we can factorize some script.

Some additional comments:
- I noticed that your regtests have a single snapshot in the trajectory. Could this be a weakness in testing? I have no idea how the code works, but would it be more robust to check that you can run through multiple steps without overwriting memory or so? Adding frames should be easy. I like that you always make the test against the CPU, which should result in diff=0.0. This means you can manipulate the trajectory as you like and you will always know what's the expected result. If you only test the difference, you might also generate the trajectory on the fly (with awk, there's some other examples in the regtests) and only check that diff=0.0
- The absence of CI is a bit scaring. I would suggest that you prepare a script somewhere so that you can easily submit tests on leonardo on a given commit. The code is quite decoupled from the rest though, so I don't expect problems.
- This calculation is without neighbor lists right? (all neighbors)

Then if you think it's ok I can just click merge

Iximiel:
> * I noticed that your regtests have a single snapshot in the trajectory. Could this be a weakness in testing? I have no idea how the code works, but would it be more robust to check that you can run through multiple steps without overwriting memory or so? A

Now that I think it, the test are more or less all weak because I'm just using the cube of 108 atoms repeatedly, so if I am not loading the atoms correctly I might use the same, not changed atoms.

> * The absence of CI is a bit scaring.

I agree. Is this a "me problem" or a "future me problem"? Maybe can be solved with the device emulation, but i need to dig further

> I would suggest that you prepare a script somewhere so that you can easily submit tests on leonardo on a given commit.

Because of the 2FA I think that will need a manual run, but the script should not be a problem

> * This calculation is without neighbor lists right? (all neighbors)
Yes.
Do you think this should be measured also against the CPU with the NL activated? It might be a fairer comparison.


Let me update a little the tests and then I think this can be accepted

GiovanniBussi:
> Maybe can be solved with the device emulation, but i need to dig further
This would be amazing

>plugins/cudaCoord/regtest/trajectories/traj-orthoBig.xyz

Maybe can we reduce a bit the footprint by generating this with awk before running the test?

Iximiel:
> > plugins/cudaCoord/regtest/trajectories/traj-orthoBig.xyz
> 
> Maybe can we reduce a bit the footprint by generating this with awk before running the test?

I collapsed both the pbc and the non-pbc trajectories in 10 frames

If you merge-squash we do not track the part with all the trajectories multiplied, and that would leave a lighter footprint

Iximiel:
I think is ready to be merged as is.

I have no idea how much time would take to set up the CI for the GPU code

PR #1033, opened 2024-03-02
More features from hack-the-tree

gtribello:
##### Description

These commits contain the new implementations of the actions belonging to the mapping module.  This is PATH, GPROPERTYMAP, PCAVARS, pathtools, ADAPTIVE_PATH and the GEOMETRIC_PATH.  These functionalities can be simplified because we can now pass vectors and matrices between actions.  

I think there is less controversial stuff here as I am mostly changing code that I authored.  The only action I have written that was not authored by me is RMSD.  This now calls the Tools::RMSD class directly (rather than using the reference module).  Furthermore, I have made the command RMSD a shortcut so colvar::RMSD is now called RMSD_SCALAR.  To be clear, however, you can still call RMSD with the old syntax:

````
rmsd: RMSD REFERENCE=myref.pdb TYPE=OPTIMAL
````

The difference is that if `myref.pdb` contains a list of reference structures you call a class called RMSD_VECTOR that was authored by me that calculates a vector of RMSD values from the reference structures.  Alternatively, if you do:

````
rmsd: RMSD REFERENCE=myref.pdb TYPE=OPTIMAL DISPLACEMENT 
````

Then RMSD_VECTOR is called again as you are calculating the vector of RMSD displacements for each atom.

##### Target release

I would like my code to appear in release 2.10 (ideally but I understand if that is too big an ask).

##### Type of contribution

##### Copyright

##### Tests

### CODE REVIEW

File: src/core/ActionWithVector.h [hunk 1]
```diff
@@ -173,7 +173,7 @@ class ActionWithVector:
 /// This is to transfer data from the buffer to the final value
   void finishComputations( const std::vector<double>& buf );
 /// Apply the forces on this data
-  void apply() override;
+  virtual void apply() override;
```
GiovanniBussi, line 176: I think "virtual" is not needed here: usually you have either virtual, or override, or nothing

src/setup/DRMSD.cpp
GiovanniBussi, line 1: Why is this in the setup subdirectory? Here we only keep Actions that inherits from ActionSetup. These actions are supposed to be before other actions because they affect how other action works. Such as UNITS, LOAD, or RESTART. I now see that there are many other actions in this directory now, I guess they ended up here by mistake, or is there something that I missed?

src/tools/RMSD.h
GiovanniBussi, line 1: If you use this during the simulations, it might be better to return a const reference, to avoid copying the vector.

GiovanniBussi:
@gtribello I added a few comments inline. I made a quick test and I noticed that RMSD is faster than before. I think that the reason is that you avoid using the reference module now, and directly use the tool (as it was in the earliest versions).

BTW I noticed that RMSD gives a segmentation fault using the benchmark trajectories after the 4th step. There might be a bug somewhere, but it looks like also previous versions of plumed have the same problem, so it's likely unrelated to this PR.

GiovanniBussi:
NB: the segmentation fault I was mentioning was a real bug, only showing up when all the atoms are arranged in a straight line (which happens in the benchmark trajectories). This is fixed in this commit now: f1da0a9b3a13f904bd97d6f92a2fb5e1b6479ac0

src/setup/DRMSD.cpp
gtribello: I moved DRMSD to the setup directory because it is a shortcut.  It works by creating a longer (more-complicated) input file that uses other action.  Consequently, functions in the DRMSD class are only called during setup.  I can move this back to Colvar if you prefer.

The other things that I put in setup are `Constant.cpp`, `Ones.cpp` and `PDB2Constant.cpp`.  These all create constants (`Ones.cpp` and `PDB2Constant.cpp` are actually shortcuts that create Constant actions).  In other words, these actions create a value during setup that then does not change during the simulation.  I think it makes sense for these to be in setup.

regtest/basic/rt-drmsd/config
```diff
@@ -1,3 +1,4 @@
+plumed_modules=adjmat
```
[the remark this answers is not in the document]
gtribello: I will try to explain this here but it would be good to have a chat about how to proceed here.  My preference would be to move some files around a little bit and then submit another PR.

So a plumed input file that could be created for doing DRMSD (remember DRMSD is a shortcut) is as follows (it is not done this way.  Although the way described below is probably better.  That is another change that should perhaps be made).

````
# Create a vector to hold the reference distances
dref: CONSTANT VALUES=1.1,1,3,1.4....
# Calculate all the distances
d: DISTANCES ATOMS1=1,2 ATOMS2=3,4 ...
# Calculate the displacement vector
dis: DISPLACEMENT ARG=d,dref
# Transpose the displacement vector
disT: TRANSPOSE ARG=dis
# Matrix multiply the transpose by the displacement to get the distance
d: MATRIX_PRODUCT_DIAGONAL ARG=disT,dis
````

This uses functionality for dealing with matrices.  All of this functionality is currently in adjmat as I wrote all the matrix stuff initially with adjacency matrices in mind.  I then started using it in other places because matrix multiplication is really useful!  So I should probably have a rethink of how the files are organised into modules.  That is quite a big job though that should maybe be done in a different PR.

File: src/setup/DRMSD.cpp
Comment (GiovanniBussi) [reply src/setup/DRMSD.cpp:1]: I see. I noticed now that with commit 79c832a97bf126b25b3f45e4a959505be9c0c4e8 (the big merge) you actually changed the meaning of ActionSetup.

Before: an action that should be at the beginning of the file because it changes how other actions are processed. Think about UNITS

[Now](https://github.com/plumed/plumed2/blob/026f7dfc2e9557e8f9775021848c8d8147aa6a4c/src/core/ActionSetup.cpp#L34): this is true only if you don't have a LABEL keyword

The point of having a separate base class was to enforce this with dynamic cast.

Maybe could you restore this? I guess you can simply use an empty `calculate` function if the purpose is to avoid calculations, right?

File: regtest/basic/rt-drmsd/config (hunk 2 above)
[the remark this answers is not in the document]
Comment (GiovanniBussi) [reply regtest/basic/rt-drmsd/config:1]: One issue that I see with the shortcuts is the following.

In a recent commit e912958d594be842d468a1480eec70478d22e0f1 I used a short script to remove unneeded dependencies. The script looks like this:
```
for file in */Makefile ; do

if grep -q "USE=" $file ; then
  deps=$(grep USE= $file | awk '{for(i=2;i<=NF;i++) print($i "/")}')
  for dep in $deps ; do
    git grep -q "^ *# *include \"$dep" ${file%Makefile}  || echo ERROR $file $dep
  done
fi

done
```

It checks if a module was USEd by mistake. And the fact that the build with all modules enables (on GitHub Actions) confirms that all needed modules where USEd.

However, we now have other dependencies that are a bit more difficult to track. For instance, the code might build with a set of modules, some action would seem available, but then would lead to an error at runtime because they generate a shortcut that need another module.

I don't have a solution however...

PR #1036, opened 2024-03-06
Sorted ActionSetup and added minimal atoms class to ensure that eggs with load keep working

gtribello:
##### Description

These are some changes in response to our discussions on 6th March.  

1. I have moved the actions for creating constants to generic and moved DRMSD to colvar.  Constant also no longer inherits from ActionSetup.  ActionSetup thus has the same role as it did in the earlier versions of the code (it is just for LOAD and UNITS).  Note that ActionForInterface actions are created **before** the plumed input is read in so some changes to ActionSetup are still necessary.

2. I have created a small class that contains the methods from Atoms that people use in the eggs use the PLUMED action LOAD as we discussed.  I have also added a regtest for this functionality. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
> Note that ActionForInterface actions are created before the plumed input is read in so some changes to ActionSetup are still necessary.

For this issue I was adding the ActionAnyorder base class, it is used by classes that can go before UNITS. But I now see that it is also a base for MOLINFO, which is not good because it looks like it uses units :-(
```
    if( ! pdb.read(reference,plumed.getAtoms().usingNaturalUnits(),0.1/plumed.getAtoms().getUnits().getLength()))plumed_merror("missing input file " + reference );
```

Maybe this is wrong?

GiovanniBussi:
In addition, I think you should replace `depreciated` with `deprecated` everywhere

GiovanniBussi:
>On the ActionForInterface thing, I think it makes sense that ActionSetup allows ActionForInterface actions before them as these are the PUT and DomainDecomposition actions that get data from the MD code. These actions have to be created before you start reading teh plumed input file, which begins with the setup actions.

Good point so maybe we should enforce that order?

gtribello:
I wouldn't enforce anything more than we have.  A person using PLUMED from python might want to pass some data to plumed, get it back from plumed and then pass some more data to PLUMED after some treatment in python.  I thus would allow people to create ActionWithInterface objects after they have created the ActionSetup objects as well.

Notice furthermore that a GET action is an ActionWithInterface object and that definitely comes after the SETUP.

GiovanniBussi:
OK makes sense.

I am still worried by UNITS and how it interact with other actions.

Maybe we should enforce some check? For instance, whenever we call the Units method we mark something as "called" and if then one tries to replace Units the code crashes. Because the way I was originally implementing this check (with ActionSetup) is not very robust

GiovanniBussi:
@gtribello meanwhile I merge this

PR #1037, opened 2024-03-07
Heuristic optimization of priority queue merging

GiovanniBussi:
##### Description

In v2.9 we added faster merge for requested atoms list (see #811). This was using two possible implementations for the merge:
- searching for the lowest element by comparing the lowest element of each vectors (default)
- using a priority queue (could be enabled with an env var for testing)

I made some quantitative test, and it looks like there is a regime where the priority queue becomes more efficient. The transition happens when we have > ~ 16 lists to be merged. In my test, this crossover happens independently of the size of each list.

I guess this might be due to some cache optimization: 16 ints would fit a cache line, so that comparing all of them is fast, whereas the corresponding priority queue would not fit the cache line. As long as we cross that barrier, the priority queue becomes more efficient.

In this commit I set the default to (a) full comparison when the number of lists is <=16 and (b) priority queue when the number of lists is >16. It is still possible to use an env var to manually choose the algorithm, but I guess this flexibility can be removed and the logic incorporated in the `mergeSortedVectors` function.

The figure compares timings for:
- yes (priority queue)
- no (full comparison)
- auto (priority queue more 16 groups)

It does so merging a variable number of CENTERs (x axis, `[10,20,30,50,70,100]`) of different size (label), and shows that the crossover is independent of the number of atoms per list. Full notebook attached.

image

[timings.zip](https://github.com/plumed/plumed2/files/14521381/timings.zip)

##### Target release

### DIFF
--- a/src/core/DomainDecomposition.cpp
+++ b/src/core/DomainDecomposition.cpp
@@ -37,13 +37,26 @@ Pass domain decomposed properties of atoms to PLUMED
 
 namespace PLMD {
 
+namespace {
+
+enum class Option {automatic, no, yes };
+
+Option interpretEnvString(const char* env,const char* str) {
+  if(!str) return Option::automatic;
+  if(!std::strcmp(str,"yes"))return Option::yes;
+  if(!std::strcmp(str,"no"))return Option::no;
+  if(!std::strcmp(str,"auto"))return Option::automatic;
+  plumed_error()<<"Cannot understand env var "<<env<<"\nPossible values: yes/no/auto\nActual value: "<<str;
+}
+
 /// Use a priority_queue to merge unique vectors.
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
I optimized the code a bit more. In particular:

- I am using small_vector optimization now, which makes the small examples slightly faster
- More importantly, I found a better way to implement the priority_queue

With the new implementation, the priority queue is always faster or comparable to the naive minimum search. This means that we can remove the option to switch from one implementation to the other and simplify a bit the code.

Below are results obtained with old priority queue (queue), new priority queue (heap), and naive search (maxelem1 and 2, two independent runs).

First results on INTEL MacBook with clang:

image

Then results with workstation using INTEL compiler:

image

PR #1038, opened 2024-03-07
Switch refactor for210

Iximiel:
##### Description

I refactored the switching function in order to make it easier to edit/add new functions by transforming the choices in a strategies.

By doing so I removed the `if else` chains in `SwitchingFunction::calculate()` and in similar methods.

I changed basic/rt20 to a series of tests for all the various switching functions.

These modification brings some performance changes that impact for example `COORDINATION` in the following way:

image

The time data is the mean value of 200 runs on at 100 frame trajectory of 108 atoms, and is taken from the plumed timing report. The plotted data is the value of the row `4 Calculating (forward loop)`).
The red data points are the time of the new implementation over the time of the original, and they are referred to the right axis
The vertical lines for the points are the standard deviation

The benches were run with the following commands:
|key| `COORDINATION` command|
|-|-|
| cosinus | `COORDINATION @GROUPS@ SWITCH={COSINUS R_0=2.6 }` |
| cubic | `COORDINATION @GROUPS@ SWITCH={CUBIC D_0=0 D_MAX=2.6  }` |
| exp | `COORDINATION @GROUPS@ SWITCH={EXP R_0=0.8 D_0=0.5 D_MAX=2.6  }` |
| fastrational-MMeq2NN | `COORDINATION @GROUPS@ SWITCH={RATIONAL R_0=1.3 NN=8 D_MAX=2.6 }` |
| fastrational | `COORDINATION @GROUPS@ SWITCH={RATIONAL R_0=1.3 NN=6 MM=10 D_MAX=2.6 }` |
| fastrationalTemplated-MMeq2NN | `COORDINATION @GROUPS@ SWITCH={RATIONAL R_0=1.3 NN=6 D_MAX=2.6 }` |
| gaussian | `COORDINATION @GROUPS@ SWITCH={GAUSSIAN R_0=1.0 D_0=0.3 D_MAX=2.6  }` |
| lepton | `COORDINATION @GROUPS@ SWITCH={MATHEVAL R_0=1.3 D_MAX=2.6 FUNC=1/(1+x^6) }` |
| q | `COORDINATION @GROUPS@ SWITCH={Q R_0=1.0 D_0=0.3 BETA=1.0 LAMBDA=1.0 REF=1.3 D_MAX=2.6  }` |
| rational-MMeq2NN | `COORDINATION @GROUPS@ SWITCH={RATIONAL R_0=1.3 NN=5 D_MAX=2.6 }` |
| rational | `COORDINATION @GROUPS@ SWITCH={RATIONAL R_0=1.3 NN=5 MM=11 D_MAX=2.6 }` |
| smap | `COORDINATION @GROUPS@ SWITCH={SMAP R_0=1.3 A=3 B=2 D_MAX=2.6  }` |
| tanh | `COORDINATION @GROUPS@ SWITCH={TANH R_0=1.3 D_MAX=2.6  }` |

For the `_ns` variants the `NOSTRETCH` keyword have been used.

Some clarification on the names for the rational function benchmarks:
 - _fastRational_ is when both `NN` and `MM` are even and so there is no need to use the square root of the distance
 - _MMeq2NN_ stand for `MM=2*NN` and gives room to an optimization of the rational
 - _fastrationalTemplated-MMeq2NN_ for some exponent combinations there is a compile-time optimized implementation (`NN=6,MM=12` here)

There are still some  corners that need to be smoothed:
 - updating the documentation to make the user aware of the pre-compiled rational
 - adding more options for the pre-compiled rationals, currently there is only `6/12` and `12/24`
 - maybe add a `SwitchingFunction::set(std::unique_ptr<baseSwitch>&&)` for declaring a new switch within any CV

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@Iximiel this is very useful I think 👏 

A few comments:
- I noticed you added many regtest, are they passing also using the old implementation? I mean: you were aiming at identical results right?
- I noticed some comments in the Q variable. I think its derivatives were fixed recently. I guess you used the latest (fixed) version.
- I really do not like the vector of Lepton functions (that I implemented indeed :-)). Do you think there could be a nicer solution? Otherwise we can leave it as is
- stretch and nostretch have identical performances, as expected. If you wanted to optimize this as well I think you should have used a template variable. However, I don't think it's worth: using the nostretch version is discouraged and I don't think we need to optimize it.

> updating the documentation to make the user aware of the pre-compiled rational

I think adding a sentence with "most commonly used rational functions are better optimized and might run faster" would be sufficient. People would not tune these coefficients for computational efficiency, but rather to have proper tail behaviors.

> adding more options for the pre-compiled rationals, currently there is only 6/12 and 12/24

I've seen using other small numbers, such as n=4. I would just do something like N=2,4,6,8,10,12 (with M=2N), it's just a few more copy and paste if I understood properly the code

>maybe add a SwitchingFunction::set(std::unique_ptr<baseSwitch>&&) for declaring a new switch within any CV

I am not sure I understand the syntax. Shall we perhaps have another register (as we have for CVs)? In the end I am not sure it's worth. We are talking about 1D functions, people can already play a lot with lepton, and then, if something becomes performance critical, one can directly modify the code. I think it's much less interesting wrt adding new CVs or biasing methods, for which we used registers to facilitate extensibility.

Iximiel:
> * I noticed you added many regtest, are they passing also using the old implementation? I mean: you were aiming at identical results right?

yes the regtest where `make reset`ed with the old implementation, I left the date in a comment the `config` of the tests

> * I noticed some comments in the Q variable. I think its derivatives were fixed recently. I guess you used the latest (fixed) version.

I ran the q test with the 2.10a to be sure about that


> > maybe add a SwitchingFunction::set(std::unique_ptr&&) for declaring a new switch within any CV
> 
> I am not sure I understand the syntax. Shall we perhaps have another register (as we have for CVs)? In the end I am not sure it's worth. We are talking about 1D functions, people can already play a lot with lepton, and then, if something becomes performance critical, one can directly modify the code. I think it's much less interesting wrt adding new CVs or biasing methods, for which we used registers to facilitate extensibility.

I don't want to propose a register for the switching function, the idea is that set would look like:
```c++
SwitchingFunction::set(std::unique_ptr<baseSwitch>&& ptr) {
function = ptr;
}
```
so that if one needs an ad hoc switch and does not want it depending on lepton in an eventual  LOADed cv.cpp can write something that inherits from `baseSwitch` and put in that cv


> * I really do not like the vector of Lepton functions (that I implemented indeed :-)). Do you think there could be a nicer solution? Otherwise we can leave it as is

I'll see what I can do about it

src/tools/SwitchingFunction.cpp
```diff
@@ -304,259 +646,232 @@ void SwitchingFunction::set(const std::string & definition,std::string& errormsg
...
+      //function = PLMD::Tools::make_unique<switchContainers::rationalSwitch>(d0,dmax,r0,nn,mm);
+    } else if(name=="SMAP") {
+      int a=0;
+      int b=0;
+      //in the original a and b are "default=0",
+      //but you divide by a and b during the initialization!
+      //better an error message than an UB, so no default
+      REQUIREDPARSE(data,"A",a,errormsg);
+      REQUIREDPARSE(data,"B",b,errormsg);
+      function = PLMD::Tools::make_unique<switchContainers::smapSwitch>(d0,dmax,r0,a,b);
```
Iximiel, line 772: @GiovanniBussi here I changed the behavior because I saw division by `a` and `b`
Do you think that this could broke something in some egg?

Iximiel:
> * I really do not like the vector of Lepton functions (that I implemented indeed :-)). Do you think there could be a nicer solution? Otherwise we can leave it as is

I refactored the thing by "compacting" the logic in a class
As now is private in the switch implementation, if it is useful it may be moved to the lepton interface

src/tools/SwitchingFunction.cpp
GiovanniBussi: I don't think so, I guess results would make no sense if one used either a or b = 0

regtest/basic/rt20-switch-fastrational-MMeq2NN/plumed.dat
GiovanniBussi, line 1: @Iximiel this test seems to fail on GitHub with INTEL compiler.

Notice that it is usually not relevant to use NUMERICAL_DERIVATIVES in regtests (except for explicitly testing the NUMERICAL_DERIVATIVES features), but results are numerically not very stable. You should use this when you develop some new functions to know if the derivatives you computed are correct. But once things are working correctly, you can just test the analytical ones.

For this specific PR what is relevant is that the analytical derivatives should be equal to those computer with a previous PLUMED version.

I would remove this from other tests of this group as well, since tests with NUMERICAL_DERIVATIVES often pop up later as numerical errors when people use different optimization options or compiler.

Iximiel: I just moves the original rt20 test,without modifying the numerical derive part in the .dat, the file do not output anwhere the action with the numerical derivatives, so it should not run, if I understood correctly
And the references test only the non numerical derivatives, as in the original

I think the problem may be another one, now that I looked at it:
the test that is failing on the derivatives on the 13th frame
```
2
10 10 10
X 0 0 0.0
X 0 0 1.3
```
for the Z coordinate of the two atoms and for the component of the virial v[2][2] in rational (tested with `NN=5 MM=11`) and fastrational (tested with `NN=6 MM=10`):

And that frame is in  `R==R_0`, so it may be related on how the step `if(rdist<=(lessThanOne) || rdist>=(moreThanOne))` is resolved in the `doRational`

Maybe I should drop all that "equals to" in floating points?

GiovanniBussi: Good point!

And good that you have a frame with R==R_0. I guess you can try if dropping the equals work better across different compilers

Iximiel: I changed `if(rdist<=(lessThanOne) || rdist>=(moreThanOne))` in `if(!((rdist < lessThanOne) && (rdist > moreThanOne)))` to mantain the single branch and the same closeness as in the original, hoping that Intel now will like it

Iximiel: I compiled in local with the avaiable oneapi:
It fails the same tests, so I made it plot a few variables:
using std::setprecision(15) to see the difference:
intel:
```
rdist 0.999999999999965
lessThanOne 0.999999999999978
moreThanOne 1.00000000000002
```
gnu:
```
```

Iximiel: I think the bug(?) is deeper that it looked:
this is the gnuplot of `awk '$2=="5"{print $1,$3}' deriv.reference > t`, the plot of the derivative of the z dimension
image
The kink is in the 13th frame, a little context: each frame has two atom on the z axis and the distance is 0.0 for the frame 0, 0.1 for the 1st, 0.2 for the 2nd and so on...

I checked with the current master and the results are the same.
I think rt20, the original coordination test that I used as a basis for the refactor, did not capture this because rt20 uses the default setting (so `NN=6, MM=12` -> "simplified + fast" rational) and it did not test the part of code that should avoid the 0/0 division.

GiovanniBussi: I am afraid you are right, there's a bug. I checked the derivative and it seems fine. If you do this change:
```
    if(rdist>(1.-1000.0*epsilon) && rdist<(1+1000.0*epsilon)) {
```
(1000 instead of 100) in the original code it seems to work, can you double check? I am afraid we should
- add a test for this
- make sure the factors are chosen properly, which could depend on NN and MM

Iximiel: I set up this demonstrator:
```c++
{
      for (auto [myN,myM]: {
             std::make_pair(6,12),
             std::make_pair(7,11),
             std::make_pair(8,13),
           }) {
        std::cerr << "N,M:" <<myN <<", "<<myM<<"\n";
        std::cerr<<"\nreference dfunc"<<std::setprecision(15)<<
                 (0.5*myN*(myN-myM)/static_cast<double>(myM))<<"\n";
        std::cerr<<std::setw(20) << "multiplier" <<"\t"
                 <<std::setw(20)<<std::setprecision(15)<< "1+multiplier*epsilon"
                 << '\t'<< "dfunc returned" << "\n";

        for (auto i : {-100.0,100.0})    {
          double res,dfunc=0.0;
          while (std::abs(i) <  (1e11)) {
            // for (auto& i:{-100000.0,-10000.0,-1000.0,-100.0,-10.0,-1.0,
            // 0.0,
            // 1.0,100.0,1000.0,1000.0,10000.0,100000.0}){
            auto x= 1.0+i*PLMD::epsilon;
            res= rational::doRational(x,dfunc,myN,myM);
            std::cerr<<std::setw(20) << i <<"\t"
                     <<std::setw(20)<<std::setprecision(15)<< x
                     << '\t'<< dfunc << "\n";
            i*=10.0;

          }
        }
      }
    }
``` 
that plot this:
<details>
  <summary>Long table hidden here, if markdown compiler works well</summary>

calculations with intel, gnu do not differ much:

```
N,M:6, 12

reference dfunc-1.5
          multiplier    1+multiplier*epsilon    dfunc returned
                -100       0.999999999999978    -3
               -1000       0.999999999999778    -3
              -10000        0.99999999999778    -3
             -100000       0.999999999977796    -3
            -1000000       0.999999999777955    -3
           -10000000       0.999999997779554    -1.12350014269583
          -100000000        0.99999997779554    -1.498799992354
         -1000000000       0.999999777955395    -1.4999246923247
        -10000000000       0.999997779553951    -1.50000371172241
                 100        1.00000000000002    -3.00333333333333
                1000        1.00000000000022    -3
               10000        1.00000000000222    -3
              100000         1.0000000000222    -3
             1000000        1.00000000022204    -3.00000033333333
            10000000        1.00000000222045    -1.12350015730417
           100000000        1.00000002220446    -1.51756524996087
          1000000000         1.0000002220446    -1.49992735633871
         10000000000        1.00000222044605    -1.49999751824281
N,M:7, 11

reference dfunc-1.27272727272727
          multiplier    1+multiplier*epsilon    dfunc returned
                -100       0.999999999999978    -2.54545454545455
               -1000       0.999999999999778    -2.54545454545455
              -10000        0.99999999999778    -2.54545454545455
             -100000       0.999999999977796    -2.54545454545455
            -1000000       0.999999999777955    -2.54545418181818
           -10000000       0.999999997779554    0.059933709635761
          -100000000        0.99999997779554    -1.29672905900383
         -1000000000       0.999999777955395    -1.27300013902193
        -10000000000       0.999997779553951    -1.27272799925931
                 100        1.00000000000002    -2.54545454545455
                1000        1.00000000000022    -2.54509090909091
               10000        1.00000000000222    -2.54545454545455
              100000         1.0000000000222    -2.54545454545455
             1000000        1.00000000022204    -2.54545454545455
            10000000        1.00000000222045    -2.54545450909091
           100000000        1.00000002220446    -1.25020454497731
          1000000000         1.0000002220446    -1.27246327715259
         10000000000        1.00000222044605    -1.27272536703273
N,M:8, 13

reference dfunc-1.53846153846154
          multiplier    1+multiplier*epsilon    dfunc returned
                -100       0.999999999999978    -3.07384615384615
               -1000       0.999999999999778    -3.07692307692308
              -10000        0.99999999999778    -3.07692307692308
             -100000       0.999999999977796    -3.07692307692308
            -1000000       0.999999999777955    -3.07692307692308
           -10000000       0.999999997779554    -1.61125454166832
          -100000000        0.99999997779554    -1.54063578537643
         -1000000000       0.999999777955395    -1.53858199953326
        -10000000000       0.999997779553951    -1.53846068954135
                 100        1.00000000000002    -3.07692307692308
                1000        1.00000000000022    -3.07692307692308
               10000        1.00000000000222    -3.07692307692308
              100000         1.0000000000222    -3.07692307692308
             1000000        1.00000000022204    -3.07692307692308
            10000000        1.00000000222045    -0.945041592730449
           100000000        1.00000002220446    -1.53663891550882
          1000000000         1.0000002220446    -1.53847950423642
         10000000000        1.00000222044605    -1.53846011065097
```

</details>

I found that 1000 is enough to pass this test, but by the demonstrator goes deeper:
I think that we need **1000000000** epsilons (that is 10^9 O_O, that is more or less 2.22e-07 on my machine) for returning a dfunc that is slightly higher, and correct in value (see the negative multipliers in 7,11 for when I am talking about correctness)

Iximiel: Ok, some numbers around this.
I added a cltool to visualize what is happening the more we go near the division 0/0 in r=R_0 in the non-simplified rational

Here are some images that show how the floating point numbers behave when you sum and multiply things with extremely different exponents:
<details>
  <summary>Example with 1e4ε gap</summary>

The derivative:
dfunc1e4
The "gradient" of the derivative
grad1e4
</details>

<details>
  <summary>Example with 1e7ε gap</summary>

The derivative:
dfunc1e7

The "gradient" of the derivative
grad1e7

</details>

<details>
  <summary>Example with 1e10ε gap</summary>

The derivative:
dfunc1e10
The "gradient" of the derivative
grad1e10
</details>

<details>
  <summary>Example with 5e10ε gap</summary>

The derivative:
dfunc5e10
The "gradient" of the derivative
grad5e10

</details>

The thing to appreciate is the difference in scale between the x and the y axis

the gradient is calculated with [numpy](https://numpy.org/doc/1.25/reference/generated/numpy.gradient.html), and I used to show that there are still some instabilities with the bigger gaps, all the data is shifted to 0, to be more clear

Since the gap around R_0 was becoming significant,as we discussed offline I changed the calculation in the gap to use a 2nd order Taylor polynomial.

The number that I pushed for the gap is 5e9,If you like it I'll update the test .reference, and update the changelog, and then I think we can call it


By doing some math and looking at some function graphs, the way N and M impact on this is that the more they are similar the more R_0 is near a local minimum of the derivative of the rational, being a "local minimum" minimum at N=M ("local minimum" in a constant function). In a few words N and M impact how the derivative in the gap is "symmetric" around R_0

Iximiel:
@GiovanniBussi 
> @Iximiel if you are able to solve the problem with the numerical instabilities of the derivatives and it turns out to be a simple recipe such as:
> 
>    * change the boundaries
> 
>    * use a Taylor expansion in the middle
>       Maybe we can backport it to v2.8?

I think it is possible, I touched not too many files and this could "just" be rebased on that.
I just have to rework the C++17 shortcuts, like  all the `if constexpr` that I used, and then I just have to prepare a small PR to revert the code to c++17 after the merge into master
In [the discussion of the error i the CI](https://github.com/plumed/plumed2/pull/1038#discussion_r1524706730) I posted some data using Taylor in the middle of the gap

> Another independent thought. Perhaps something even better than a Taylor expansion (for our purposes) could be:
> 
>    * choose the boundaries
> 
>   * replace the function in the middle with a linear interpolation between the boundaries.
> 
> 
> If the function can be computed in a numerically stable way at the boundaries, then it's sufficient to call it twice when the SwitchingFunction is created to find the expression of the straight line fitting those two points, and save the two coefficients (A*x+B). The value at r=R_0 will not be anymore exactly NN/MM, but this is not a problem. The derivative will be A over the entire range.
>
> Even better: one could compute both the values and the derivatives at the boundary, and interpolate with a cubic spline. This will make also the first derivative continuous. If you have 4 conditions (two values and two derivatives) you can fit a polynomial function such as A_x^3+B_x^3+C*x+D, for which the derivative is analytical.
 
It should work, 
We just need to measure where the numerical instabilities stops, to put reasonable boundaries there

GiovanniBussi:
Great! For backporting I just meant the derivative calculation, because it is a bug, not the full implementation... thanks!

GiovanniBussi:
@Iximiel is this ready for merging?

Iximiel:
@GiovanniBussi I updated the test to be correct to with the change of the rational, if it passes also the intel CI I think it is ready to push

PR #1040, opened 2024-03-08
Track shortcut action usage

gtribello:
##### Description

Added functionality to track how actions are used in shortcuts to create more complex inputs.  

There are two ways that I use shortcuts.  The first is the straightforward way.  For these I want the use of shortcuts to be visible to users.

The other is when you call things like CUSTOM, which is a shortcutAction that calls CUSTOM_SCALAR, CUSTOM_VECTOR, CUSTOM_MATRIX or CUSTOM grid.  Here I want the user to use CUSTOM comfortable in the knowledge that PLUMED knows what to do with the input action.  They don't need to know that the implementation has these are separate actions.  They are only separate actions because I think that makes the code more readable.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@gtribello great!

If I understand correctly from the change in the regtest, before this fix there were cases where the user was forces to use `ARG1` even with a single argument, whereas now it is possible to just use `ARG`. I agree this is better, especially if you want to later transform some other functionality in a shortcut, without the users need to update their input files.

For me it can be merged

GiovanniBussi:
Except that there are conflicts with another PR from @gtribello that I just merged (I guess easy to fix) and that some regtests are failing (not clear to me why)

gtribello:
OK @GiovanniBussi 

I hope that should be all of those problem fixed now

PR #1042, opened 2024-03-12
Extensive improvements in benchmark tool

GiovanniBussi:
Added:

- stop after a maximum time
- benchmark domain decomposition runs
- comparative analysis on different versions
- sleep to simulate MD code calculation

See [documentation](https://github.com/plumed/plumed2/blob/27e789cfe1e3bfd895efd89c695c6468d4bb75b8/src/cltools/Benchmark.cpp#L45-L156)

@carlocamilloni could you share the SAXS test case you were running on MPI? I want to test it with this benchmark tool, to see if by running multiple versions at a time we get more precise results.

I would need:
- plumed input file
- all accessory files (pdb etc) if needed

no trajectory should be needed

### DIFF
--- a/src/cltools/Benchmark.cpp
+++ b/src/cltools/Benchmark.cpp
@@ -193,8 +193,8 @@ extern "C" void signalHandler(int signal) {
   }
 }
 
-/// Local structure handling a kernel and the related timers
-struct Kernel {
+/// This base class contains members that are movable with default operations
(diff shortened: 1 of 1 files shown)


carlocamilloni:
@GiovanniBussi here it is:

[Archive.zip](https://github.com/plumed/plumed2/files/14588508/Archive.zip)

PR #1048, opened 2024-03-15
Patched the rational non simplified switching function around X==R_0

Iximiel:
##### Description
First of all, sorry for the Friday afternoon PR.

As we discussed in #1038, [here](https://github.com/plumed/plumed2/pull/1038#issuecomment-1999677314)
I patched the numerical instabilities around R0 in the non simplified rational switching function.
And I used the test from #1038 for ensuring the correctness of the change (v2.8 won't pass the new test).

Before merging: should I update the changelog?

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Great! I see that other tests are not affected, I am afraid there was no test on the 2 N!= M case, or maybe no test was probing that point? 

I would add something in the change log for version 2.8

Thanks!

Iximiel:
> I am afraid there was no test on the 2 N!= M case, or maybe no test was probing that point?

I think that was the reason it slipped trough the tests. We have to thank intel that approximates differently than gcc 

I update the changelog, I think this can be squashed+merged

PR #1051, opened 2024-03-27
deactivate restart during parse only runs

gtribello:
##### Description

As discussed in the meeting today, if we make this change many eggs in the nest will pass as the problem with them is that people have upload a PLUMED input file that has RESTART in it but no hills file.  If we thus suppress RESTART when we run parse-only mode the tests should pass.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/setup/Restart.cpp
+++ b/src/setup/Restart.cpp
@@ -114,7 +114,7 @@ Restart::Restart(const ActionOptions&ao):
     if(md) log<<"  Switching off restart\n";
     plumed.setRestart(false);
     log<<"  Not restarting simulation: files will be backed up\n";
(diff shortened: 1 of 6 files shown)


src/setup/Restart.cpp
GiovanniBussi, line 1: @gtribello instead of modifying Restart.cpp, isn't it less intrusive to change PlumedMain::setRestart so that, if it's in parse only mode, it doesn't make anything?

gtribello: OK @GiovanniBussi I made that change

carlocamilloni (approved):

carlocamilloni:
It had a good impact, 12 eggs and around 50 inputs are now correct

PR #1054, opened 2024-03-29
New version of dimensionality reduction from hack-the-tree

gtribello:
##### Description

I also want to modify the code that I wrote for doing dimensionality reduction with PLUMED.  This new version of this code does all these things using shortcut actions.  You can thus get rid of a huge amount of code that was offering another way of passing data between actions.  I think no one is really using the dimensionality reduction code from PLUMED so I don't imagine this change will be very controversial.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hi @GiovanniBussi 

I added some information to the CHANGELOG to explain a bit about the changes in this PR.  I thought I would also provide some context here.

So the aim of the this and my two previous PRs is to get rid of dependencies on the code in the old modules vesselbase and reference.  The code in these modules grew rather organically and was becoming increasingly difficult to maintain.  Three features (I implemented) relied on the code that is in this module:

1. multicolvar - that was reimplemented without using the modules in the first of these PRs
2. paths - that was reimplemented without using the modules in the second of these PRs
3. dimensionality reduction - this is reimplemented here

The biggest change (and most contentious of these changes) was (1) the rewriting of multicolvar.  Rewriting dimensionality reduction here is a relatively small change to the code by comparison.  This PR looks huge, however, because, now that there is nothing that depends on reference and vesselbase, I can delete the code in those two modules.  Those two modules have thus been deleted here.

I hope that helps give some better context for the PR.

GiovanniBussi:
Thanks @gtribello . I checked the regtest, and indeed it mostly affect dimensionality reduction stuff.

Can you confirm that no single egg on plumed-nest will be invalidated by this? I was mostly worried about PCA, on which I've seen discussions on the mailing list.

Also, can you confirm that registers for vessels and for metrics will be gone then? I am asking because I am putting my hands into registers (to fix the behavior of LOAD).

Then I think this can be merged

gtribello:
Hi @GiovanniBussi 

I can confirm that the registers for vessels and metrics are both gone.

In terms of the nest.  I went through and searched for the action names that I have modified here.  I found the following eggs:

https://www.plumed-nest.org/eggs/23/019/

This one is currently showing as broken as it is using the syntax that was in the hack-the-tree branch, which was slightly different. I will get in touch with Matteo and see if we can update it.

https://www.plumed-nest.org/eggs/19/078/

This one is also using features from hack-the-tree.  Not sure what to do about that one as I am not sure implementing the ITRE thing in PLUMED that it is using is a good idea.  It is already showing as broken though

PR #1059, opened 2024-04-15
Benchmark distributions

Iximiel:
##### Description

I did some modification to the Benchmark.
After removing a few warnings and adding some `const`, I added a few options for the "simulated atoms" and changed how existing one, "the line", works.

The options that I added are:
- *an uniform distributed cube*
- *an uniform distributed sphere*
- *two uniform distributed sphere that should not touch each other*
- *a static "simple cubic"* distribution

the distributed ones are randomly generated at each step, so I think that are not suited to test a neigbour list
the simple cubic do not change with advancing steps, but I may add to it a way to make the atoms "oscillate" around their positions

The original line was adjusted from 
```c++
for(unsigned j=0; j<natoms; ++j)
  pos[j] = Vector(step*j, step*j+1, step*j+2);
```
to
```c++
for(unsigned j=0; j<natoms; ++j)
  pos[j] = Vector(j,0,0)+oscillation(step);
```
because the first version produces on step 0 all atoms in the same `(0,1,2)` position, and then a series of atoms each one at distance `(s,s,s)` from the previous one, where `s` is the current step number, and this brings some actions, like COORDINATION, to skip the costly part of the calculation.
I think keeping "the line" is important to check that that component in some actions works.

These modifications also add a "contract" with the user: the uniformly distributed configuration are made such that in average each atom has a volume of 1: with `nat` being the number of atoms, the volume of the cube is `nat`, the volume of the sphere is `nat` and the sum of the volumes of the two spheres is again `nat`.
Similarly the atoms on the line oscillate on a linear grid with step `1` and the `a` for the grid of the simple cubic is `1`.

All the distribution, except from the line now return a box.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

---

unrelated: `Random::RandInt` is declared, but not implemented

### DIFF
--- a/src/cltools/Benchmark.cpp
+++ b/src/cltools/Benchmark.cpp
@@ -261,7 +261,7 @@ struct Kernel :
   }
 };
 
-namespace benchDistributions {
+namespace  {
 
 class UniformSphericalVector {
   //double rminCub;
@@ -409,7 +409,7 @@ struct tiledSimpleCubic:public AtomDistribution {
 
   }
 };
-std::unique_ptr<AtomDistribution> getDistribution(std::string_view atomicDistr) {
+std::unique_ptr<AtomDistribution> getAtomDistribution(std::string_view atomicDistr) {
   std::unique_ptr<AtomDistribution> distribution;
   if(atomicDistr == "line") {
     distribution = std::make_unique<theLine>();
@@ -427,7 +427,7 @@ std::unique_ptr<AtomDistribution> getDistribution(std::string_view atomicDistr) 
   }
   return distribution;
 }
-} //namespace  benchDistributions
+} //anonymus namespace for benchmark distributions
 class Benchmark:
   public CLTool
 {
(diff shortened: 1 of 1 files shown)


src/cltools/Benchmark.cpp
```diff
@@ -255,6 +261,173 @@ struct Kernel :
   }
 };
 
+namespace benchDistributions {
```
GiovanniBussi, line 264: If this is only used here, I think it's better to use an anonymous namespace to avoid the risk of clashes

GiovanniBussi:
For me this can be merged

src/cltools/Benchmark.cpp
GiovanniBussi: Why? I think you just have to write
```
std::unique_ptr<AtomDistribution> distribution
```

I think we should avoid to add custom namespace because at some point there will be a name clash. If it's only in a cpp file, anonymous is usually the right thing to do.

Iximiel: Made it anonymus, I also changed the name of the factory function, so that is more clear

PR #1063, opened 2024-04-24
Added a simple module map

Iximiel:
##### Description

@gtribello, @GiovanniBussi I have set up a simple module map
It can be used as `PLMD::getModuleMap().at("actionName")`[^1] to return the name of the module that includes that action (even if the module is not active in the current installation)

for 2.10 it is possible to add also an extra function that returns a `std::optional` with the module name to not doing the `.count("actionname")` before calling the at (that may throw), since `std::optional` is c++17

Since regex are scary I put a link to a site tha visualize the one used in the new maketool script
The new maketool script uses pure bash, I think there should not be problem between bash versions for the regex that I used, and I used pure bash to avoid having the user to install gawk or perl
I do not know if you like the requirements for the `.inc` file
[^1]:`operator[]` has no `const` overloads for `std::map` and `std::unordered_map`
##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/maketools/makeModuleMap
+++ b/src/maketools/makeModuleMap
@@ -1,0 +1,59 @@
+#! /usr/bin/env bash
+# shellcheck disable=SC2059
+
+problems=""
+toPrint='  {"%s", "%s"}'
+#regex are scary, here's a link with the flow of the regex
+# https://regexper.com/#PLUMED_REGISTER_ACTION%5Cs*%5C%28%5Cs*%5B0-9a-zA-Z_%5D%2B%5Cs*%2C%5Cs*%22%28%5B0-9a-zA-Z_%5D%2B%29%22%5Cs*%5C%29
+regex='PLUMED_REGISTER_ACTION\s*\(\s*[0-9a-zA-Z_]+\s*,\s*"([0-9a-zA-Z_]+)"\s*\)'
+
+for module in ../*/module.type; do
+  moduledir=${module%%/module.type}
+  modulename=${moduledir##*/}
+  {
+    for file in "$moduledir"/*.cpp; do
+      if grep -q PLUMED_REGISTER_ACTION "${file}"; then
+        # gawk and perl implementation in case your bash dialect do not process correctly the regex
+        # actionName=$(gawk 'match($0,/PLUMED_REGISTER_ACTION\s*\(\s*[0-9_a-zA-Z]+\s*,\s*"([0-9a-zA-Z_]+)"\s*\)/, captured){print captured[1]}' "${file}" )
+        # # in case gawk does not work, here's the perl implementation:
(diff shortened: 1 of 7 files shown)


src/maketools/makeModuleMap
```diff
@@ -0,0 +1,59 @@
+#! /usr/bin/env bash
+# shellcheck disable=SC2059
+
+problems=""
+toPrint='  {"%s", "%s"}'
+#regex are scary, here's a link with the flow of the regex
+# https://regexper.com/#PLUMED_REGISTER_ACTION%5Cs*%5C%28%5Cs*%5B0-9a-zA-Z_%5D%2B%5Cs*%2C%5Cs*%22%28%5B0-9a-zA-Z_%5D%2B%29%22%5Cs*%5C%29
+regex='PLUMED_REGISTER_ACTION\s*\(\s*[0-9a-zA-Z_]+\s*,\s*"([0-9a-zA-Z_]+)"\s*\)'
```
Iximiel, line 8: In the mac CI seems to ignore the action regitered like these:
```
PLUMED_REGISTER_ACTION(Dimer, "DIMER")
PLUMED_REGISTER_ACTION(DynamicReferenceRestraining, "DRR")
PLUMED_REGISTER_ACTION(EDS, "EDS")
PLUMED_REGISTER_ACTION(FuncPathGeneral, "FUNCPATHGENERAL")
PLUMED_REGISTER_ACTION(JCoupling, "JCOUPLING")
PLUMED_REGISTER_ACTION(Loss, "MAZE_LOSS")
PLUMED_REGISTER_ACTION(Memetic, "MAZE_MEMETIC_SAMPLING")
PLUMED_REGISTER_ACTION(OptimizerBias, "MAZE_OPTIMIZER_BIAS")
PLUMED_REGISTER_ACTION(Random_Acceleration_MD, "MAZE_RANDOM_ACCELERATION_MD")
PLUMED_REGISTER_ACTION(Random_Walk, "MAZE_RANDOM_WALK")
PLUMED_REGISTER_ACTION(Simulated_Annealing, "MAZE_SIMULATED_ANNEALING")
PLUMED_REGISTER_ACTION(Steered_MD, "MAZE_STEERED_MD")
PLUMED_REGISTER_ACTION(fusionPoreExpansionP, "FUSIONPOREEXPANSIONP")
PLUMED_REGISTER_ACTION(fusionPoreNucleationP, "FUSIONPORENUCLEATIONP")
PLUMED_REGISTER_ACTION(memFusionP, "MEMFUSIONP")
```
with only a space after the ","

GiovanniBussi:
I guess the script is not portable to different (non gnu) core utils

Iximiel:
I changed the regex to something more restrictive, now accepts only standard whitespaces instead of `\s` that is "any kind of space" and it works.
I still do not understand why `,\s*` does not work but `, *` yes, since `\s` do not cause problem in any other place... But I guess that the important things is that it works

PR #1064, opened 2024-04-29
Removed Doxygen characters from first line of documentation

gtribello:
In the nest and tutorials the first line of the documentation for actions appears in the tooltip that appears when you hover over the action name.  This line of text should thus provide a brief summary of what the action does.  To make this renderable this description should not include anything that needs to be rendered by Doxygen (i.e. equations and ref commands).  I found some places in PLUMED where people have included equations and ref commands in that first line of doxygen input.

This commit adds a check when the json syntax file is rendered to ensure folks do not have backslash characters in that first line of syntax. I think this should capture all the doxygen commands that people might use in these lines.

It also removes the few places where people have included equations and so on in that first line of the documentation

##### Description

See description above

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/json/Makefile
+++ b/json/Makefile
@@ -35,4 +35,4 @@ endif
 endif
 
 clean:
-	rm -fr syntax.json 
+	rm -fr syntax.json action_list
--- a/src/analysis/SelectWithStride.cpp
+++ b/src/analysis/SelectWithStride.cpp
@@ -24,7 +24,7 @@
 
 //+PLUMEDOC LANDMARKS LANDMARK_SELECT_STRIDE
 /*
-Select every \f$k\f$th landmark from the trajectory.
+Select every kth landmark from the trajectory.
 
 \par Examples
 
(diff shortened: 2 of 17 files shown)


GiovanniBussi (approved):
OK for me!

json/Makefile
```diff
@@ -35,4 +35,4 @@ endif
 endif
 
 clean:
-	rm -fr syntax.json 
+	rm -fr syntax.json action_list
```
GiovanniBussi, line 38: @gtribello is this change related to this PR?

gtribello: Not necessarily.  I think it fixes a typo I made.  When you run that json makefile it creates two files: syntax.json and action list.  If you run make clean those two files should be deleted as these are what the makefile creates.

What should I do to resolve this type of problem in general.  Just commit and push it separately and say it fixes a typo?

GiovanniBussi: It's fine I was just wondering.

PR #1070, opened 2024-05-07
Fix for plumed2html

gtribello:
##### Description

Hello @GiovanniBussi I know I said I was done with big PRs for a while but.... sorry.

I think this one addresses an oversight that we made when we added descriptions of the components to the manual.  When we did that we didn't add descriptions for the value with the action label.  If you thus read the description of the value in the syntax file it just says "this is a scalar".

In this PR I am thus adding these docstrings for all the values calculated by actions.  I think this will make the descriptions that we have of the inputs in the nest richer.  

I also realised that there was some stuff in Value and MultiValue that was part of vesselbase.  I am no longer using any of this so I deleted it all.

Lastly, made some changes to driver so that you can output a dictionary describing all the values in the input at the same time as you output all the shortcut files.  I hope to use this for generating the annotated inputs in the nest and tutorials in the future.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hello @GiovanniBussi 

Thanks for the information on the scripting for the regtests.  I have now used this in the renewed regtests.  In most cases we thus don't check documentation and it can be changed without breaking regtests.  I kept some in, though, because there are  some special cases where the doctoring that is output by plumed actually depends on the input file.  For example, if you use READ the description of the value tells you want column of the file was read in.

I also checked what happens if you forget to add a description of a value.  The code still runs but it outputs a warning.  You can see why that happens if you look at the code here:

```c++
void ActionWithValue::addValue( const std::vector<unsigned>& shape ) {
  if( !keywords.outputComponentExists(".#!value") ) warning("documentation for the value calculated by this action has not been included");
  plumed_massert(values.empty(),"You have already added the default value for this action");
  values.emplace_back(Tools::make_unique<Value>(this,getLabel(), false, shape ) );
} 
```

Lastly, I checked what happens to the nest (in general) if you have a load action.  You can see an example here:

https://www.plumed-nest.org/eggs/23/027/data/plumed.dat.html

It is actually quite nifty.  You get a tooltip that tells you that the feature was loaded on the name of the loaded action.  It doesn't generate the tooltips from the syntax or the description of the value.  It still shows you how the value from the action were used in the input file, though.

I hope that all means this PR is really to merge.  Let me know what you think, please.

GiovanniBussi:
Looks good to me  thanks!

PR #1076, opened 2024-05-15
Compiling with NVHPC

Iximiel:
##### Description

In this PR I will set up some modifications to the code in order to make possible to compile all plumed with nvhpc v24.3.

As now is a work in progress, I do not know if the CI will even start and some regtests are failing on my pc, the PR is set up to check easily if the modifications do not break other compilers against the master branch.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/tools/MergeVectorTools.h
+++ b/src/tools/MergeVectorTools.h
@@ -26,7 +26,12 @@
 #include <vector>
 #include <algorithm>
 #include <type_traits>
-
+#ifdef __NVCOMPILER
+#define PLMDAUTO typename C::value_type
+#else
+//I prefer letting gcc doing his work
+#define PLMDAUTO auto
+#endif
 namespace PLMD {
 
 namespace mergeVectorTools {
@@ -51,7 +56,7 @@ static void mergeSortedVectors(const C* const* vecs, std::size_t size, std::vect
     /// to allow using a priority_queu, which selects the highest element.
     /// we here (counterintuitively) define < as >
     bool operator< (Entry const& rhs) const { return top() > rhs.top(); }
-    const auto & top() const { return *fwdIt; }
+    const PLMDAUTO & top() const { return *fwdIt; }
     void next() { ++fwdIt;};
   };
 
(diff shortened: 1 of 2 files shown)


src/tools/MergeVectorTools.h
```diff
@@ -51,7 +51,13 @@ static void mergeSortedVectors(const C* const* vecs, std::size_t size, std::vect
     /// to allow using a priority_queu, which selects the highest element.
     /// we here (counterintuitively) define < as >
     bool operator< (Entry const& rhs) const { return top() > rhs.top(); }
-    const auto & top() const { return *fwdIt; }
+#ifdef __NVCOMPILER
+#define PLMDAUTO typename C::value_type
+#else
+//I prefer letting gcc doing his work
+#define PLMDAUTO auto
```
GiovanniBussi, line 58: I would avoid setting a macro `PLMDAUTO` in a .h file which could be included somewhere else. I think this is fine also with gcc:
```
// nvc++ cannot deduce the type correctly
const typename C::value_type & top() const { return *fwdIt; }
```

Also, I recently noticed this also with `icpc` installed on SISSA modules. So maybe it is a widespread bug

src/tools/MergeVectorTools.h
GiovanniBussi, line 1: I would leave a comment like
// nvc++ and icpc do not deduce the type

otherwise sooner or later we will "fix it" back to "auto"

Iximiel:
> @Iximiel it looks ready to be merged, I would retarget it for v2.10 though so that it is easier for other people wanting to contribute gpu code to do it

I retargeted v2.10, since it was the original aim for this, can you rerun the CI? I do not remember why -debug- and -intel- are failing

carlocamilloni:
@Iximiel it cannot be rerun from web because it is too old, you should push some empty or amend commit i think

Iximiel:
> @Iximiel it cannot be rerun from web because it is too old, you should push some empty or amend commit i think

I removed the `mpicc++ --version` debug line from the WF

Iximiel:
@carlocamilloni Now it passes -debug- and -intel-, I suspect that was the line I used to "debug" the mpicc version

> I removed the `mpicc++ --version` debug line from the WF

PR #1081, opened 2024-05-22
Fixes to documentation generators for PlumedToHTML

gtribello:
##### Description

I have been doing some work on improving the PlumedToHTML code that generates the annotated input files for the tutorials and nest.  The work is described in [this PR](https://github.com/plumed-nest/plumed-nest/pull/118).  As part of this I have had to make some modifications to PLUMED.  In particular, when you call PLUMED driver in parse only mode you can ask it to output two json dictionaries that contain:

1. How any shortcut keywords are expanded into the longer input
2. The definitions of the quantities that have labels in the input file

Getting the definitions of the labels directly from plumed when you run driver is superior to getting this information from the syntax file as:

1. You are able to distinguish whether objects are scalars, vectors, matrices etc this way.
2. You can find what the labels for actions that have custom labels (i.e. labels that depend on the arguments).
3. This way you can get the meanings of components that are created by files that are created by LOAD actions.

I tested the new version of PlumedTo HTML on all the inputs in the plumed nest and found a small number places where the new code does not work.  This PR fixes those problems.

@GiovanniBussi can you please check whether I have done the code for the first commit correctly?  The problem here is that I need to call registerKeywords for all the actions defined in the input file.  This will not work if you do:

```` 
actionRegister().getKeywords( name, keys )
````

The problem with that is that the code for the actions in LOAD is not accessible that way.  I thus came up with a way of doing by looking at what you did for creating actions.  I hope this is correct.  The tests that were not working now run.

Also note that I found one input in the nest like this:  

```` 
d1.mydist: DISTANCE ATOMS=1,2
PRINT ARG=* 
````

I managed to get it working but I also put a warning in about putting a . in the names of actions.  I think this is a recipe for trouble.  Not sure if we should make that an error.  What do you think?

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/cltools/Driver.cpp
+++ b/src/cltools/Driver.cpp
@@ -787,16 +787,16 @@ int Driver<real>::main(FILE* in,FILE*out,Communicator& pc) {
           }
           ActionWithValue* av=dynamic_cast<ActionWithValue*>(pp.get());
           if( av && av->getNumberOfComponents()>0 ) {
-            Keywords keys; p.getKeywordsForAction( av->getName(), keys ); 
+            Keywords keys; p.getKeywordsForAction( av->getName(), keys );
             if( firsta ) { valuefile.printf("  \"%s\" : {\n    \"action\" : \"%s\"", av->getLabel().c_str(), keys.getDisplayName().c_str() ); firsta=false; }
             else valuefile.printf(",\n  \"%s\" : {\n    \"action\" : \"%s\"", av->getLabel().c_str(), keys.getDisplayName().c_str() );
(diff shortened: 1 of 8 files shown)


GiovanniBussi:
>I hope this is correct. The tests that were not working now run.
Great, it means that the code was not too bad :-). I should probably replace the `std::vector<void*>` with some custom class `std::vector<DLimage>` or so.

>I managed to get it working but I also put a warning in about putting a . in the names of actions. I think this is a recipe for trouble. Not sure if we should make that an error. What do you think?

I would actually forbid this. Or we might start phasing it out and give a warning now (and a failure in plumed 2.11)

PR #1087, opened 2024-06-14
Action for reading masses and charges from a PDB file

gtribello:
##### Description

While creating all the stuff for the test centre I realised that some codes (I-Pi) don't have any mechanism for passing masses and charges of atoms to PLUMED.  Furthermore, I'm told it would be difficult to add the code to pass masses and charges to PLUMED using these codes.  I thus created this action that allows you to set the values of the masses and charges from the PLUMED input.  To set the masses and charges you just give the action a PDB file with the masses and charges in the beta and occupancy columns in the usual way.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
@gtribello I think this is useful, but I would not use PDB format. In PLUMED we only use PDBs when we have to provide coordinates (for RMSD) or names / residues (for MOLINFO). Here it's just reading a vector, I would save the user from the potential issues with misformatted files.

Why not using the same format that you can now write (with [DUMPMASSCHARGE](https://github.com/plumed/plumed2/blob/master/src/generic/DumpMassCharge.cpp)) and read (with `driver --mcfile`)?

In addition, I would call the action `READMASSCHARGE` for consistency.

If you like the PDB, we can add the option of writing a PDB with charges in DUMPMASSCHARGE and reading it in READMASSCHARGE, but I am not sure it is a good idea.

gtribello:
Thanks for these comments @GiovanniBussi.  I have changed the name as you suggest now and have also added the option to read either a pdb file or the mass charge file.  This is the same as what we allow with driver so I think it makes more sense.

Thanks again.

GiovanniBussi:
> This is the same as what we allow with driver
True, I didn't remember this... Thanks!

gtribello:
did we merge this?  I can't remember why we didn't.  Do you remember @GiovanniBussi?

Issue #1086, opened 2024-06-14
Error when patching plumed (v2.9.1) to Gromacs 23.5


zhongxiang117:
As a report, since this is caused by the new version of Plumed and Gromacs, I opened a new issue.

I downloaded the least version of plumed-v2.9.1, tried to patch Gromacs 2023.5, failed with the same error like #981, and solved  by the same solution mentioned below by using version of `git checkout fff0fa526`. Thus there must be some hidden problems in codes.

zhongxiang117:
Besides, there are one addition problem. If I compile plumed into a non-standard path (achieved by using `export DESTDIR=mydir; make`), after patch, in Gromacs root folder, there will be three plumed related files (symbolic link) added.
```
Plumed.cmake    # symbolic link, to my custom location, Good
Plumed.inc      # symbolic link, to my custom location, Good
Plumed.h        # symbolic link, to standard location,  Error!
```

Nonetheless, in whatever situations, these three files should be created at `runtime` patching, relative to the path where the `plumed patch` executes, not to its compiled paths.

GiovanniBussi:
Regarding [this](https://github.com/plumed/plumed2/issues/1086#issue-2352382656) I think you are right, I reopened issue #981 

Regarding [this](https://github.com/plumed/plumed2/issues/1086#issuecomment-2167133134) I think you are not expected to use `DESTDIR` to change the location. You should use `--prefix /path`. `DESTDIR` is to temporarily place the installed version in a different root directory for staging. This allows for instance to do `make install` without root permission, pack the installation directory in a single zip file, then unpack it to the correct prefix with root permissions.

I close this issue now, please reopen it if the answer is not clear. Thanks for reporting!

zhongxiang117:
Hello @GiovanniBussi, thanks for your work!

For my second report of compilation path, I surely know how does `DESTDIR` work, and it is not a standard way to install package. But, I do not think it is a proper way to "hard-coded" it, I mean by using absolute path to create symbolic link, inside the source codes. As I mentioned, the files used for patching should be resolved during runtime execution, it should not depend on pre-configrautions where `--prefix` specifies.

GiovanniBussi:
I am not sure what's special with files used during patching. 

Unfortunately we need to hard code paths for other reasons. It might be possible to strip down the absolute path form our dynamic library using dladdr (but it is not 100% supported) and some bash trick in our scripts. There are a couple of places however where I think we cannot, namely:

[...]

PR #1088, opened 2024-06-14
dynamically resolve environment variables

zhongxiang117:
##### Description

This PR is to use solve the problem mentioned in #1086 section 2 about the dynamical link of "Plumed.h" header file.

##### Type of contribution

##### Copyright

##### Idea for update

I thoroughly go over the `plumed=patch` codes, after compilation, I found the call sequence of `plumed=patch` is:

```
a) plumed patch -p
    1) bin/plumed                       # plumed executor
    2) lib/plumed/scripts/patch.sh      # script interface
    3) lib/plumed/patches/patch.sh      # real codes do patch

b) plumed-patch -p
    1) bin/plumed-patch                 # plumed-patch executor
    2) lib/plumed/scripts/patch.sh      # script interface
    3) lib/plumed/patches/patch.sh      # real codes do patch
```

Be aware of that call `a` and `b` are different, `a` is run `plumed` first, and then use its internal solver to find patch codes, while `b` is directly run patch script.

The difference is environment variable `PLUMED_INCLUDEDIR` is hard-coded inside `plumed` binary object, where inside `plumed-patch` script, it can be customized.

Thus, to make it work, I made edition of two files, 1) `src/lib/Makefile`: add some lines to check environment variables before scripts executes, relating to `b`. 2) `patches/patch.sh`: to exclusively check `PLUMED_INCLUDEDIR` validation, relating to `a`.

I think the explanations should be clear, if there are any opaques, please let me know, thanks!

### DIFF
--- a/patches/patch.sh
+++ b/patches/patch.sh
@@ -284,14 +284,19 @@ case "$action" in
       test -n "$quiet" || echo "Executing plumed_before_patch function"
       plumed_before_patch
     fi
+    if test -f "$PLUMED_INCLUDEDIR/$PLUMED_PROGRAM_NAME/wrapper/Plumed.h"; then
+      header="$PLUMED_INCLUDEDIR/$PLUMED_PROGRAM_NAME/wrapper/Plumed.h"
+    else
+      header=`realpath "$PLUMED_ROOT/../../include/plumed/wrapper/Plumed.h"`
+    fi
     if test -n "$include" ; then
       test -n "$quiet" || echo "Including Plumed.h, Plumed.inc, and Plumed.cmake ($mode mode)"
-      echo "#include \"$PLUMED_INCLUDEDIR/$PLUMED_PROGRAM_NAME/wrapper/Plumed.h\"" > Plumed.h
+      echo "#include \"$header\"" > Plumed.h
       echo "include $PLUMED_ROOT/src/lib/Plumed.inc.$mode" > Plumed.inc
       echo "include($PLUMED_ROOT/src/lib/Plumed.cmake.$mode)" > Plumed.cmake
     else
       test -n "$quiet" || echo "Linking Plumed.h, Plumed.inc, and Plumed.cmake ($mode mode)"
-      ln -fs "$PLUMED_INCLUDEDIR/$PLUMED_PROGRAM_NAME/wrapper/Plumed.h" Plumed.h
+      ln -fs "$header" Plumed.h
       ln -fs "$PLUMED_ROOT/src/lib/Plumed.inc.$mode" Plumed.inc
       ln -fs "$PLUMED_ROOT/src/lib/Plumed.cmake.$mode" Plumed.cmake
     fi
--- a/src/lib/Makefile
+++ b/src/lib/Makefile
@@ -446,16 +446,36 @@ ifeq ($(PLUMED_INSTALL),Install)
 # If plumed is installed, single paths can be overridden using env vars
 	@echo "export PLUMED_IS_INSTALLED=yes" >> $@
 	@echo "export PLUMED_ROOT=\"\$${PLUMED_ROOT-$(libdir)/$(program_name)}\"" >> $@
-	@echo "export PLUMED_INCLUDEDIR=\"\$${PLUMED_INCLUDEDIR-$(includedir)}\"" >> $@
-	@echo "export PLUMED_HTMLDIR=\"\$${PLUMED_HTMLDIR=$(htmldir)}\"" >> $@
+	@echo "PLUMED_INCLUDEDIR=\"\$${PLUMED_INCLUDEDIR-$(includedir)}\""      >> $@
+	@echo "if test -d \"\$$PLUMED_INCLUDEDIR/plumed\"; then"                >> $@
+	@echo "    export PLUMED_INCLUDEDIR=\$$PLUMED_INCLUDEDIR"               >> $@
+	@echo "else"                                                            >> $@
+	@echo "    export PLUMED_INCLUDEDIR=\$$PLUMED_ROOT/../../include"       >> $@
+	@echo "fi"                                                              >> $@
+	@echo "PLUMED_HTMLDIR=\"\$${PLUMED_HTMLDIR=$(htmldir)}\""               >> $@
+	@echo "if test -d \"\$$PLUMED_HTMLDIR\"; then"                          >> $@
+	@echo "    export PLUMED_HTMLDIR=\$$PLUMED_HTMLDIR"                     >> $@
+	@echo "else"                                                            >> $@
+	@echo "    export PLUMED_HTMLDIR=\$$PLUMED_ROOT/../../share/doc/plumed" >> $@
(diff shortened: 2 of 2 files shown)


GiovanniBussi:
Hi, thanks for the contribution. I am not sure this is robust. If I understand correctly:

- You check if the usual path (the one used now) exist and if so you use that
- If not, you use `PLUMED_ROOT/../../include`

This is fragile for a number of reasons:
- If another version of plumed has been already installed on that path, the solver will use that
- You are assuming that include dir is identical to prefix/include, and that libdir is identical to prefix/X/Y (with X Y arbitrary directories). This is not enforced anywhere else I think, and in some distributions it is not true.

Regarding your comments, let me add that:
- You are not supposed to use the `patches/patch.sh` script directly. This is very much like the scripts present in `scripts`. It should be called via a wrapper that setups the environment correctly. Let me know if there is any documented use of the `patches/patch.sh` script, which should be fixed.
- Even though the path to include files is hard coded in the binary executable, it can [still be adjusted](https://github.com/plumed/plumed2/blob/2e07dd1ebd5ac717b16f86a9eb5a481a384a9f20/src/config/Config.inc.in#L88-L98). Did you try to set `PLUMED_INCLUDEDIR` before running plumed?
- Using `PLUMED_ROOT` to adjust this is a last resort.

I am still a bit confused about what you are trying to provide. Something like this?

- install plumed to a prefix
- move it somewhere else
- set some env var to consider the new path
- patch an MD code
- compile the MD code

I think that the solution for relocatability is to use:
- 'dladdr` in the cpp code that I linked above to know where the library is
- some bash trick in the bash wrappers to find the path of "this" executing script

This will still have the open issues that I mentioned [here](https://github.com/plumed/plumed2/issues/1086#issuecomment-2167273011). But it would allow you to relocate plumed and use it (without setting and env var), though on Linux only.

zhongxiang117:
Thanks for your comments. Now I know where your confusion is, before that let me explain the things you have talked about.

1) If `PLUMED_ROOT/../../include` is dangerous, I can normalize that by using `realpath` command.
2) I am not directly using `patches/patch.sh` script, the above writing is to show the call sequences when the script under `bin` folder executes. After compilation and installation, there are three files under `bin` folder, which is supposed to be added to `PATH` environment variable, then user can have access to `plume`, `plume-config`, and `plume-patch` commands. So I tested the differences between `plume` and `plume-patch` scripts, and found that they do have differences.

This is not about on the installation stage, installation can be anywhere as long as configurations, `./configure`, success, no matter whether `--prefix` or `DESTDIR` used or not.

What I am trying to provide is to make the environmental setting more concise and standard, I mean usually the program will be compiled and installed in a standard way, like installing to `PREFIX/bin`, `PREFIX/lib`, and etc.. Seldom, normally for advanced users, the files will be separately installed to `PREFIX_A/bin`, `PREFIX_B/lib`, and etc. customizing each type of object paths.

So after installation, even for different versions of program, the compiled files under like `PREFIX/bin`, `PREFIX/lib`, and so on, the paths of them will be corresponded to each other. Thus, the way I provide is to by only setting “PROGRAM_ROOT” once, all other its same-time compiled files can be relocated automatically, when using patch scripts.

I know by additionally setting `PLUMED_INCLUDEDIR` to different and the “same-time” installed folder works and meets my needs, but I want to solve it in a more standard way by only using `PLUMED_ROOT` variable, and I don't find any docs talking about this, if has, please enlighten me and let me know.

More important, the update only take effects on patch scripts, and make it more robust to find include headers, it will not influence how `bin/plumed` executable to relocate configuration files or static library(if compiled in this way), so your worries are not a problem.

GiovanniBussi:
Ok let me try again to understand what you are trying to do.

1. install plumed somewhere
2. move the installed package somewhere else
3. export PLUMED_ROOT and fix PATH
4. use plumed patch to patch an MD code

With the current implementation, if you skip step 2 everything works as expected, correct?

At point 4 you are only interested in `plumed patch`, correct?

If so, I guess your solution should not work. Even if you fix the paths seen in the patch.sh script so that the links work correctly, the make and cmake include files will still contain absolute paths. Please have a look at files `$prefix/lib/plumed/src/lib/Plumed.cmake` and `$prefix/lib/plumed/src/lib/Plumed.make`.

The only solution, to my understanding, is to:
- modify the patch script so that it copies those files rather than providing links
- while copying those files, edit them on the fly to adjust absolute paths

Alternatively, if possible, modify the `Plumed.cmake` and `Plumed.make` files so that they can automatically find their own location when included. I know this is possible with bash scripts (see e.g. [here](https://stackoverflow.com/questions/4774054/reliable-way-for-a-bash-script-to-get-the-full-path-to-itself)). I don't know if it's possible for make and cmake files.

Related to the problem of having `include` directory not equal to `plumed_root/../../include`, I think the clean solution is to find the correct relationship as a [relative path](https://stackoverflow.com/questions/2564634/convert-absolute-path-into-relative-path-given-a-current-directory-using-bash). I guess you can do it at run time, when the directories exist already.

zhongxiang117:
Hi @GiovanniBussi, your understanding of what I am trying to do is correct. I should have explained it at the very beginning.

I know where is your worry about, it is not about the compilation stage, or any `make` or `cmake` files.

The PR of new codes just adds an additional layer when running `plumed patch` or `plumed-patch`, or other BASH scripts, it will help find the correct `PLUEMD_INCLUDEDIR` path.

Let us assume the compilation has been done, the `plumed` program has been successfully created with all its needed files and objects into a compact folder, and we know some paths inside `bin/plumed` binary uses absolute path, and it works as expect.

Now, I move this folder to another place, and correctly setting `PLUMED_ROOT` to that new path, the `plumed` still works(this is very important), but patches cannot find its correct header file(this is the problem that I encountered).

Thus, I provided a way to solve it, by firstly checking the validation of the folder path before doing really patch. It will not influence all other settings, it simply adds a second layer to help locate header files. If no folder path changes, those codes will be never run.

To conclude, there are two very important things that I should mention, to relieve any of your concerns.

1) The codes in PR only adds a warranty to help set the correct `PLUMED_INCLUDEDIR` path when something happens
2) `bin/plumed` will work as always, no matter its path of folder renamed or moved to other place (I have tested it, 100% sure)

zhongxiang117:
One more thing, the "standard" way to solve the relative path `plumed_root/../../include`, the same solution can be used as referenced from `configure` file, by using `realpath` during its runtime.

Also, I see all those scripts are generated and assumed to be run `#!/usr/bin/env bash` environment, no other type SHELLs are defined, thus, I do not think it is a big problem.

GiovanniBussi:
I see that your patch is not affecting the rest of the code. It is just making `plumed patch -p` able to properly patch MD codes after plumed has been moved, adding symbolic links to the correct files. To my understanding, however, **it will not be possible to compile the patched code** (because the cmake and make include files still contain the absolute path to the location of plumed before it was moved). Thus, I don't think it is worth merging it.

I am closing this but feel free to reopen if there's something I misunderstood.

PR #1090, opened 2024-06-27
Removed old-style-cast from Plumed.h

Iximiel:
##### Description

I removed the old-style casts in Plumed.h.

Now when Plumed.h is in "c++ mode" will use static_cast or reintrpet_cast 

I used `-Werror=old-style-cast` to find the casts and i changed to static when gcc suggested to change to static and reintepret otherwise or when no suggestion were made.

I am not 100% sure to have changed every casts, but at least the ones that affects the gromacs compilation are done plus some more found with searching `\(.*?\)\w`.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/wrapper/Plumed.h
+++ b/src/wrapper/Plumed.h
@@ -3852,11 +3852,16 @@ __PLUMED_WRAPPER_INTERNALS_END
 /**
   Utility to search for a function.
 */
+#ifdef __cplusplus
+#define __PLUMED_WRAPPER_SEARCHF_CAST(func,tmpptr) func=reinterpret_cast<decltype (func)>(tmpptr)
+#else
+#define __PLUMED_WRAPPER_SEARCHF_CAST(func,tmpptr)  *(void **)&func=tmpptr
+#endif
 #define __PLUMED_SEARCH_FUNCTION(tmpptr,handle,func,name,debug) \
   if(!func) { \
     tmpptr=dlsym(handle,name); \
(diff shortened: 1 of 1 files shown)


Iximiel:
> Indeed, I see you only use it in regions of the code that are C++ only, so no reason to use this macro

I changed it in the std::filesystem part
but some of them are in the f2c, c2f, c2v and v2c functions that are wrapped by `__PLUMED_WRAPPER_C_BEGIN/END`, are also these only c++?

GiovanniBussi:
You are right, those are also for C. So we can either keep it as is or maybe you can put explicit reinterpret_cast in the filesystem part, just because it's clearer...

Iximiel:
> You are right, those are also for C. So we can either keep it as is or maybe you can put explicit reinterpret_cast in the filesystem part, just because it's clearer...

i changed CAST to REINTERPRET_CAST, so it is clearer, and in the c++ parts I wrote it explicitly

Iximiel:
The changes should not be accepted, using clang12 for compiling raises a warning about some of the cast to be undefined behaviour. I think I should address this before considering the work done

Iximiel:
I think I solved the issue, I do not get anymore the UB warnings. And the solution should be compatible with C++98

GiovanniBussi:
@Iximiel I think I can merge this, but you should first make it "not draft" I think

Issue #1075, opened 2024-05-14
Writing the coordination with openACC


Iximiel:
As I did with CUDA(#1028) and I tried to do with Arrayfire(#1049) and [pytorch](https://github.com/Iximiel/plumed2/tree/torchCoord/plugins/torchCoord), I tried to [rewrite](https://github.com/Iximiel/plumed2/tree/openACCcoord/plugins/openACCCoord) the COORDINATION cv with _openACC_ as accelerator.

Here's the result, using the new benchmark tool

![sc_100](https://github.com/plumed/plumed2/assets/5535617/12d1e0d0-86aa-47e3-b800-8f2d32399e5e)
Is slower than CUDA, but writing in openACC may be more familiar, because it looks like openMP and also because you can leave the compiler to guess how to implement the parallelization of the loops and you do not have to use the `<<<>>>` to launch kernels like in  CUDA. And is way more flexible than the tensor libraries.
On the compilation I have some mixed feelings, as you can read in the spoiler below.

<details>
  <summary>Details about compilation and script used</summary>

I run everything on my workstation (NVIDIA T1000 8GB + AMD Ryzen 5 PRO 5650G)
I used nvhpc24.3, downloaded already compiled from the Nvidia site.

The environment used is actually slightly complex:
I compiled plumed from master with plain gcc+mpi
Then I compiled the plugin with my wild Makefile that uses nvc++ for the accelerated part and g++ for the main body of the CV.
Then I ran the benchmark **without nvhpc in the environment**, because it conflicts with the mpi that I used with plumed:
```bash
nsteps=100
list_of_natoms="500 2000 4000 6000 8000 10000 12000 14000 16000"
export PLUMED_NUM_THREADS=8
useDistr="line sc"
useDistr="sc"

for distr in $useDistr; do
  for natoms in $list_of_natoms; do
    fname="${distr}_wACC_${PLUMED_NUM_THREADS}threads_${natoms}_Steps${nsteps}"
    plumed benchmark --plumed="plumed.dat:cudasingleplumed.dat:accplumed.dat" \
      --natoms=${natoms} --nsteps=${nsteps} --atom-distribution=${distr} >"${fname}.out"
    grep -B1 Comparative "${fname}.out"
  done
done
rm -f bck.*
```
(I have to try to make everything run compiled with plain nvhpc
[...]

PR #1092, opened 2024-07-01
Typed vector and tensor

Iximiel:
##### Description

I started to work on this in the context of #1076 and in sync with what I did with #1075.
My idea was to set up some types of plumed to be more flexible and usable in contexts like sycl, cuda, openACC or targetted openMP. Then I dropped this for a while, but I thought it could be interesting to bring it out again.

Most of the problems this new approach gives are in the friend functions within the vector and the tensor, they give strange warnings. I think due to the fact that I am declaring friend functions like
`template<typename U, unsigned m> U dotProduct(const VectorGeneric<U, m>&,const VectorGeneric<U, m>&);` to a ` template<typename T, unsigned n> VectorGeneric<T, n>`, with all types and all dimensions. On one hand, I do not think is a problem, but on the other hand, it does make me a little uncomfortable.

If I understood correctly those functions are declared friend to appear in the documentation page of Tensor/Vector, maybe by putting `\ingroup TOOLBOX` in their docstring they will appear in the toolbox, or in another doxygen module and we can remove the friend declarations.

In any case the part with the friend-declared functions must be finished because I applied two different approaches in Vector and Tensor: in tensor, I eliminated the friend nearly all the friend declarations because I was having an hard time in trying to compile with all of them.

---

I also added a `.data()` method to access the underline array in both classes. This new method should be used in place of the various `&variable[0]` to get the pointer to the first element of the array. But as now, I did not refactored it in the code (but I am using something similar in #1075 ).

##### Target release

##### Type of contribution?

##### Copyright

##### Tests


GiovanniBussi:
Quick note, replace `VectorGeneric<T,N>` with `VectorTyped<T,N>` to avoid breaking existing code, as we discussed

Iximiel:
Now this is passing cppcheck (hopefully), but I had to remove the `static_asserts` with the clearer error message
@GiovanniBussi I think I should also add `local_ssyevr`? Or I can keep it as a TODO?

carlocamilloni:
@Iximiel @GiovanniBussi is this ready to merge? The CI issue is not clearly related to the PR I think but there is an old comment from Daniele unaswered

GiovanniBussi:
> local_ssyevr

@Iximiel if possible and easy I would add local_ssyevr. so that `diagMatSym` can be used also on single precision vectors. Thanks!

Iximiel:
> > local_ssyevr
> 
> @Iximiel if possible and easy I would add local_ssyevr. so that `diagMatSym` can be used also on single precision vectors. Thanks!

I added ssyevr to `Tensor.h`, then for adding the tests for it I modified the calls to blas/lapack in Matrix.h

---

`rt-lowest-eigenpair` has a built in benchmark for the diagonalization, I think it is interesting that (at least on my machine) the new lowestEigenpairSym is faster for `double` than for `float`
(diagMatSym behave as expected)

PR #1093, opened 2024-07-04
Start on work on moving calculation of derivatives to apply loop for actions that pass vectors and matrices

gtribello:
##### Description

This is the PR for the benchmarking that we discussed on Wednesday.

**DO NOT MERGE**

I will provide more explanation tomorrow.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hi @Iximiel  and @GiovanniBussi 

I ran out of time to give the details of this yesterday so here they are.

You can calculate a coordination number using this version of the code using the following input:

````
c1: CONTACT_MATRIX GROUPA=1-3 GROUPB=1-512 SWITCH={EXP D_0=4 R_0=0.5 D_MAX=6}
ones: ONES SIZE=512
cc: MATRIX_VECTOR_PRODUCT ARG=c1,ones
mts: SUM ARG=cc PERIODIC=NO
PRINT ARG=mts FIle=colvar
BIASVALUE ARG=mts
````

I think it should be possible to reproduce the benchmarks that Daniele was doing on the various versions of the coordination number using the input above.

By default everything the calculation will be run in the same way that it is run in the current master version of the code.  That is why all the regtests are passing for this branch.  Basically the regtests are doing the same thing that they would be doing in the current master.  In other words, when the regtests are run all the linked lists are still used.

To run without using the linked lists you use the following environment variable:

````
export PLUMED_FORBID_CHAINS=yes
````

It then does all the calculation of forces during the backwards (apply) loop.

It would be really nice if we could look at the performance of the code with the following options:

* Calculation without BIASVALUE (i.e. no derivatives) and without PLUMED_FORBID_CHAINS option set.
* Calculation without BIASVALUE (i.e. no derivatives) and with PLUMED_FORBID_CHAINS=yes option set.
* Calculation with BIASVALUE (i.e. with derivatives) and without PLUMED_FORBID_CHAINS option set.
* Calculation with BIASVALUE (i.e. with derivatives) and with PLUMED_FORBID_CHAINS=yes option set.

If we could have some slides with the performance of those four options by next Wednesday so we can discuss next steps in the meeting that would be really great.  The benchmarks I am thinking off would look at the performance of the code as you increase the number of atoms.  Essentially the sorts of grants @Iximiel has been producing to look at the various GPU flavours.

Thanks

**Comment by Iximiel** [comment]
@gtribello I am setting up the benches right now

A couple of questions:
 - In how the atom systems of the benches are set up, the first 3 atoms in each system are in a corner, and I think that this might test the branch `if dist>dmax then skip` of the various systems for most of the couples
   - If I do not change the input file: How should I scale the systems? `3*n` vs `512*n` or `3` vs `512*n`?

**Comment by gtribello** [comment]
Hi @Iximiel 

I don't understand the queries.  However, I made the input more complicated than it needed to be.  You can do:

````
c1: CONTACT_MATRIX GROUP=1-100 SWITCH={EXP D_0=4 R_0=0.5 D_MAX=6}
ones: ONES SIZE=100
cc: MATRIX_VECTOR_PRODUCT ARG=c1,ones
mts: SUM ARG=cc PERIODIC=NO
PRINT ARG=mts FIle=colvar
BIASVALUE ARG=mts
````

To increase the number of atoms, you change the number of indices specified to GROUP in the CONTACT_MATRIX.  You must also increase the SIZE in ONES to the number of atoms.

Does that help?

**Comment by Iximiel** [comment]
Ok, I'll go with  `GROUP=@mdatoms` then, I was wondering if there was a reason for the groupA-groupB  input.

On that note, there is a special kw for the number of atoms for `ONES SIZE=?`

**Comment by gtribello** [comment]
There is no special keyword for the number of atoms for ONES sorry.

**Comment by Iximiel** [comment]
As you asked, here's my benchmark set up, the example is for a local run, on LEONARDO I create a single run file per natoms-distr combination:

```bash
#!/bin/bash

nsteps=1000
list_of_natoms="500 2000 4000 6000 8000 10000 12000 14000 16000"
export PLUMED_MAXBACKUP=0

export PLUMED_NUM_THREADS=1
mpiprocesses=1
useDistr="line sc"

function setFname() {
  echo "${distr}_${mpiprocesses}mpi_${PLUMED_NUM_THREADS}threads_${natoms}_Steps${nsteps}"
}
mpicmd=""
if [[ $mpiprocesses -gt 1 ]]; then
  mpicmd="mpirun -n $mpiprocesses"
fi

if [[ $1 = run ]]; then
  module purge
  module load gnu9/9.4.0 openmpi4/4.1.1
  module load plumed/master09-07
  for distr in $useDistr; do

    for natoms in $list_of_natoms; do

      fname=$(setFname).out

      echo "output file: ${fname}"
      $mpicmd plumed benchmark \
        --plumed=plumed_master.dat:plumed_other.dat \
        --kernel=this:../../src/lib/install/libplumedKernel.so \
        --nsteps="${nsteps}" \
        --natoms="${natoms}" \
        --atom-distribution="${distr}" \
        >"${fname}"

      grep -B1 Comparative "${fname}"

    done
  done
fi
#postprocess the output
for distr in $useDistr; do
  echo "collecting ./$distr"
  {
    for natoms in $list_of_natoms; do
      echo -n "$natoms "
      fname=$(setFname).out
      grep Calculating "${fname}" |
        awk '{printf "%f ", $7}'
      echo ""
    done
  } >"timing_calculate_${distr}_s${nsteps}.data"
  {
    for natoms in $list_of_natoms; do
      echo -n "$natoms "
      fname=$(setFname).out
      sed -n '/PLUMED: *Cycles *Total *Average *Minimum *Maximum/{n ; p}' "${fname}" |
        awk '{printf "%f ", $3}'
      echo ""
    done
  } >"timing_total_${distr}_s${nsteps}.data"
done
```

`--plumed` is the list or the plumed file,  `--kernel` is the kernel or the list of kernels
you can have a `1 inputfile - n kernels` or a `n inputfiles - 1 kernel` or `n inputfiles - n kernels` and kernel1 will get input1, kernel2 input2 and so on

Iximiel:
Hi @gtribello!
Here's the results with the last commit d15c12f00d2b98f103e6bb43bc0035de43795095 (green) and the old version https://github.com/plumed/plumed2/pull/1093/commits/b9b8969c9abb7476ab4f5ec375b887197bb32438 (red)
image

the old one uses the time I measured yesterday

gtribello:
Hi @Iximiel 

Thanks for running the benchmarks again.

Iximiel:
@gtribello here are the benchmarks I did on LEONARDO:
image

image

The color code is the usual: in lighter color the "4 Calculating (forward loop)" part.
Initially, I did the benches with `export PLUMED_NUM_THREADS=8`, but they diverged a lot from the one you posted, so I added a run with `PLUMED_NUM_THREADS=1`
everything was run with
```bash
fname=sc_${PLUMED_NUM_THREADS}threads_2000nat_500steps
plumed benchmark --plumed="sc_cm2000master.dat:sc_cm2000target.dat" \
  --kernel="/path/to/master/lib/libplumedKernel.so:/path/to/plumed-derivatives-from-backpropagation/lib/libplumedKernel.so" \
  --maxtime=840 \
  --natoms=2000 --nsteps=500 --atom-distribution=sc >"${fname}.out"
```

The plumed file is, adapted to the number of atoms
```
c1: CONTACT_MATRIX GROUP=1-4000 SWITCH={EXP D_0=4 R_0=0.5 D_MAX=6}
ones: ONES SIZE=4000
cc: MATRIX_VECTOR_PRODUCT ARG=c1,ones
mts: SUM ARG=cc PERIODIC=NO
PRINT ARG=mts FIle=sc_colvar-bias-master1706-4000
BIASVALUE ARG=mts
```

<details>
  <summary>I also did the same calculations, but with `--atom-distribution=line`, they are in the spoiler here: </summary>
  
image


image

  
</details>

gtribello:
Hello @Iximiel and @GiovanniBussi 

I've been working more on writing a version of PLUMED where the derivatives are calculated during the backward loop.  I've also gotten confused.  It would help me @Iximiel if you could run another benchmark just so that I can check that my conclusions about the state of the project are correct.  Sorry and thank you in advance.

So, firstly, the confusion.  I realised that the switching function parameters we used in that example input were not particularly sensible. With those switching function parameters, every single atom in the box with 8000 atoms was within the D_MAX cutoff of 6 nm that we set, which is not really how this type of CV is used. 

I thus created a new input with more sensible switching function parameters.  It is below:

````
c1: CONTACT_MATRIX GROUP=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} 
ones: ONES SIZE=8000
cc: MATRIX_VECTOR_PRODUCT ARG=c1,ones
mts: SUM ARG=cc PERIODIC=NO 
PRINT ARG=mts FILE=colvar 
BIASVALUE ARG=mts 
````

Separately @Iximiel it would be useful to add some detail to the manual for `plumed benchmark` about how the various possible configurations that can be used for benchmarking are setup.  I picked the switching function parameters above by trial and error.  It would be useful to understand what the configurations are.  In other words, it would be good to know  the typical spacing between atoms for the configurations used for Benchmarking.

When I run benchmarks, I get the following results:

new_speed_update

I am thus concluding the following:

1. My version of coordination is WAY faster than COORDINATION - hurrah for link cells.
2. The new version of my coordination is faster than the version in master if you are not calculating derivatives
3. The new version of my coordination is slower than the version in master if you are calculating derivatives.

Thanks.

**Comment by Iximiel** [comment]
Hi @gtribello, with coordination you simply used 
```
c: COORDINATION GROUPA=@mdatoms SWITCH={EXP D_0=4 R_0=0.5 D_MAX=6}

PRINT ARG=* FILE=@distr@_COORDColvar@natoms@ FMT=%8.4f STRIDE=1

FLUSH STRIDE=1
```
or something more elaborate?

---

I'll write something in the --help for plumed benchmark.
The rule of thumb is that the mean density should be ~ 1 atom pern nm^3 for the 100% random distribution (sphere, cube, etc.) and that the distance between first neighbors oscillates around 1 nm in the line and sc(simple cubic) distributions:
 - the line one is a line of atoms 1x1xnatoms in dimension, and I don't remember if it is compatible with pbcs
 -  with "sc" the atoms are placed in a simple cubic reticle. The fill rule is: take a sc reticle of qxqxq atoms where q^3 is the nearest upper cubic number to natoms (q^3=1,8,27...) and the atoms will fill the reticle from the bottom up

I think it is better to use sc for a "realistic system" and line for stressing the NN optimizations

**Comment by gtribello** [comment]
Hello again @Iximiel 

I did indeed write the input for coordination in the way you said.

I also used the sc for the atom distribution as I agree that it is a better test for the benchmarking. 

Separately, if you could add the list of options for the --atom-distribution flag in its description that would be a good first thing to add to the documentation.

**Comment by gtribello** [comment]
Hello again

I found a bottleneck and got the new version to go a bit faster.  You can see the plot here:

further-speed

What I wrote in my previous message is still true, though:

* With one thread and no derivatives, the new version is faster
* With four threads and no derivatives, master is faster
* With derivatives, master is faster with one and four threads

**Comment by Iximiel** [comment]
Hi @gtribello, I took the times for 7e0f435 on LEONARDO: with "sc" as the system
immagine
I did these benchmarks three times: in two of them that I am not showing here I got a suspicious t(8thread)>t(1thread), for 7e0f435 , and in the one I am showing the two times are similar

Whereas for the system "line" the 8 threads runs do not conclude the first step, the queue kills them due to excessive memory consumption (I asked for "only" 64000 MB due to the memory per user in the queue I am using), and it is a lot slower than for the "sc" one in the other cases

**Comment by Iximiel** [comment]
@gtribello I recompiled plumed with  `CXXFLAGS='-O3 -g -fno-omit-frame-pointer' CFLAGS="-g -O2 -fno-omit-frame-pointer" ./configure`, `no-omit-frame-pointer` is necessary for [this reason](https://www.brendangregg.com/FlameGraphs/cpuflamegraphs.html#C)

The plumed used for benchmark is [765f97f](https://github.com/plumed/plumed2/commit/765f97f45648d2bc4be89ce72fad31ef4a8c9169), from the master branch.

I tortured my poor workbench with:
```bash
nsteps=100
list_of_natoms="500 4000"
export PLUMED_MAXBACKUP=0
for system in sc line; do
  for threads in 1 4 8; do
    export PLUMED_NUM_THREADS=$threads
    for natoms in $list_of_natoms; do

      fname="${system}_${PLUMED_NUM_THREADS}threads_${natoms}nat_${nsteps}steps"

      flameDir=$HOME/scratch/repos/FlameGraph
      plumedfile="${system}_cmbias${natoms}.dat"
      sed -e "s/@natoms@/${natoms}/g" \
        -e "s/@distr@/${system}/g" \
        <"cmbiasTEMPLATE.dat" >"${plumedfile}"

      echo "output file: ${fname}.out"
      echo "plumedfile: ${plumedfile}"

      perf record -F99 --call-graph dwarf --output="${fname}.data" \
        plumed benchmark \
        --kernel="$HOME/installs/plumed-7e0f435/lib/libplumedKernel.so" \
        --plumed="${plumedfile}" \
        --natoms="${natoms}" \
        --nsteps="${nsteps}" \
        --atom-distribution=${system} \
        >"${fname}.out" &&
        perf script --input="${fname}.data" >"${fname}.perf" &&
        "${flameDir}"/stackcollapse-perf.pl "${fname}.perf" >"${fname}.folded" &&
        "${flameDir}"/flamegraph.pl "${fname}.folded" --subtitle "${fname}" >"${fname}.svg"

    done
  done
done
```
Here are the 3 images for the three sc/4000 atoms simulations

sc_1threads_4000nat_100steps
sc_4threads_4000nat_100steps
sc_8threads_4000nat_100steps


I made a package with all the SVG (if you open it with a browser you can zoom-click on the various columns, and other qol improvements)
[perf7e0f435.tar.gz](https://github.com/user-attachments/files/16360700/perf7e0f435.tar.gz)

gtribello:
HI @Iximiel 

Thanks for doing those flame graphs.  

I am very confused now.  There seems to be no evidence that much time is being spent in the backwards loop?  I would expect to see `ActionWithVector::checkForForces featuring more in those flame graphs.  Do you have any idea why that is not there?  Did you run with `export PLUMED_FORBID_CHAINS=yes` and with the BIASVALUE action?  If you look at the timings at the end of a typical calculation, you see:

````
PLUMED: 1 Prepare dependencies                           100     0.000274     0.000003     0.000001     0.000011
PLUMED: 2 Sharing data                                   100     0.360542     0.003605     0.000036     0.030037
PLUMED: 3 Waiting for data                               100     0.000242     0.000002     0.000001     0.000009
PLUMED: 4 Calculating (forward loop)                     100     5.652675     0.056527     0.040888     0.124369
PLUMED: 5 Applying (backward loop)                       100     6.162441     0.061624     0.043200     0.084362
PLUMED: 6 Update                                         100     0.001336     0.000013     0.000006     0.000191
````

This is why I am surprised not to see anything that would be called in the backwards loop in the graph.  `ActionWithVector::runAllTasks` is only called during the calculate loop.

What have I not understood?  Thanks

**Comment by gtribello** [comment]
OK, I managed to squeeze a bit more speed out of the new implementation. However, I still don't really understand why the threading version of the code is so slow.  

The new comparison is here:

further-speed-2

If you have any suggestions on what to do next @Iximiel or @GiovanniBussi I would like to hear them.  We could perhaps have a chat and investigate whether @Iximiel and I can work on this more closely together cos I am well and truly stuck!  

I don't really have time for a meeting today or tomorrow morning as I am teaching a summer school.  During tomorrow afternoon would work, though, if that also works for you.

**Comment by Iximiel** [comment]
I forgot to `export PLUMED_FORBID_CHAINS=yes` :cry:
On the bright side, we can confront the graphs with and without the chain now, on the not bright side I have to redo them because they refer to https://github.com/plumed/plumed2/commit/765f97f45648d2bc4be89ce72fad31ef4a8c9169

Same simulations as before:
sc_1threads_4000nat_100steps
sc_4threads_4000nat_100steps
sc_8threads_4000nat_100steps

And here there is everything else
[perf7e0f435_nochain.tar.gz](https://github.com/user-attachments/files/16373106/perf7e0f435_nochain.tar.gz)

**Comment by gtribello** [comment]
Thanks @Iximiel.  That makes so much more sense now.  Phew!

**Comment by Iximiel** [comment]
@gtribello 
>I don't really have time for a meeting today or tomorrow morning as I am teaching a summer school. During tomorrow afternoon would work, though, if that also works for you.

For me it is ok

**Comment by gtribello** [comment]
Hi @Iximiel 

I'm sorry to mess you about.  Can we discuss this next week, please?  I am not feeling very well, so I will go home after this summer school and rest.

**Comment by gtribello** [comment]
Hello @Iximiel and @GiovanniBussi 

I looked at this branch and think I made it much faster.  The results of my latest set of benchmarks are shown below:

plumed-store-speed

I am starting to think this is a viable way to proceed with developing this part of the code.

@Iximiel, it would be nice if you could confirm that you get similar results to the ones I show on my graph. I'm sorry that I keep asking you to run these benchmarks. I know they are not very interesting.  If you can also build the flame graphs, that would be really great.  Thanks.

Once that is done, I am not sure how to proceed.  A first step would be to run all the regtests with the PLUMED_FORBID_CHAINS=yes option enabled, work out which tests fail and then fix them so they work with this option.  It would then be useful to benchmark all the Steinhardt stuff and compare the master speed and the new implementation.

The question is, how much do you folks (@Iximiel and @GiovanniBussi) want to be involved in these efforts?  I know that we started this project with the intention of using the GPU to parallelise this part of the code.  I don't know when we begin this effort and bring @Iximiel into the project.  Do we (A) have me build a version of everything that works without the linked list and then GPUify the whole thing, or (B) try to GPUify the code while simultaneously removing the linked lists from the CPU code?    

Given our progress towards completing the draft of the tutorial paper, I think we should revisit this discussion next week. This timeline aligns with our current milestones. What are your thoughts on this?

**Comment by Iximiel** [comment]
I made a python package that post-processes the benchmark output and creates the graphs in 4-10 lines without touching awk and creating secondary files, so setting up and creating the bar chart is nearly frictionless
~While the benchmarks run on Leonardo~ Leonardo today is under maintenance, so we need to wait for the benchmarks. 
As of now I can only show you the flame graphs, that I created on my pc:

sc_1threads_4000nat_100steps
sc_4threads_4000nat_100steps
sc_8threads_4000nat_100steps


[perfe23214d_nochain.tar.gz](https://github.com/user-attachments/files/16441660/perfe23214d_nochain.tar.gz)

**Comment by GiovanniBussi** [comment]
@gtribello regarding this:

> The question is, how much do you folks (@Iximiel and @GiovanniBussi) want to be involved in these efforts? I know that we started this project with the intention of using the GPU to parallelise this part of the code. I don't know when we begin this effort and bring @Iximiel into the project. Do we (A) have me build a version of everything that works without the linked list and then GPUify the whole thing, or (B) try to GPUify the code while simultaneously removing the linked lists from the CPU code?

I am happy to be involved, though I don't know how much I can help directly on the code... But certainly we can have @Iximiel investing more time on it, possible organizing a visit (to Trieste or to Belfast). And for other future developments, I am interesting in trying to get these improvements also to other parts of the code.

Regarding the two options, I would suggest B. Just because we want to make sure there is an advantage before investing too much effort in it

**Comment by gtribello** [comment]
I have been doing some more experimentation with this way of doing things.  I am now benchmarking some calculations run using Steinhardt parameters.  These are the results I obtain for the following input:

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM 
PRINT ARG=q6_sum FILE=colvar 
BIASVALUE ARG=q6_sum
````

q6_speed

I have also done some work with the following input for the local Q6 parameter:

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0}
lq6: LOCAL_Q6 SPECIES=q6 SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM
PRINT ARG=lq6_sum FILE=colvar
BIASVALUE ARG=lq6_sum
````

with forces and 500 atoms I get the following timings:

|              | 1 thread  | 4 threads |
|--------| --------  | ---------   |
| New     | 2.374      | 7.335         |
| Master | 5.026      | 1.974.        |

The new version of the code crashes if I use more than 500 atoms.  Similarly, the new version of the code crashes if I use 8000 atoms in the Q6 calculation with derivatives.  I think I understand why.

I think that even though these results are not totally convincing, this direction is still worth pursuing.

**Comment by gtribello** [comment]
Hello @Iximiel and @GiovanniBussi 

I am finally in a position to discuss how we move forward with this tomorrow morning.  The latest timings for Coordination number, Q6 and LQ6 are shown below.  The bars show the times for 20 steps.  You can see that the new code performs much better than the current master.

latest-times

In the graph above, the colours are as follows:

Blue = Current master 1 thread
Orange = New version 1 thread
Green = Current master 4 threads
Red = New version 4 threads

A small problem is that the code crashes when you try to run calculations of Q6 for 8000 atoms. I think my laptop doesn't have enough memory to run calculations this large.  This problem can probably be fixed.  Before working too much on that, however, I would like to try to work out how to involve Daniele in this project and how we can start having some of this code run on GPUs. 

See you tomorrow.

--

For reference, input for Q6 is:

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM 
PRINT ARG=q6_sum FILE=colvar 
BIASVALUE ARG=q6_sum
````

and for LQ6

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} 
lq6: LOCAL_Q6 SPECIES=q6 SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM 
PRINT ARG=lq6_sum FILE=colvar 
BIASVALUE ARG=lq6_sum
````

**Comment by GiovanniBussi** [comment]
Thanks @gtribello 

I guess there's some wrong x-label, but if they all refer to something describing an increase of the number of atoms I think it's very promising. A few comments/questions:

- If I understand correctly, the diffference is how you handle backpropagation. The new way on handling it is both better for the number of operations (so it gets faster now) and easier to port on GPU.
- I am a bit confused by the timings. In the blue bar, are they expected to grow quadratically with N or linearly? If the correct x-label is the one in the top right figure I guess they are growing linearly, right? Do you change the scaling when you pass to the new implementation?
- It's interesting that the memory problem is only with 4 threads. Perhaps the way you allocate the buffers for OMP reduction can be modified?
- Again about the memory issue, would it be conceivable to rewrite this variable in blocks? Like, for coordination, it would be `GROUPA=1-10 GROUPB=11-20` then `GROUPA=11-20 GROUPB=1-10` then `GROUPA=1-10 GROUPB=11-12` then `GROUPA=11-20 GROUPB=11-20` and then take the sum? This could be even done automatically.

**Comment by gtribello** [comment]
I think the timings should grow linearly with the number of atoms because we are using a neighbour list.

**Comment by gtribello** [comment]
OK @GiovanniBussi and @Iximiel 

I've added some further code and the memory problem is now fixed.  The times including the 4 threads and 8000 atoms are as follows:

latest-times-2

File: src/core/ActionWithVector.cpp [hunk 1]
```diff
@@ -528,131 +273,77 @@ void ActionWithVector::getNumberOfTasks( unsigned& ntasks ) {
...
-  if( isActive() ) {
-    myvals.setTaskIndex(current); myvals.vector_call=true; performTask( current, myvals );
+  const ActionWithMatrix* am = dynamic_cast<const ActionWithMatrix*>(this);
+  myvals.setTaskIndex(current); myvals.vector_call=true; performTask( current, myvals );
+  for(unsigned i=0; i<getNumberOfComponents(); ++i) {
+    const Value* myval = getConstPntrToComponent(i);
+    if( am || myval->hasDerivatives() ) continue;
+    Value* myv = const_cast<Value*>( myval );
+    if( getName()=="RMSD_VECTOR" && myv->getRank()==2 ) continue;
+    myv->set( current, myvals.get( i ) );
```
Iximiel, line 286: This is a const function in which non-const things happen, it is called from non-const functions(runAllTasks and checkForForces), so it does not need the const modifier.
It may be easier to unflag this from its constness and change to code to something like, so there is not need to `const_cast` the values to modify them:
```suggestion
  myvals.setTaskIndex(current);
  myvals.vector_call=true;
  performTask( current, myvals );
  if (const ActionWithMatrix* am = dynamic_cast<const ActionWithMatrix*>(this);!am){
    const bool IamRMSD_VECTOR=getName()=="RMSD_VECTOR";
    for(unsigned i=0; i<getNumberOfComponents(); ++i) {
      if(Value* myval = getPntrToComponent(i);
          !myval->hasDerivatives() && 
          !( IamRMSD_VECTOR && myval->getRank()==2 )) {
        myval->set( current, myvals.get( i ) );
      }
     }
```

I also moved some conditions outside the loop and changed the scope of `am`

Maybe the discrimination for `RMSD_VECTOR` can be done in it? Or with an ad-hoc overloaded method `virtual bool skipIf(Value *)const{return true};`?

Then for the parallelization, we can manipulate MultiValue into a Data sitter for cpu/gpu adresses

gtribello:
This PR is superseded by #1251

PR #1095, opened 2024-07-09
Added sigterm to the signal handler for Benchmark

Iximiel:
##### Description

I added the handler for SIGTERM to the benchmark CL.
I tried with `kill -TERM pid` and it seems to work
I made the info message more pedantic.

@GiovanniBussi I added the handler correctly or should I add another function?
##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
The motivation is that on an HPC queue the signal sent on "time expired" usually is SIGTERM and as now the benchmark exit gracefully only with SIGINT

PR #1096, opened 2024-07-10
Small optimization in CoordinationBase

Iximiel:
##### Description

I changed how the couples are traversed in the CoordinationBase `calculate()` method, by passing from a "striped" traversal to a "block" one.
I get a 3%-5% speedup on my local machine.

---
Here are some results I got on my workstation, the graph show the time spent in `calculate()` relative to the time spent in calculate in the master branch

for 4 mpi processes and 2 threads each one
image

for 1 process with 8 threads (Here I did not expect performance improvement)
image

---

Before removing the draft status I'd like to see if it is possible to not store the couples in the neighbor list if the user asks for a fixed list

##### Target release

I would like my code to appear in release 2.10 (but I think this can rebased on 2.9)

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/colvar/CoordinationBase.cpp
+++ b/src/colvar/CoordinationBase.cpp
@@ -149,17 +149,21 @@ void CoordinationBase::calculate()
   const unsigned nn=nl->size();
   if(nt*stride*10>nn) nt=1;
 
+  const unsigned elementsPerRank = std::ceil(double(nn)/stride);
+  const unsigned int start= rank*elementsPerRank;
+  const unsigned int end = ((start + elementsPerRank)< nn)?(start + elementsPerRank): nn;
+
   #pragma omp parallel num_threads(nt)
   {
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
I think you get an improvement also with pure openmp because also in that case the access to continuous memory is faster

Iximiel:
> Before removing the draft status I'd like to see if it is possible to not store the couples in the neighbor list if the user asks for a fixed list

I think it may be a better idea to do that in a separate PR

If you think it may be better to try another benchmark I'll do it, otherwise, I think it can be merged/rebased on 2.9

GiovanniBussi:
I think we can merge it. It can even go to v2.9 because the change is minimal.

Iximiel:
Ok, then I'm rebasing this to the v2.9

Iximiel:
I changed the base and I pushed for the CI. while the CI is running I'm also comparing the branch with the master v2.9

PR #1102, opened 2024-07-25
set up a very simple way for visualizing the benchmark trajectories

Iximiel:
##### Description

I am writing a small piece for the school explaining how the benchmark can be used, and I needed to visualize the trajectories, these modifications were the simplest thing to obtain the trajectories.

`plumed benchmark --dump-trajectory trj.xyz +other options` dumps the trajectory that would be used in the benchmark to the "trj.xyz" file.
By default write an XYZ compatible with `plumed driver`, but with an environment variable set (`PLUMED_BENCH_EXTENDED_XYZ`) it can output an extended xyz file;

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/cltools/Benchmark.cpp
+++ b/src/cltools/Benchmark.cpp
@@ -456,6 +456,7 @@ void Benchmark::registerKeywords( Keywords& keys ) {
   keys.add("compulsory","--maxtime","-1","maximum number of seconds (-1 means forever)");
   keys.add("compulsory","--sleep","0","number of seconds of sleep, mimicking MD calculation");
   keys.add("compulsory","--atom-distribution","line","the kind of possible atomic displacement at each step");
+  keys.add("optional","--dump-trajectory","dump the trajectory to this file");
   keys.addFlag("--domain-decomposition",false,"simulate domain decomposition, implies --shuffle");
   keys.addFlag("--shuffled",false,"reshuffle atoms");
 }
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
I don't like very much the `getenv(PLUMED_BENCH_EXTENDED_XYZ)` ...

Can you make a list of software that can read this format?

If it's not super standard, I would remove this option. You can easily use awk to add those lines afterwards.

If it's super standard we can think about it also in other parts of the code. But for the purpose of visualizing the output of benchmark, I would just add the line a posteriori. Otherwise it's very confusing (driver will only write our "custom" xyz instead)

Iximiel:
The extended xyz is identical to the "standard" xyz but instead of ignoring the comment line, it uses it to define the box and/or to store extra information, like indicating what the columns are.
It is compatible with all readers like vmd that ignore the comment as per standard
If I remember well, it was originally proposed by quip, and the reader I use most, ovito, uses the extra data to render the box or to take information from eventual extra columns

By default, the XYZ files are in the format compatible with `plumed driver`

GiovanniBussi:
Yes I remember you mentioning ovito. I was wandering if there is other software supporting it, and if we should make the transition (keeping the possibility to read "old style" xyz with just cell on second line).

But this is a more general question. For this specific PR, I would remove this format from the plumed benchmark tool since we do not support it anywhere else. And you can easily adjust the file with awk if it's just for opening it with ovito

Iximiel:
> Yes I remember you mentioning ovito. I was wandering if there is other software supporting it, and if we should make the transition (keeping the possibility to read "old style" xyz with just cell on second line).
> 
> But this is a more general question. For this specific PR, I would remove this format from the plumed benchmark tool since we do not support it anywhere else. And you can easily adjust the file with awk if it's just for opening it with ovito

I rebased to the driver-xyz only

PR #1105, opened 2024-08-07
Plumed h for gromacs

Iximiel:
##### Description

While trying to add a plumed module to gromacs I got a few `-Wstringop-truncation` warnings
with this PR I am trying to address them in a way that does not imply too many changes in the original plumed codebase.

I am starting this as a draft because likely this branch will get some more commit

--- 

the changes in the tests are to make the compiler complain about the tested code and not the code of the test.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/wrapper/Plumed.h
+++ b/src/wrapper/Plumed.h
@@ -3831,7 +3831,7 @@ void* plumed_attempt_dlopen(const char*path,int mode) {
       __PLUMED_FPRINTF(stderr,"+++ Allocation error +++\n");
       __PLUMED_WRAPPER_STD abort();
     }
-    __PLUMED_WRAPPER_STD strcpy(pathcopy,path);
+    __PLUMED_WRAPPER_STD memcpy(pathcopy,path,strlenpath+1);
     pc=pathcopy+strlenpath-6;
     while(pc>=pathcopy && __PLUMED_WRAPPER_STD memcmp(pc,"Kernel",6)) pc--;
     if(pc>=pathcopy) {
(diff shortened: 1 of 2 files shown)


src/wrapper/Plumed.h
```diff
@@ -2346,7 +2346,7 @@ class Plumed {
     char msg[__PLUMED_WRAPPER_CXX_EXCEPTION_BUFFER];
     void init(const char* msg) __PLUMED_WRAPPER_CXX_NOEXCEPT {
       this->msg[0]='\0';
-      __PLUMED_WRAPPER_STD strncat(this->msg,msg,__PLUMED_WRAPPER_CXX_EXCEPTION_BUFFER-1);
+      __PLUMED_WRAPPER_STD strncpy(this->msg,msg,__PLUMED_WRAPPER_CXX_EXCEPTION_BUFFER);
```
Iximiel, line 2349: since this starts with `this->msg[0]='\0';`, I think that, only in this case, strncpy should work exactly like strncat

removing the "-1" ensures that the warning is gone

src/wrapper/Plumed.h
```diff
@@ -3831,7 +3831,7 @@ void* plumed_attempt_dlopen(const char*path,int mode) {
       __PLUMED_FPRINTF(stderr,"+++ Allocation error +++\n");
       __PLUMED_WRAPPER_STD abort();
     }
-    __PLUMED_WRAPPER_STD strncpy(pathcopy,path,strlenpath+1);
+    __PLUMED_WRAPPER_STD strcpy(pathcopy,path);
```
Iximiel, line 3834: The ideal would be strdup, but it is not ansi C (actually is c23, from my understanding)

the +1 on the length with strncpy activates the warning about an overflow

by reading the manual strcpy copies the whole string plus the `\0`.

carlocamilloni:
@Iximiel isn't this ready for merging?

Iximiel:
It should aim 2.10, that was 'master' at time of setting this up, I am changing it

PR #1106, opened 2024-08-13
Improvmements to documentation by adding types of value

gtribello:
##### Description

This commit changes the Keywords class so that the type of object (scalar/vector/matrix/grid) that is stored in each value is provided.  In addition, I have also modified the way that you specify ARG keywords so that you can specify what types of value are accepted in the input to the action.  The idea would be to incorporate all this information in the manual eventually.

It looks like a massive change but the only changes in most functions are in the registerKeywords function.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
HI @GiovanniBussi 

We discussed merging this in the meeting a couple of weeks ago.  Is it OK to merge it now?

Thanks

Issue #1091, opened 2024-06-28
Weird behavior of LOWER_WALLS


riccardocapelli:
Today I used a LOWER_WALLS potential on a distance and set the exponent of the function to 1. 

Looking at the bias, I got a value of 0 when d>threshold, but a negative value (i.e. an attractive potential) when d<threshold. 

Is this the expected behavior (and I should put a - in front of the KAPPA) or, as the name of LOWER_WALLS suggests, is this a bug?

Thanks a lot!

Riccardo

Iximiel:
I think the problem is here:
https://github.com/plumed/plumed2/blob/669793fc2818012bbd60c39882031e92ae5c5f4d/src/bias/Walls.cpp#L97

LOWER_WALLS now gets shorcut to a custom func `step(-x)*x^pow` so that if pow is *odd* `x^pow<0` when `x<0`.

whereas UPPER_WALLS to `step(x)*x^pow`, but UPPER_WALLS does not care about the sign of x in the exponential part


The solution should be to change the custom function to either:
 -`step(-x)*abs(x)^pow` 
 - `step(-x)*(-x)^pow`.

We need to test the faster between the two and PR it I think (and maybe add a test for this)

GiovanniBussi:
I agree that this is a bug, should be fixed as @Iximiel suggested (the two approaches should have the same measurable performance)

Iximiel:
I think @riccardocapelli just kicked a beehive 🐛 

#1109 may not be enough:

I'll use a "fancy nomenclature":
 - `UPPER_WALL` is "above", so it should "push down" when you get above it
 - `LOWER_WALL` is "below", so it should "push up" when you get under it

With @Puncist we started to set up the most simple test for this (walls on the x coordinate), and we discovered a few interesting things, already in 2.9:
- `UPPER_WALLS` works as intended, unless `EXP=1`
- `EXP=1` set the force to 1, always, no matter the distance from the wall, and always push "down", 
- `LOWER_WALLS` pushes "down" instead of "up"

<details><summary>Some examples in spoiler</summary>

We are running with this config
```bash
type=driver
arg="--plumed plumed.dat --ixyz trajectory.xyz  --debug-forces forces --dump-forces-fmt=%8.4f"


function plumed_regtest_before() {
    for x in {-5..5}; do
        echo "1" 
        echo "100 100 100"
        echo "C $x 0 0"
    done > trajectory.xyz    
}

function plumed_regtest_after() {   
[...]

PR #1109, opened 2024-09-09
FIX - Changing the sign of the argument of the exponential in  LOWER_WALLS

Iximiel:
##### Description

This should solve #1091, see the issue for the problem

I want to benchmark the  (-x) version against the (abs(x)) version, but maybe that is a thing for another PR.

Shall we add a new test to check the sign in this and maybe other COLVARs?

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/Walls.cpp
+++ b/src/bias/Walls.cpp
@@ -94,7 +94,7 @@ Walls::Walls(const ActionOptions&ao):
       readInputLine( getShortcutLabel() + "_pow_" + argn + ": CUSTOM PERIODIC=NO FUNC=step(x)*x^" + exp[i] + " ARG=" + getShortcutLabel() + "_scale_" + argn );
     } else if( getName()=="LOWER_WALLS" ) {
       readInputLine( getShortcutLabel() + "_scale_" + argn + ": CUSTOM PERIODIC=NO FUNC=(x-" + offset[i] +")/" + eps[i] + " ARG=" + getShortcutLabel() + "_cv_" + argn );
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
I think the abs(x) vs -x has no impact on performance. I would use the most robust one.

Something that instead might be relevant (if you have many of these biases) is that they construct a lot of actions. @gtribello is it necessary? Are you aware that you can define intermediate values in lepton?

A test for it is [here](https://github.com/plumed/plumed2/blob/669793fc2818012bbd60c39882031e92ae5c5f4d/regtest/basic/rt-lepton/plumed.dat#L7) though it is not very instructive. But you can do something like:

```
FUNC=7*a^n;a=(x-x0);x0=3;n=2
```

The only counterintuitive thing is that values are parsed from right to left, so that you first use quantities then define them.

The example above should do `7*(x-3)^2`

The advantage is that:
- there is a single lepton call (performance gain should be measurable if you have many of these)
- lepton simplifies the expression (should be irrelevant in this case)

carlocamilloni:
@Iximiel @GiovanniBussi @gtribello this needs to be retarget to 2.10, I would then merge it to fix the bug. The lepton change is about performances and can be done in a second moment

carlocamilloni:
@Iximiel there are two lines added in multicolvardensity, why is that so?

Iximiel:
> @Iximiel there are two lines added in multicolvardensity, why is that so?

I think that they are some things that comes from the various merges of master and v2.10
I am rebasing those changes away

PR #1111, opened 2024-09-16
FIX-2.9 Now the forces from LWALLS should be correct 

Iximiel:
##### Description

This is about #1091, now if a lower wall is asked with an odd exponent, it should give the correct force, and I think also the energy because ` ene += k * power;` now will contribute with a positive and repulsive term in any case.

I think we need some more tests to check this.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/bias/LWalls.cpp
+++ b/src/bias/LWalls.cpp
@@ -133,7 +133,7 @@ void LWalls::calculate() {
     if( lscale < 0.) {
       const double k=kappa[i];
       const double exponent=exp[i];
-      double power = std::pow( lscale, exponent );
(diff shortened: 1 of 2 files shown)


Iximiel:
> shouldn't we back port it as back as possible?

I agree

Iximiel:
The new commit changes the documentation: I noted that the offset for UPPER was added, and for LOWER was removed, so I slightly changed the documentation accordingly, this should also reflect that the wall is repulsive with the absolute value before the power, (it is correct to put the absolute value in the documentation like that?)

PR #1112, opened 2024-09-17
Proposing: compacting lower and upper walls into a single file

Iximiel:
##### Description

In the context of the WALLS, I compacted Uwalls and Lwalls into a single file (without the update of the documentation from #1111, but with the code correction)

The idea is to remove the repetition (there are only 17 different lines in two files of 158 and 157 lines, which should work in a mirrored way) and get to debug\update a single file.
Then I would like to port something similar also for the non-scalar version of the walls since I think it could bring advantages (no jit and compile-time optimizations and less cache misses).
And since the code itself would be the very same maybe compact also the non-scalar variant into a single one.

What do you think?

This is a `constexpr` solution, so to back-port it will lose some of it "slimness".

##### Target release

I would like my code to appear in release ???

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
@Iximiel ho fatto casino con il rebase :) comunque volevo provare a recuperarla

Iximiel:
@carlocamilloni I am giving a look into this in a few evenings!

PR #1114, opened 2024-09-19
Proposing: removing lines with more than one statements.

Iximiel:
##### Description

Don't merge this right now.
I am starting from the v2.9 since v2.10 bumped from C++11 to C++17 and so *I think* that v2.9 will be maintained a little longer.

I am changing the astyle settings to have 1 line 1 operation, this should make the code more diff friendly and more readable.
This should also make the coverage by line more accurate

Another thing that this PR will help with is some comments like this:
```C++
atom_lab.resize(0); unsigned nnodes; // Delete all the atom labels that have been created
```
that (I think) they are not as clear as they should be:
```C++
// Delete all the atom labels that have been created
atom_lab.resize(0);
unsigned nnodes;
```

When everything is clear I will squash everything in a single commit and remove the "legacy option" from astyle

##### Target release

I would like my code to appear in release v2.9++

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
I agree this way is more readible for both us and the the software tools

GiovanniBussi:
I agree. I see that sometime having no newline can make the code more compact. However:
- it does not show untested code as such in coverage scan
- it makes the code less readable when there are multiple instructions in the same line (e.g. `if(a) b=1; c=2;`).

So I agree in having a new line after an if or for statement. As long as we have a new line, I think braces should be there for clarity. With more and more people using python daily, we need to avoid ambiguous code such as:
```
if(a)
    b=1;
    c=2;
```

Iximiel:
The state is now far more strict than before:
The option that I did not manage to set up is to add brackets to consecutive nested branches/loops:
for example, this
```c++
for(int i=0; i<2; i++) for(int j=0; j<2; j++) for(int k=0; k<2; k++) shifts[i][j][k].clear();
```
became this
```c++
  for(int i=0; i<2; i++)
    for(int j=0; j<2; j++)
      for(int k=0; k<2; k++) {
        shifts[i][j][k].clear();
      }
```
but not this
```c++
  for(int i=0; i<2; i++) {
    for(int j=0; j<2; j++) {
      for(int k=0; k<2; k++) {
        shifts[i][j][k].clear();
      }
    }
  }
```

I do not think that astyle enforces this passage (at least, I am convinced about this because I did not find the option in the manual), but both versions pass the style check, and changing code manually is a `long long` manual work (that needs to be redone for 2.10, and master)

Still is clearer, and can be done atomically when someone edits that file and happens to remember that

Apart from that this is fine and ready
Then there is the merge to v2.10 (and to master) but with @GiovanniBussi we discussed how to do them 

we'll need to open an issue or something like that for external contributors that are developing on older branches with the series of instructions like:
```bash
git merge hash-2.9-before-format
git merge -s ours hash-2.9-after-format
make astyle
git merge v2.9
```

Iximiel:
> Another thing that this PR will help with is some comments like this:

This sentence is not true anymore, to do that I would need to check more or less every file,
And since the merge strategy will be something like:
```bash
merge hash-before-this-pr
merge -s ours hash-of-this-pr
# some other commands in the middle
git checkout hash-of-this-pr .astyle.options
make astyle
# finalize the commit
```
All the comment editing will need to be redone by hand for each misplaced comment :/

carlocamilloni:
shall we move on with this?

Iximiel:
I think so,
But to merge the modification of 2.9 into 2.10 and then into master we should follow #1157, and I'd prefer to open a PR for each version bump, to update the `.git-blame-ignore-revs` (for a clearer history change).

And I think I need to rebase this to the current 2.9 before

Iximiel:
@carlocamilloni 
I did the rebase on the 2.9.3

carlocamilloni:
then let's start..

Iximiel:
@carlocamilloni could you revert the merge?
this should be merged, not squashed+merged

carlocamilloni:
@Iximiel done, if you reopen it I can then merge it without squashing

PR #1116, opened 2024-10-03
Reuse result from getenv

mabraham:
##### Description

In theory, calling getenv a second time could return NULL after previously returning non-NULL, which would be undefined behavior when passed to strcmp. Re-using the non-NULL value is slightly safer, and keeps clang's static analyzer happier.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/wrapper/Plumed.h
+++ b/src/wrapper/Plumed.h
@@ -4015,6 +4015,8 @@ void plumed_retrieve_functions(plumed_plumedmain_function_holder* functions, plu
   void* p;
   char* debug;
   int dlopenmode;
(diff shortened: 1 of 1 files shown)


mabraham:
Found during GROMACS CI version upgrade. Upstreaming the change for future convenience.

GiovanniBussi:
I see, this would happen if:
- getenv is called
- context switch to another thread
- the other thread changes the environment variable
- context switch to the initial thread
- getenv is called again

Unlikely, but I agree should be fixed, thanks!

PR #1122, opened 2024-10-07
Handler class does not have a `watch` field

Luthaf:
##### Description

Remove extraneous field assignment. This breaks the build for me when trying to compile the master branch using https://github.com/plumed/conda/.

Since this was not caught in the 6 months since this code was introduced, maybe it is not used anywhere in the code base? I'm not sure why this is failing now.

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/tools/Tools.h
+++ b/src/tools/Tools.h
@@ -368,7 +368,7 @@ public:
           section=handler.section;
           key=std::move(handler.key);
         }
-        handler.watch=nullptr;
+        handler.section=nullptr;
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
Yes I think it's not used. Actually I think it should be:
```
handler.section=nullptr;
```
Can you confirm that this would compile correctly? Thanks!

Luthaf:
Yes, this builds correctly with `section`!

PR #1129, opened 2024-10-10
Do not hold Python's GIL when running commands

### LINKED ISSUE #1128: Consider releasing the GIL in the Python interface
The current Python interface to PLUMED uses cython to call the C++ code, and by default cython takes and hold the Global Interpreter Lock (GIL).

Unfortunately, this conflicts with the `metatensor` (and likely `pytorch`) modules, because PyTorch checks that the GIL is not held when running `backward()` (i.e. computing gradients): 

https://github.com/pytorch/pytorch/blob/f69bf005f7b3be70b9a38ba663e6bc825dfe7ffd/torch/csrc/autograd/python_engine.cpp#L130-L135

This means that trying to use PLUMED with a `METATENSOR` action from Python will always fail with the above error.

Would it be possible to annotate the functions in `python/cplumed.pxd` with `@cython.nogil`? (cf https://cython.readthedocs.io/en/latest/src/userguide/nogil.html)

This would fix the above problem, and should be fine to do as long a there is no function call from PLUMED back to the Python interpreter. I'm not sure if the latter is the case or not though (i.e. can someone define custom CV as pure Python scripts?)

Luthaf:
##### Description

Fixes #1128. I can confirm this fixes the issue for `metatensor` actions.

However, I'm not sure how to test that `pycv` still works fine, is it included in the CI test suite?

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/python/cplumed.pxd
+++ b/python/cplumed.pxd
@@ -31,7 +31,7 @@ cdef extern from "Plumed.h":
         size_t nelem
         size_t* shape
         size_t flags
-        # ignore other 
+        # ignore other
     ctypedef struct plumed_nothrow_handler:
         void* ptr
         void (*handler)(void*,int,const char*,const void*)
@@ -46,12 +46,12 @@ cdef extern from "Plumed.h":
         plumed_error_filesystem_path path1
         plumed_error_filesystem_path path2
         # ignore other members
-    void plumed_cmd_safe_nothrow(plumed p,const char*key,plumed_safeptr safe,plumed_nothrow_handler nothrow)
-    void plumed_error_set(void*ptr,int code,const char*what,const void* opt)
-    void plumed_error_init(plumed_error* error)
-    void plumed_error_finalize(plumed_error error)
-    plumed plumed_create()
-    plumed plumed_create_dlopen(const char*path)
-    plumed plumed_create_invalid()
-    void plumed_finalize(plumed p)
-    int plumed_valid(plumed p)
+    void plumed_cmd_safe_nothrow(plumed p,const char*key,plumed_safeptr safe,plumed_nothrow_handler nothrow) nogil
+    void plumed_error_set(void*ptr,int code,const char*what,const void* opt) nogil
+    void plumed_error_init(plumed_error* error) nogil
+    void plumed_error_finalize(plumed_error error) nogil
+    plumed plumed_create() nogil
+    plumed plumed_create_dlopen(const char*path) nogil
+    plumed plumed_create_invalid() nogil
+    void plumed_finalize(plumed p) nogil
+    int plumed_valid(plumed p) nogil
--- a/python/plumed.pyx
+++ b/python/plumed.pyx
@@ -36,6 +36,7 @@ import sys
 import warnings
 import types
 
+import cython
 from cython.operator import dereference
 
 if sys.version_info < (3,):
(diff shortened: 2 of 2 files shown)


Iximiel:
> However, I'm not sure how to test that `pycv` still works fine, is it included in the CI test suite?

pycv has its workflow run in `linux(-pycv-mpi-)`, which runs its regtests in its plugin subdirectory.

But we do not have a "run pycv from plumed in python" test that runs on the CI

GiovanniBussi:
> > However, I'm not sure how to test that `pycv` still works fine, is it included in the CI test suite?
> 
> pycv has its workflow run in `linux(-pycv-mpi-)`, which runs its regtests in its plugin subdirectory.
> 
> But we do not have a "run pycv from plumed in python" test that runs on the CI

This possibly means that, due to these GIL issues, also "run pycv from plumed in python" is not working? I though @Iximiel was testing this at least locally, is this correct? Otherwise I think we need a test

Iximiel:
> 
> This possibly means that, due to these GIL issues, also "run pycv from plumed in python" is not working? I though @Iximiel was testing this at least locally, is this correct? Otherwise I think we need a test

I tried, and I think I need to find a solution to this:

```
what():  Can't add new modules after the interpreter has been initialized
```

Luthaf:
Is this an error on the current master or with these changes?

Iximiel:
> Is this an error on the current master or with these changes?

The error is in master, don't worry.

I opened a separate issue about it

Luthaf:
@Iximiel and @GiovanniBussi, could one of you do a review of this? Thanks!

GiovanniBussi:
I think it's fine, I can merge it. For the records, @Luthaf could you reply to these questions? I would then add some documentation in the pyx file after merging to keep track:
- Why both the `nogil` attribute in the `pxd` file and the `with nogil():` in the `pyx` file?
- Other calls to C functions in the `pyx` file are not protected (e.g. `cplumed.plumed_error_init()`). I guess we cannot have a single `with nogil():` at the beginning of a block where we use python. However, would it be recommended to have a block around every call to a `cplumed.*` function? Which are pros and cons?
- Beside affecting the correctness in metatensor, is this change expected to affect performance in a positive or negative way?

Thanks!

Luthaf:
> Why both the nogil attribute in the pxd file and the with nogil(): in the pyx file?

The one we really want is the `with.nogil` in pyx, asking cython to release the GIL when running this function. The `nogil` in pxd are declarative, telling cython that these are pure C functions that don't require the GIL. They are required because otherwise cython refuses to run the `with.nogil` block.

> Other calls to C functions in the pyx file are not protected (e.g. cplumed.plumed_error_init()). I guess we cannot have a single with nogil(): at the beginning of a block where we use python. However, would it be recommended to have a block around every call to a cplumed.* function? Which are pros and cons?

So anytime the code is constructing Python objects, you can not have `nogil` around it, meaning you can not have a global `with nogil` around the whole thing. 

It could make sense to put a `with nogil` block around all calls to `cplumed` functions, this would allow for more complex calculations to happen in the other functions. It might also help for pycv (and in general any time a function call ends up back inside a Python interpreter) by preventing a deadlock where both py-plumed and the interpreter trying to get the same lock.

I only wrapped `cmd` since that's seems like the only function that would be likely to end back inside a Python interpreter as far as I understand the API.

> Beside affecting the correctness in metatensor, is this change expected to affect performance in a positive or negative way?

No,  I don't think so. This would potentially allow to have multiple Python threads call into plumed concurrently, but I don't even know if that's a supported use case anyway, and doing actual multi threading in Python is very painful.

PR #1145, opened 2024-10-20
Added new method for links to switching function documentation

gtribello:
##### Description

This commit addresses something that has bothered me for a while; namely, how we should provide the links to the parts of the manual on swiching function when you encounter a keyword like switch.  I would ideally like to provide the link to this page both in the tooltip that is generated for the annotated inputs and in the tables that contain the syntax.

This commit addresses this problem.  You can now use the command keys.linkActionInDocs when you register the keyword in registerKeywords.  Links to the manual pages for relevant actions will then be included in the manual and in tooltips.

As part of this I have also moved the documentation for switching function to the action LESS_THAN and the documentation for HistogramBead to BETWEEN as this is where these classes are used in their most basic form.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
If you want to see an example of what this would look like see:

https://gtribello.github.io/plumed-manual/new-manual/CONTACT_MATRIX.html

You can see the links in the tooltips in the example input here and how the links are also added to the syntax pages.

gtribello:
Hi @Iximiel @GiovanniBussi and @carlocamilloni 

I am trying to two PRs that I wrote for the work on updating the manual in contention again.  This is the easier one of the two.  This is basically for documenting switching functions.  As you can see here:

https://gtribello.github.io/plumed-manual/new-manual/COORDINATION.html

If you scroll right down to the bottom of the page, you can see that in the documentation for SWITCH there is a link to the manual page for the action LESS_THAN. Basically, the change here provides a mechanism for putting the requirement to create that link in the json file.  When I then build the manual, the link is created.

carlocamilloni:
[tools/Keywords.cpp:249] (performance) :passedByValue: Function parameter 'a' should be passed by const reference.

GiovanniBussi:
Hi @gtribello I think it's fine. I only have a small objection: the documentation of switching function is now farther from the code where new switching functions are added. I think we should have clearly written in the SwitchingFunction.cpp file something like "If you add a new switching function please document it in this other file".

gtribello:
OK I think that should be those two issues resolved now with that last commit.

PR #1148, opened 2024-10-29
Fix for output of ADAPTIVE_PATH

gtribello:
This change ensures that when you output the newly fitted path that you get by running the ADAPTIVE_PATH method the output pdb file contains the same residue numbers, beta and occupancy values as the input path.  The change to mapping/rt-adapt-alt-rmsd comes about becuase the residue numbers in the output are now correctly coppied from the input

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
This is the PR I just submitted but based on 2.10

PR #1158, opened 2024-11-19
Macro for fasterMulticolvar+named modes

Iximiel:
@gtribello 
I am starting to do some work to bring `Multicolvar.h` to be more compatible with the GPU

This first commit just aims to clearer the "mode" variable, and add a small layer of automatism to setting up a colvar to be compatible with a Multicolvar with a few macros. I think this could make it easier to change the function signatures faster. 

For running on a GPU I may need to restart implementing #1092

### DIFF
--- a/regtest/crystdistrib/rt-dops-shortcut/config
+++ b/regtest/crystdistrib/rt-dops-shortcut/config
@@ -1,4 +1,4 @@
 type=driver
-plumed_modules=crystdistrib
+plumed_modules="crystdistrib adjmat"
 arg="--plumed plumed.dat --mf_xtc em.xtc --mc mcfile --timestep=0.001 --trajectory-stride=25"
 extra_files="../../trajectories/em.xtc"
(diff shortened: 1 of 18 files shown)


Iximiel:
Update:
I simplified the `calculateCV` interface for the output part, this will likely make it easier to manipulate how that will work with managing the output data without changing the function signature.

I also want to do something similar for the input variables, but I have to deal with the fact that some CVs, like the dipole one, use some of the input values as temporary values, so the signature changes with some input not set as const ref but as simple refs

Iximiel:
Update: I set up also an Input helper with a test-guide.

I am working on changing the action atominstic ptr to something similar to the input/output classes
I was thinking on setting up something similar to the `mode` variable
The aim of this is to set up an object simple -and light- enough to be loaded on the GPU without excessive complications

Iximiel:
@gtribello can you look at the `Quaternion::calculateCV` method?

in the `//convert to quaternion` section`, the one with all the "ifthenelseif" blocks in the first block `S` is a `double` in the others is a `float`, since everything else is double precision I fear that it is a copy-paste error

Iximiel:
this have been superseeded by the various ParallelTaskManager PRs

PR #1160, opened 2024-11-26
Depracated action tracking

gtribello:
##### Description

This PR adds a facility where you can note that an action should no longer be used by adding the command:

```c++
setDeprecated("LESS_THAN");
```

In the `registerKeywords` method of an action.  The deprecated actions are then tracked in the syntax.json file.  By putting this information in that file this way we can note eggs and lessons that are using actions that we have moved on from at some stage in the future.

The first commit in this sequence fixes a regtest that appears to be broken on master.  I am not sure where this regtest came from.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
@gtribello this needs to be rebased to merge it

PR #1162, opened 2024-12-02
V2.10 patches/gromacs2025

Iximiel:
##### Description

@carlocamilloni @GiovanniBussi I am setting up the patch for gmx 2025.

Due to the gromacs implementation that I set up, the patch is quite lighter.
Due to the gromacs implementation that I set up, the patch lacks a few features you are used to.

As of now, I am simply adding the multisim capabilities that are not present in the "vanilla" plumed-in-gromacs implementation.

**Since gmx2025 is not out yet I am targeting the branch `release-2025` and in particular the commit `78e37c9cca4d87cfcfc8b9919ce044d54a29e6eb`.**

I added a new patch "mode" that does not copy the extra headers and configuration bits, but that might be an overshoot

##### Target release

I am targeting 2.10, is that ok?

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/patches/gromacs-2025.0.config
+++ b/patches/gromacs-2025.0.config
@@ -1,0 +1,39 @@
+
+
+function plumed_preliminary_test(){
+# check if the README contains the word GROMACS and if gromacs has been already configured
+  grep -q GROMACS README 1>/dev/null 2>/dev/null
+}
+
+function plumed_patch_info(){
+cat << EOF
+A basic PLUMED interface is already present in GROMACS
+
+This patch previes extra features to be implemented in the official PLUMED integration in the next GROMACS version
+
+Patching must be done in the gromacs root directory  _before_ the cmake command is invoked.
+
+The GROMACS integration has some improvements with the compatibility wiht the multi-threading implementation
+but it is still preferable to configure gromacs as
+
+cmake -DGMX_THREAD_MPI=OFF and add -DGMX_MPI=ON if you want to use MPI.
+
+To enable PLUMED in a gromacs simulation one should use
+mdrun with an extra -plumed flag. The flag can be used to
+specify the name of the PLUMED input file, e.g.:
+
+gmx mdrun -plumed plumed.dat
+
+For more information on gromacs you should visit http://www.gromacs.org
+
+EOF
+}
+
+plumed_before_patch(){
+  plumed_patch_info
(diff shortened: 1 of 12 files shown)


Iximiel:
for not losing track of what I am building:
[look here](https://gitlab.com/iximiel/gromacs/-/compare/release-2025...plumedpatch2025?from_project_id=17679574) to see the diff directly applied on the 2025 release.

Here I am adding a way to access plumed outside the mdmodules along with some feature that I am trying to implement into the 2026, and with the replica exchange (that currently does not work at full correctness)

Iximiel:
@carlocamilloni I changed the patch to be simpler than the previous one:
there is only the multisim implementation now.

Now is actually targeting 2025.0 and need some testing on the multisim part, but, at least on my PC, it compiles

This is a starting point (and should be aligned with its [MR](https://gitlab.com/gromacs/gromacs/-/merge_requests/4622) on the gromacs repo)
The more complete patch still exist here https://github.com/Iximiel/plumed2/tree/v2.10_patches/gromacs2025-advanced, it needs a rebase on this one and some tests

carlocamilloni:
@Iximiel @GiovanniBussi i am merging but I need a few more days to finalize the testing, please wait for that before releasing v2.10

carlocamilloni:
@Iximiel @GiovanniBussi I confirm it works correctly in my tests (multiple walkers via MPI + metainference averaging over replicas). At this point I would release v2.10!

carlocamilloni:
@GiovanniBussi in my case it fails because it complains about the property of the systems being all the same (this is a check we turned off with our previous patches), I am pretty sure it will works with incorrect results otherwise

GiovanniBussi:
@Iximiel should this be fixed in some way? Is there a way to detect from the plugin if one is running -replex?

This is an extremely likely mistake, so I would try to hard code a failure

carlocamilloni:
@Iximiel @GiovanniBussi we would need somethign more speciifc, the issue arises only if plumed adds a bias, otherwise its fine

Iximiel:
:thinking: so I should interrogate Plumed to check if there is a bias.
Do we have a command for asking that in the API?

Iximiel:
I added the check, I opened a PR #1233 to have the discussion there

PR #1165, opened 2024-12-17
fixed deprecation warning from pandas

Iximiel:
##### Description

I am working on some (small) modifications on the python package, while running the tests I found that the python module uses `delim_whitespace` that is deprecated from `pandas2.20`, (see [the manual](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html))

I used the suggested ` sep="\s+" ` option, which should be backward-compatible also with older pandas versions

##### Target release

I would like my code to appear in release 2.11 (or before?)

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/python/plumed.pyx
+++ b/python/plumed.pyx
@@ -551,7 +551,7 @@ def read_as_pandas(file_or_path,enable_constants=True,enable_conversion=True,ker
 
 # read the rest of the file
 # notice that if chunksize was provided the result will be an iterable object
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
@Iximiel could you perhaps backport this to v2.9? If it's a sort of LTS, I guess we should fix these small issues as they appear, at least when it's easy (as it is now)

Issue #1167, opened 2024-12-18
[Bug] Error shadows checkArgumentType warning


Iximiel:
The CI is failing on `/regtest/secondarystructure/rt-coverage-antibetarmsd`, see #1166
I found an error that may not be linked directly to that test failure:

The Error text is quite cryptic: 
> [!CAUTION] 
> WARNING: type for input argument has not been specified

Because it does not specify the Action that generates the problem

The "warning" that raises the error is raised here:
```c++
/*** from tools/Keywords.cpp ***/
bool Keywords::checkArgumentType( const std::size_t& rank, const bool& hasderiv ) const {
  for(auto const& x : argument_types ) {
    if( rank==0 && x.second.find("scalar")!=std::string::npos ) return true;
    if( hasderiv && x.second.find("grid")!=std::string::npos ) return true;
    if( rank==1 && x.second.find("vector")!=std::string::npos ) return true;
    if( rank==2 && x.second.find("matrix")!=std::string::npos ) return true;
  }
  plumed_merror("WARNING: type for input argument has not been specified");
  return false;
}
```

and shadows the original intended WARNING message, which reports the name of the action that generates the WARNING:

```c++
/*** from core/ActionWithArguments.cpp ***/
void ActionWithArguments::interpretArgumentList(const std::vector<std::string>& c, const ActionSet& as, Action* readact, std::vector<Value*>&arg) {
  /*a lot of other code*/
  for(unsigned i=0; i<arg.size(); ++i) {
    if( !readact->keywords.checkArgumentType( arg[i]->getRank(), arg[i]->hasDerivatives() ) ){
     readact->warning("documentation for input type is not provided in " + readact->getName() );
    }
  }
}
```

I think the intended use was to transform the warning into an error
and that `/regtest/secondarystructure/rt-coverage-antibetarmsd` 

I'm opening a PR with the  `->warning` changed to `->error` as soon I verify some other things

PR #1166, opened 2024-12-18
[FIX][TEST] rewrote /regtest/secondarystructure/rt-coverage-antibetarmsd for master

Iximiel:
##### Description

I removed the `*` from the dumpforces in the "title" test, since it give some problems with checkArgumentType().

Here, `DUMPFORCES` tries to dump the forces of ALL of the 46 previous actions (spawned by the shortcuts). Maybe it needs some push in the wanted direction

I obtained the following error message by commenting `// plumed_merror("WARNING: type for input argument has not been specified");` in  Keywords::checkArgumentType):
```
PLUMED: Action DUMPFORCES
PLUMED:   with label @47
PLUMED:   with stride 1
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
PLUMED: WARNING for action DUMPFORCES with label @47 : documentation for input type is not provided in DUMPFORCES 
[...]

PR #1168, opened 2024-12-18
[FIX] Moving the checkArgumentType error to its caller

Iximiel:
##### Description

This is the PR about the problem I saw in #1167 and #1166

I think that the intended use of the original change I am fixing was to change the warning into an error, since I saw that it generates SEGFAULTs in case of keeping it as 
 a warning.

But the `plumed_merror` in the `Keywords` scope lose the information of having the information of the "culprit action" (and I think that `plumed_merror` is deprecated in favour of `plumed_error`, right?). So I simply changed the `->warning` into an `->error` to have the name of the action that arises the problem

I think this is more a "dev-side" problem than a "user-input" one, but  it is still better to state to the user what is the action that may cause the error, since it can be "workarounded" as I did in #1166

`checkArgumentType` is only used in a single instance, so I think the modification is safe:
```
>git grep -n checkArgumentType
src/core/ActionWithArguments.cpp:195:    if( !readact->keywords.checkArgumentType( arg[i]->getRank(), arg[i]->hasDerivatives() ) ) {
src/tools/Keywords.cpp:693:bool Keywords::checkArgumentType( const std::size_t& rank, const bool& hasderiv ) const {
src/tools/Keywords.h:191:  bool checkArgumentType( const std::size_t& rank, const bool& hasderiv ) const ;
```
##### Target release

I would like my code to appear in release I ma targetting master but maybe v2.10 may be better?

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/ActionWithArguments.cpp
+++ b/src/core/ActionWithArguments.cpp
@@ -243,7 +243,7 @@ void ActionWithArguments::interpretArgumentList(const std::vector<std::string>& 
   }
   for(unsigned i=0; i<arg.size(); ++i) {
     if( !readact->keywords.checkArgumentType( arg[i]->getRank(), arg[i]->hasDerivatives() ) ) {
-      readact->warning("documentation for input type is not provided in " + readact->getName() );
+      readact->error("documentation for input type is not provided in " + readact->getName() );
     }
   }
 }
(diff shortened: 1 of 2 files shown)


carlocamilloni:
@Iximiel is this still valid? In case if you rebase I'll merge it

Iximiel:
Let's see if now passes the CI

PR #1176, opened 2025-01-16
Now cregex "appears" to be included (for backward compatibility)

Iximiel:
##### Description

> [!Note]
>rt-coverage-antibetarmsd fails due to what is being solved here #1166 

As what's told in #1175 I'm pushing a couple of commit that should solve the backward compatibility with "absence" of `cregex`.

In the first two I declare that cregex is there in config.txt (`plumed config has cregex`) and in `PLMD::config::hasCregex()`

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/regtest/basic/rt-cregex-backward-compatibility/Makefile
+++ b/regtest/basic/rt-cregex-backward-compatibility/Makefile
@@ -1,0 +1,1 @@
+include ../../scripts/test.make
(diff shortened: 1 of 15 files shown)


Iximiel:
I rebased on the current v2.10 because #1166 is now merged, so the tests _should_ pass

I also added a test to check the "phantom" cregex  with a comment that explain why it is there (and ran astyle because I forgot that...)

carlocamilloni:
I am merging it and then we can reduce the numerical precision of that single failing test

PR #1188, opened 2025-01-28
adding the "nobadge" badge to gen_example

Iximiel:
##### Description

>afterthought:
>The aim of this was to cover some segmentation fault.
> This does not solve those segmentation faults
> But I think that that is due some problems that cam from what happen after `error` is used. I just solved the no badge problem in some part of the manual, but some pages like the one about the `...` now it should render correcly, or at leas running an example with standalone things when you do not care about the status badge

 In #1183 I randomly discovered a sigsegv when running gen_example on plumed tools.

It looked like the "nobadge" status was missing from the options when creating a badge, now it displays a  in that case.

`"nobadge"` looks like it is the default status when `gen_example` is called with no arguments, so I do not know it this is the correct solution (but

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/cltools/GenExample.cpp
+++ b/src/cltools/GenExample.cpp
@@ -174,6 +174,8 @@ int GenExample::main(FILE* in, FILE*out,Communicator& pc) {
     ofile<<"with-LOAD-yellow";
   } else if(status=="incomplete") {
     ofile<<version<<"-incomplete-yellow";
+  } else if ( status=="nobadge" ) {
+    ofile<<"nobadge-36454F";
   } else {
     error("unknown status");
   }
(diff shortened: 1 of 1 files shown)


carlocamilloni:
@Iximiel lo farei su 2.10 no? a parte questo e' pronta?

Iximiel:
@carlocamilloni , it slipped on master, sorry.
Originally was aimed at 2.10, 


it solves the missing example in the continuation lines, but those are now rendered in an very wrong way....

Iximiel:
@carlocamilloni, I have no idea why (maybe I should have checked AFTER the process had finished...) but now the user-manual pages look like

PR #1192, opened 2025-02-05
adding a faster way to run cppcheck locally

Iximiel:
##### Description

Lately I am producing a fer error in CI/codecheck
So I needed to run cppcheck (and plumed check) locally.
But the plain script is a bit of "heavy", timewise, to be ran locally, and the error I cause are hidden among the warnings that have been accumulated before.

This reduces the noise in the terminal and makes the time spent runnig the checks from ~40 minutes to 1-2

So I stole this idea from another repository (that was used for the style formatting): by running cppcheck only on the files that differ from a given base branch/commit

The "help" text is this
```
localcppcheck.sh Runs cppcheck ONLY on the files in src that have changed from the selected base branch
By default it uses "master" as the base branch, but you can change it with --base=REV:

usage: localcppcheck.sh [--base=REV] change the base branch/commit to use as base for the diff
       localcppcheck.sh -h|--help   print this help

for example `localcppcheck.sh --base=upstream/master`
for example `localcppcheck.sh --base=380bb46eea1bc2a38fba94f3f72af2d8d6e41669`

you can also export the variable PLMD_MYCCPCHECK_REV to change the base branch/commit
the option --base has precedence over PLMD_MYCCPCHECK_REV
```

I think it can be useful for you.

I think I need to change a few things:
- [ ] the place for the scritpt
- [ ] Should I add a plumedcheck call here?
- [ ] I should advertise it in the developer manual

What do you think?

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
Now cppcheck/codecheck/plumedcheck is integrated in the makefile
When merged it need to be modified for be compatible with the 2.10 updates (`small_vector` and the extra discard for `grep -v :duplInheritedMember:`)

src/maketools/plumedcheck
```diff
@@ -88,7 +88,7 @@ BEGIN{
       case /--autoconf=.*/:
         sub("^--autoconf=","",opt);
         autoconf=opt;
-      case /-.*/:
+      case /^-.*/:
```
Iximiel, line 91: This modification is here because when passed a full path that contains an `-` that will be interpreted as an option

Iximiel:
The local check now passes full paths and works "from any directory", so that it can test also the files in the plugins directory

It should not be too difficult to change how the list of files is set up to go back to a relative path (assuming that it is launched in `src/`)

carlocamilloni:
no, should i merge it?

Iximiel:
I think It should be better to do the merge manually to avoid the override of the `small_vector` part
I'm opening a PR for doing that, if you are ok

Iximiel:
done here:#1195

PR #1115, opened 2024-09-27
Add DOI information to the json syntax file

gtribello:
##### Description

I have added a command so you can put the DOIs of papers that describe the functionality in an action in the syntax.json file.  To add a doi to your paperyou simply use the command:

```c++
keys.addDOI("<DOI>");
```

in the registerKeywords method of your action.  Hopefully this will help users quickly build documentation for their methods when we get the new manual working

##### Target release

I would like my code to appear in release 2.11

##### Type of contribution

- [x] changes to code or doc authored by PLUMED developers, or additions of code in the core or within the default modules
- [ ] changes to a module not authored by you
- [ ] new module contribution or edit of a module authored by you

##### Copyright

- [x] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

- [ ] the module I added or modified contains a `COPYRIGHT` file with the correct license information. Code should be released under an open source license. I also used the command `cd src && ./header.sh mymodulename` in order to make sure the headers of the module are correct. 

##### Tests

- [ ] I added a new regtest or modified an existing regtest to validate my changes.
- [x] I verified that all regtests are passed successfully on [GitHub Actions](https://github.com/plumed/plumed2/actions).

PR #1199, opened 2025-02-17
Citation management with DOI stuff

gtribello:
##### Description

This is a rework of PR #1115.  It allows you to put the DOIs of papers that describe the functionality in an action in the syntax.json file. To add a DOI to your paper, you simply use the command:

```c++
keys.addDOI("<DOI>");
```

in the registerKeywords method of your action. Hopefully, this will help users quickly build documentation for their methods when we get the new manual working.

Now, however, you can also cite a DOI in the PLUMED log file as well using:

```c++
log<<plumed.cite("<DOI>")<<"\n";
```

This works because I reuse Max's magic python script to create a std::map from all the DOIs that have been included in the json file.  I then read the map in the Citation Manager object.  A few observations  

1. In making this change, I realised that the citations in the log do not play nice with shortcut actions.  This was my fault and due to a misunderstanding of how the citations work.  Do you think I should do a bug fix PR on 2.10 to resolve the problem there @GiovanniBussi?  
2. Should we change everywhere in the code where someone has used an old style cite command so that we use the DOIs of the paper instead?
3. How should I ensure that the python script is used to keep the map of DOIs and up to date with what is in the syntax file?  Currently, I just assume that folks will run the script after they have added a DOI and commit the changes to the file that holds the map.  If they don't the DOI will be referenced in the log.  

On a completely separate note, what is the point of asking on this form: "I verified that all regtests are passed successfully on GitHub actions."?  If they don't pass all the tests, you see it below the PR.  Is it time to retire that bit of this form?

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/PlumedMain.cpp
+++ b/src/core/PlumedMain.cpp
@@ -1067,7 +1067,11 @@ void PlumedMain::readInputLine(const std::string & str, const bool& before_init)
     return;
   }
   std::vector<std::string> words=Tools::getWords(str);
-  citations.clear();
+  if( before_init ) {
+    plumed_assert( citations.empty() );
+  } else {
+    citations.clear();
+  }
   readInputWords(words,before_init);
   if(!citations.empty()) {
     log<<"Relevant bibliography:\n";
@@ -1729,7 +1733,7 @@ bool PlumedMain::DeprecatedAtoms::usingNaturalUnits() const {
 }
 
 void PlumedMain::DeprecatedAtoms::setCollectEnergy(bool b) const {
-  plumed.readInputLine( plumed.MDEngine + "_energy: ENERGY" );
+  plumed.readInputWords( Tools::getWords(plumed.MDEngine + "_energy: ENERGY"), false );
   plumed.setEnergyValue( plumed.MDEngine + "_energy" );
 }
 
--- a/src/tools/Citations.cpp
+++ b/src/tools/Citations.cpp
@@ -26,14 +26,23 @@
 
 namespace PLMD {
 
+const static Tools::FastStringUnorderedMap<std::string> doi_map = {
+#include "CitationMap.inc"
+};
+
 std::string Citations::cite(const std::string & item) {
+  std::string myref=item;
+  if( doi_map.find(item)!=doi_map.end() ) {
+    myref=doi_map.find(item)->second;
+  }
+
   unsigned i;
   for(i=0; i<items.size(); ++i)
-    if(items[i]==item) {
+    if(items[i]==myref) {
       break;
     }
   if(i==items.size()) {
-    items.push_back(item);
+    items.push_back(myref);
   }
   plumed_assert(i<items.size());
   std::string ret;
--- a/json/doi_manager.py
+++ b/json/doi_manager.py
@@ -1,0 +1,51 @@
+import subprocess
+import json
+
+def get_dois_from_syntax() :
+  with open("syntax.json") as f :
+    try:
+       plumed_syntax = json.load(f)
+    except ValueError as ve:
+       raise InvalidJSONError(ve)
+
+  doilist = []
+  for key, value in plumed_syntax.items() :
+      if not isinstance(value,dict) or "dois" not in value.keys() : continue
+      for doi in value["dois"] : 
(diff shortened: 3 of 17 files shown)


src/tools/Citations.cpp
```diff
@@ -26,14 +26,23 @@
 
 namespace PLMD {
 
+const static Tools::FastStringUnorderedMap<std::string> doi_map = {
```
GiovanniBussi, line 29: @gtribello can you double check if makefile automatically detects this as a dependency? I mean, in this directory:
- `make obj`
- `touch Citations.cpp`
- `make obj` # here Citations.o should be recompiled

GiovanniBussi: (if not we can add it manually to the Makefile)

src/core/PlumedMain.cpp
```diff
@@ -1067,7 +1067,11 @@ void PlumedMain::readInputLine(const std::string & str, const bool& before_init)
     return;
   }
   std::vector<std::string> words=Tools::getWords(str);
-  citations.clear();
```
GiovanniBussi, line 1070: @gtribello what's the reason for this change? Is it related to the PR?

src/tools/Citations.cpp
gtribello: I tried this and observed the following:

````
(plumed-cf) garethtribello@Gareths-MacBook-Pro tools % make obj                
compiling  BitmaskEnum.cpp
compiling  Citations.cpp
compiling  DLLoader.cpp
compiling  FileBase.cpp
compiling  HistogramBead.cpp
compiling  IFile.cpp
compiling  Keywords.cpp
compiling  LeptonCall.cpp
compiling  LinkCells.cpp
compiling  OFile.cpp
compiling  PDB.cpp
compiling  PlumedHandle.cpp
compiling  SwitchingFunction.cpp
compiling  Tree.cpp
(plumed-cf) garethtribello@Gareths-MacBook-Pro tools % touch Citations.cpp     
(plumed-cf) garethtribello@Gareths-MacBook-Pro tools % make obj           
compiling  Citations.cpp
````

I assume that means we are all good?

**Comment by GiovanniBussi** [comment]
> Which is the problem? If citations are incorrect in v2.10 anyway I would solve the problem there. Is this related to changing from readInputLine to readInputWords?

Maybe I got it.

readInputFile reads a full file calling multiple times readInputWords

readInputLines create a temporary file with all the passed lines and calls readInputFile

readInputLine (quite counterintuitively) reads a single line and flushes the bibliography

The original reason for this was to allow proper bibliography when reading input file one line at a time from the MD code. But I understand it's not clear.

So yes, in shortcuts you should normally use readInputWords (which does NOT flush the bibliography)

File: src/core/PlumedMain.cpp (hunk 2 above)
Comment (gtribello) [reply src/core/PlumedMain.cpp:1070]: I replaced that line with the following:

```c++
if( before_init ) {
    plumed_assert( citations.empty() );
  } else {
    citations.clear();
  }
```

So, the citations.clear is still done in most cases.  I am just not clearing the citations if plumed.init has not been called before the call to readInputLine.  Note readInputLine is used by actions that pass data from the MD code to PLUMED now.  

In other words, I am doing a sanity check here instead to make sure that people are not trying to cite anything.

gtribello: Not trying to cite anything in the interface.  We could change it to call readInputWords instead.  Basically, the problem with shortcuts is that I am calling readInputLine, which clears the citations.  This is the source of the mess.  I added that sanity check in the hope that would prevent folks from making a mess again in the future.

gtribello:
Yes you got it.  I will split the PR in two tomorrow.  I do a PR on 2.10 that fixes the issues with the shortcuts and the bibliography, and then I will move the stuff for the DOIs to a third PR on master.

Plumed::cite and Action::cite are identical Action::cite is a call to Plumed::cite.  I changed all the Action::cite commands into Plumed::cite commands in the first PR as it is easier to do a grep for plumed.cite than cite.  If we were going to change everything, I would have been able to find them all then.  I think that is unnecessary, though, so I will throw away that commit.

PR #1207, opened 2025-02-21
New manual version

gtribello:
##### Description

This is the first step towards fixing the PLUMED manual.  I have completely rewritten the build system for the manual and have also reorganised the documentation.  I think this will make the manual easier to maintain.  I also think this new manual fits better with the documentation efforts that we have done with plumed tutorials and plumed nest.

I have made a video that gives you a brief tour of the new manual [here](https://youtu.be/GZy70mhx1Hw).

Ideally, I would like to get this merged in the next couple of weeks.  There are two things that need to be fixed before we can merge this.  These are:

1. We need to stop cppcheck throwing an error if it doesn't find a `\par Examples` statement in the documentation, as that is not needed anymore.
2. I would also stop cppcheck throwing an error if actions are registered and there is no PLUMEDOC section.  The problem is that there are now some actions (e.g. CUSTOM_VECTOR) that are only ever created from shortcuts.  These actions do not need documentation sections.  Furthermore, the report page for the new manual shows you where the problems are.

In the longer term, we need to convert all the current documentation from Doxygen to GitHub markdown.  I have done a lot of this already for a lot of the code that I wrote.  I actually rewrote the documentation for these actions because, although the syntax is mostly the same, the way I would recommend using the code changed significantly since we made the updates for plumed 2.10.  My next step would thus be to merge all the new documentation I have written so far and put that in a second PR to merge after this one.  I would then rework the rest of the code documentation once those new documentation items are merged.

The thing I want to avoid is having to keep this branch of the code up to date with master for a long time.  If the decision is made to change the manual I think we can do a few short development spints with PRs at the end of them to get everything updated before the new manual is released in 2.11.  

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hi @GiovanniBussi, @carlocamilloni and @maxbonomi 

I think this is ready to go now.  I added in the stuff that we talked about.  I am not sure if the manual pushes anywhere when it is on a branch like this.  I think I got it to build on GH actions, though.

Could we merge into master and see what happens next?

GiovanniBussi:
Yes! I think it does not push when you are on a branch without a proper name, because it does not know where to push...

gtribello:
Hi @GiovanniBussi 

Should we merge this then?

GiovanniBussi:
Yes I think you can merge on master!

PR #1206, opened 2025-02-20
OpenACC (GPU) for MulticolvarTemplate

Iximiel:
I am opening this PR to have a clearer diff and to make a few checks

In particular I need to understand why some tests do not pass anymore the forces part. I did not touch that part yet... (could be that skipping of the pbc passage 🤔)
So I need a clearer diff.

At the time of writing this the `/targetGPU/rt-matrixsum` that is `adjmat/rt-matrixsum` passes the "value" test but not the "force test"

If you try to compile/modify this remember that the dependency part of the build tree is somehow broken with nvc++ (at least with mine) **and as now I am not trying to fix it**. So to compile after a modification you have to manually delete the interested files (like this `rm src/colvar/Distance.o ; make -j12`)

PR #1211, opened 2025-02-24
Moving the support code in a easier to use place

Iximiel:
##### Description

@gtribello
as we discussed I am moving the code into easier-to-use places
I am doing a commit per piece of code, starting from the View.

The plan is
 - [X] View
 - [ ] View2D
 - [ ] MultiColvarInput
 - [ ] ColvarInput
 - [ ] ColvarOutput

right?
Colvar input and ouput will likely end up in the same file

I will set up the PR directly with the versions of those classes that works for openacc, to simplify the rebasing/merging of #1206 

For the changes in Pbc, as now I am using `using` to speed up, but in a second moment we can change "memoryView" to plain view

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/ParallelTaskManager.h
+++ b/src/core/ParallelTaskManager.h
@@ -22,44 +22,15 @@
 #ifndef __PLUMED_core_ParallelTaskManager_h
 #define __PLUMED_core_ParallelTaskManager_h
 
-#include "tools/Communicator.h"
 #include "ActionWithVector.h"
 #include "ActionWithMatrix.h"
+#include "tools/Communicator.h"
 #include "tools/OpenMP.h"
+#include "tools/View.h"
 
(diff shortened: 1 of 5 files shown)


gtribello:
HI @Iximiel 

I think that the following tools can be moved to the tools directory:

* View
* View2D
* ColvarOutput

MultiColvarInput will likely only be ever used in MultiColvarTemplate.  If you want to move it to a separate file in the colvar module that is fine with me.  However, if it were me, I would just leave those definitions in MultiColvarTemplate.  

Similarly, ColvarInput is likely only going to be used in files that include MutiColvarTemplate.  I would thus also be tempted to leave these definitions in MultiColvarTemplate.  It could also be put in a separate file in the colvar module, though.

PR #1221, opened 2025-03-12
Avoid warning about strncpy

mabraham:
GCC 12.1 can warn like

```
warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound
512 equals destination size [-Wstringop-truncation]
```

##### Description

PLUMED intends to ensure that the C string is null terminated and has a known maximum length, which this change makes clear to the compiler.

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/wrapper/Plumed.h
+++ b/src/wrapper/Plumed.h
@@ -2461,7 +2461,7 @@ private:
     char msg[__PLUMED_WRAPPER_CXX_EXCEPTION_BUFFER];
     void init(const char* msg) __PLUMED_WRAPPER_CXX_NOEXCEPT {
       this->msg[0]='\0';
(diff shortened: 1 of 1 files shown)


mabraham:
Looks like this doesn't work as I thought. Progress is at https://gitlab.com/gromacs/gromacs/-/merge_requests/5030

mabraham:
For GROMACS, I've proposed a compiler-command-line suppression of this warning for the PLUMED code. That's good because we won't lose anything when new versions of Plumed.h are contributed.

For PLUMED itself, you might consider approaches like
```
#if __GNUC__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wpragmas"
#pragma GCC diagnostic ignored "-Wstringop-truncation"
#endif
    strncpy(...)
#if __GNUC__
#pragma GCC diagnostic pop
#endif
```
but I haven't proposed that here because I don't know how you feel about that.

GiovanniBussi:
Thanks @mabraham !!!!

PR #1223, opened 2025-03-13
New documentation for many actions

gtribello:
##### Description

This commit updates the documentation for many actions after the big changes of version 2.10.  It also ensures that features that have never had proper documentation now have some documentation.  Significant work is still required to update the rest of the documentation so that it is compatible with the new manual structure, but this a reasonable start on the process.  

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hi @GiovanniBussi 

The problem with codecheck here occurs because I have included a 

```c++
PLUMED_REGISTER_ACTION(Distance,"DISTANCE")
```

in the documentation for LOAD.  In other words, the error being thrown is definitely a false positive.  Is there anyway we can fix this so that it doesn't throw an error?

GiovanniBussi:
Currently there is no way. Could you perhaps change this:

# before
Notice that the `Distance3.cpp` uses the command:

```c++
PLUMED_REGISTER_ACTION(Distance,"DISTANCE")
```

Instead of the command:

```c++
PLUMED_REGISTER_ACTION(Distance,"DISTANCE2")
```

# after
Notice that the `Distance3.cpp` registers the shared library as `DISTANCE` rather than as `DISTANCE2`

PR #1225, opened 2025-03-17
Gpu/secondary structure DRMSD

Iximiel:
##### Description

# This is not complete and will likely need a rebase and some code arrangement 

Like removing the cout and cerr calls

I did a few things:
 - I changed the vector of maps into a sort of sparse matrix of structs for the DRMSD
 - Since there are some problems with the reductions in the secondaryStructureBase I had to remove them from the main loop in the GPU implementation, the MulticolvarTemplateBase is still working, I just have to signal that it needs the standard procedure
 - I added a way to have a bit more control on the GPU memory and it should be the way if we want something allocated on the GPU between steps
 - The changes in Matrix (the extra pointer) are due to memory management problems with openACC


Iximiel:
@gtribello, it looks like the DRMSD is working now (based on that test). Are you ok with the names for the new settings?

gtribello:
Hi @Iximiel 

Can you give a bit more details on what you mean by "the names for the new setting," please.  I am not sure what you are asking, sorry.

Iximiel:
@gtribello Sorry, I was referring to the new enums that I used to activate/deactivate part of the parallel task manager, and to the names of the new functions

gtribello:
Hi @Iximiel 

Sorry, I am only getting to this now.  The sun was shining yesterday (this is a rare event here), so I took a few hours off to enjoy being outside.

I had a look at the code you have written.  I am fine with the idea of parallelising over the atoms rather than over the individual tasks.  I think it would be useful to rearrange the code a little.  The spirit of this thing we are writing is that all the ACC parallelism stuff should be in the ParallelTaskManager.  In other words, we want to avoid having ACC pragmas in methods like SecondaryStructureBase.  The ACC pragmas should be in ParallelTaskManager.  With this in mind, I would recode this with the parallel loop from SecondaryStructureBase inside ParallelTaskManager (PTM).  This way of gathering the forces can then be triggered whenever `nthreaded_forces==action->getNumberOfDerivatives()`.

I know this is a bit of a pain, but I think it will save time in the long run, as you will likely be doing something similar for RMSD_VECTOR and also, once we get around to implementing the parallel version of the adjacency matrices.  For all these actions the condition above is satisfied, so if we have one general implementation for a parallel loop to gather the forces when this is true, we will be writing (and maintaining) less code in the long run.

Thanks
Gareth

Iximiel:
> The spirit of this thing we are writing is that all the ACC parallelism stuff should be in the ParallelTaskManager. In other words, we want to avoid having ACC pragmas in methods like SecondaryStructureBase.

The `#pragma acc loop seq` in the  "transposed" gather forces is a necessary evil to ensure that loop is executed sequentially in that kernel, because openacc removes some control on how the code behave and I do not want to risk that loop to be parallelized without a reduction.

To solve that I can move the "transposed gather" into the PTM and offer it as a "standard" for when that kind of gather is necessary (is what you were proposing for RMSD_VECTOR?). As it is already done for the virial here.


> This way of gathering the forces can then be triggered whenever nthreaded_forces==action->getNumberOfDerivatives().

The problem of that is that I know that the reduction of the virial works on 9 elements, but I do not know when it stops working with `9<nthreaded_forces<action->getNumberOfDerivatives()`. I can move the trigger on `nthreaded_forces==action->getNumberOfDerivatives()`, since as now we only have this two cases. The strength of the current implementation, with the "constexpr settings" is that the "if" to chose between the standard gather and the "transposed gather" is resolved at compile time, the alternative is to duplicate the loop for the two cases.

gtribello:
Hi @Iximiel 

I think I understand some of the issues you raise, but I am not sure.  Knowing the other things that we have to implement, I wouldn't worry too much about the virialSize issue.  The only actions where you would use the transposed gather are actions where the input is a collection of atom positions.  We can worry about that when we get to those actions.  I think they are not very urgent as it is the actions that calculate things from atoms that are slow.

In terms of getting stuff to run at compile time, I would still add the compile time variable to set this gather mode in the class where the PTM is being used.  Notice that in `setupParallelTaskManager` you have the following lines:

```c++
if( nt<0 ) {
    nthreaded_forces = action->getNumberOfDerivatives();
  } else {
    nthreaded_forces = nt;
  }
```

If it were me, I would add a `plumed_massert` command in the nt<0 branch that triggers an error if the option that sets the compile tag flag to do the atom-wise reduction. 

I hope that makes sense.

Iximiel:
@gtribello now I removed the enums to add a SFINAE thing that automagically detects the presence of the "custom" gather forces in the PTM, so there is no need to modify other PTM users. Now it checks for the virial size only it the custom_GatherForces is present

I set up the `plumed_massert` in `setupParallelTaskManager` but, at least on my machine, gives `double free or corruption (!prev)` when plumed exceptions are thrown (I think I should blame the compiler here)

gtribello:
@Iximiel I think my previous comment may not have been clear.  I wouldn't have a `gatherForces_custom` function in the SecondaryStructureBase class.  I would instead implement everything from `gatherForces_custom` in `ParallelTaskManager` something like this:

```c++
#pragma acc loop seq
  for(unsigned t=0; t<partialTaskList.size(); ++t) {
    std::size_t task_index = partialTaskList[t];
    auto fdata = ForceInput { input.ncomponents,
                              values+input.ncomponents*task_index,
                              nderivPerComponent,
                              derivatives+ndev_per_task*t};

      int ipos = T::getPositionInDerivArray( t, forcesIndex, actiondata );
      if( ipos<0 ) continue;
      for(unsigned c=0; c<input.ncomponents; ++c) {
             notThreadSafeForces[derivIndex] += ff*fdata.deriv[c][ipos];
      }
}
```

`int T::getPositionInDerivArray( t, derivIndex, actiondata )` is a static method in SecondaryStructureBase that returns -1 if `forcesIndex` is not used in the tth task.  If it is used it returns the position of that derivative with respect to that input component in deriv.

My point is that this `gatherForces_custom` method can be used for more than just the secondaryStructure variables, so why not code it so that the function is reusable?

Iximiel:
I see.

I would prefer to do migrate to a common function when we have another PTM user that needs its own gatherForces_custom, so it will be easier to make the new one work and then collapse the two functions into one.

Iximiel:
I have rebased the changes with the merge on master

gtribello:
OK @Iximiel so is this ready to merge?

Iximiel:
Not quite

I have recompile and test after the merge and I should understand when to suggest the user to use the GPU implementation. I have also to push a few changes that make the DRMS slightly faster, I think, also for the CPU

As now only the DRMSD part is working

Iximiel:
@gtribello if you are ok with this, in the current state is mergeable , but I managed to set up only the Secondary structure with DRMSD is on the GPU.

It has also the same problem of [final-backpropegation](https://github.com/plumed/plumed2/tree/final-backpropegation) of non passing  gridtools/rt-spherical-integral-2/, pamm/rt-pamm-aperiodic/, volumes/rt-coordination-insphere/, volumes/rt-lq6-vol/ and volumes/rt-q3-insphere in the CI, but passing them locally...

If not, I think that it should a good idea at least to cherry-pick the commits with the tests that I should have added some time ago for the views and the one with the configure fix

PR #1231, opened 2025-03-24
Reduced precision of numbers in output files for two regtests for ves…

gtribello:
… module that are failing on github

##### Description

This hopefully fixes the problem with the ves regtests that we have recently seen on GitHub.

To make this work, I added an option GRID_FMT that sets the precision of the grids that are output by the VES_LINEAR_EXPANSION action.  The same format is then used for all the output files that are output by this command.  This is why there are more changes to the regtests than just the single file that is currently failing.

This needs to be checked by @valsson we can merge it as this is his code.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Fine to me!

GiovanniBussi (approved):

GiovanniBussi:
I will wait for @valsson to confirm this before I prepare v2.10 release

valsson:
@gtribello thank you for this! 

Overall, this looks fine to me. However, there is one small change I would like to make, to change the variable name from `gfmt_` to `grid_fmt_` to be consistent with the other grid variable names in this part of the VES code (e.g., `grid_min_`, etc.), so that would make the code clearer to me. 

@gtribello can you make this change since you have a branch for this change, or should I do them? 

Omar

Issue #1205, opened 2025-02-18
ENERGY bug, gromacs 2024 + plumed 2.9


invemichele:
There is an issue with gromacs 2024+plumed 2.9 when using ENERGY as CV.

Here is the result of [this tutorial](https://github.com/invemichele/masterclass-22-03/blob/master/1-opes_expanded/multithermal/plumed.dat) with gromacs 2023 (blue) vs gromacs 2024: 
(dumping energy force with `DUMPFORCES ARG=ene STRIDE=500 FILE=FORCES`)

![Image](https://github.com/user-attachments/assets/2348df0b-e4d0-4e78-b943-e816f6390a89)
![Image](https://github.com/user-attachments/assets/0a1fde2b-f2b8-4e95-aec0-649fc8d4c8e1)

As you can see with the new gromacs version the multithermal simulation is not working anymore.

I tried with METAD as well, `metad: METAD ARG=ene PACE=500 SIGMA=10 HEIGHT=0.1`, and the results are more subtle but I think its quite clear there is a bug. When biasing e.g. phi the plots are almost identical.

![Image](https://github.com/user-attachments/assets/22b7e286-52b9-4c13-ad8e-9b236e7ddc64)
![Image](https://github.com/user-attachments/assets/8ed666d0-7dbc-4670-a377-036b94688b47)

I have no idea where to look to debug this, but let me know if I can do anything to help!

invemichele:
I used Claude Opus 5 to fix this bug, see #1444 #1445 #1446 (apologies for the very verbose description in the PRs).
While doing so, Claude spotted two minor bugs in v2.10, and realized that #567 is actually fixed now.
I did have a look at the code and tried to test it extensively to avoid introducing further issues, but I only performed  small test runs, not full production MD.
Please let me know if I can do anything else to help with this! @carlocamilloni @GiovanniBussi @gtribello

PR #1233, opened 2025-03-27
Checking for replex in the gromacs 2025 patch

Iximiel:
##### Description

I am adding the replex check following the discussion in #1162

##### Target release

I would like my code to appear in release __V2.10__

##### Type of contribution

##### Copyright

##### Tests


invemichele:
It would also be nice to raise an error if ENERGY is used for biasing, see https://github.com/plumed/plumed2/issues/1205.
Unless of course this bug is no longer present with gmx 2025...

GiovanniBussi:
@Iximiel this is a good point you can use the API to check this, "isEnergyNeeded"

GiovanniBussi:
@Iximiel sorry, probably it's not necessary as you are not passing the energy from this interface. So, ignore my previous comment

PR #1234, opened 2025-03-31
An eigensolver specialized for the 4x4 matrices that are used in RMSD

Iximiel:
##### Description

I want to implement the parallel task manager on GPU (see #1225) also for the secondary structure, and secondary structure passes through the RMSD class. The problem is that `diagMatSym` is a wrapper to a blas/lapack call. GPU do not like that, unless I compile the [call tree of dysevr](https://www.netlib.org/lapack/explore-html-3.6.1/d2/d8a/group__double_s_yeigen_ga2ad9f4a91cddbf67fe41b621bd158f5c_ga2ad9f4a91cddbf67fe41b621bd158f5c_cgraph_org.svg) for the GPU. 

So, I look at the 4x4 matrix, and I said "it is only a 4x4 matrix, that's a small matrix, there will be a big formula for that" in the hope of finding something that is not iterative. Hope crushed and a month passed and I am here with a collection of confusing results.

Thanks for @GiovanniBussi that pointed me at https://journals.iucr.org/a/issues/2005/04/00/sh5029/index.html and a good dose of Wolfram I put down those formulas.

 - The eigenvalues are fine.
 - The eigenvector not.
 - on my pc, with the basic configuration The test that use the RMSD more or less pass, like the secondary structure ones. I find this quite confusing. Since my own test for the solver fails
 - basic/rt-43 fails with:
```
FAILURE
Diff for deriv:
13c13
<  0.100000 1   0.0157   0.0024   0.1590  -0.4591   0.0157   0.0024   0.1590  -0.4590
---
>  0.100000 1   0.0157   0.0024   0.1590  -0.4591   0.0157   0.0024   0.1590  -0.4591

```

 - But my new test with the reference data generated with `diagMatSym` does not pass. Those references are quite synthetic, maybe is that, but I need another opinion.

The code needs a lot of improvements, but if I look at the performance I think that looks like a success from that point of view (this is the data from my pc on the new test I added):
```
                                              Cycles        Total      Average      Minimum      Maximum
diagMatSym                                       407     0.003853     0.000009     0.000003     0.000274
diagMatSym_all                                     1     0.003920     0.003920     0.003920     0.003920
mysolver                                         407     0.000238     0.000001     0.000000     0.000004
mysolver_all                                       1     0.000262     0.000262     0.000262     0.000262
```
My code does uses std::complex, pow and sqrt a lot, so there is room for (a lot) of improvement

##### Target release

I would like my code to appear in release v2.11 or before?

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Superseeded  by #1246

PR #1236, opened 2025-04-04
Moving the Trajectory parser in src/tools from Driver

Iximiel:
##### Description

This is a think that I and @gtribello were talking about some weeks ago: I need a way to run some more "interesting" benchmarks (with the secondary structures) and he would like to have a proprietary way of loading files in python. So I moved the file parsers from the driver in an ad hoc class, that can be embedded in the benchmark and/or in a new action.

The header is set up in pimpl fashion to avoid to propagate that `using namespace PLMD::molfile;`

I separated the various parsers because with the tree of ifs I had problems in understanding where I did something wrong. (and it might make easier to add a new parser in the future).

Before you ask, the singleton for molfile is there because somehow the static maps did not get initializated, on a plus side now those maps are more protected.

Should I add some extra make-type regtest to see that if everything appears broken, that might be actually the parser fault? As now every 'driver' test is a test for the parser

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
I didn't check the code, but I agree, this is something I also wanted to do since a very long time

PR #1237, opened 2025-04-07
Fixed documentation for command line tools to work with new manual

gtribello:
##### Description

This PR puts the documentation for the items in the cltools module into the format that is used in the new manual.  I moved all the information about bash autocompletion for these tools to the module page for the module as it makes sense there.  Also note that PlumedToHTML has been updated so tooltips are added when you have an input like this:

```plumed
plumed driver --noatoms --parse-only
```

to explain the various flags in the command line input.  Similarly, if you use a tool like simplemd that has an input file with keyword value pairs you can have tooltips on the values to explain what they do.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
This is working now @GiovanniBussi and @carlocamilloni

PR #1240, opened 2025-04-08
Changing the parallel action input to not own the argument bookeping data

Iximiel:
##### Description

@gtribello this is what I was mentioning in https://github.com/plumed/plumed2/pull/1214#issuecomment-2786469822

Currently is not astyled because is still not working on the GPU.

Are you ok with the Input set up like this?

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
@Iximiel this change is OK.  I don't understand why you have only had to change the calls from bookkeeping.data() to bookeeping in MatrixTimesVector base.  Surely, I use that bookeeping array in some other places as well?

Iximiel:
>Surely, I use that bookeeping array in some other places as well?

Yes, but usually you get the pointer from the BookeepingHolder or in other ways and not from the input class.

I configure with `./configure --enable-openacc=yes --enable-modules=all:-membranefusion CXXFLAGS="-O3 -gpu=cc75"` to cover the more code possible (membranefusion has a beef with nvc++, so I have to exclude it), and looking at the CI, I think I covered all the cases

src/core/ParallelTaskManager.h
```diff
@@ -462,68 +578,102 @@ void ParallelTaskManager<T>::applyForces( std::vector<double>& forcesForApply )
...
+    auto partialTaskList_data = ptl.devicePtr();
 
+    OpenACC::memoryManager ffa {forcesForApply};
+    auto forcesForApply_data = ffa.devicePtr();
+    const auto forcesForApply_size = ffa.size();
+
+    OpenACC::memoryManager ofd {omp_forces[0]};
+    auto omp_forces_data = ofd.devicePtr();
+    const auto omp_forces_size = ofd.size();
+    const auto reduction_size = (has_custom_gather)
```
Iximiel, line 603: @gtribello as you can see I cherry-picked some modifications from the [final-backpropegation](https://github.com/plumed/plumed2/tree/final-backpropegation) to ease part of the implementation.

But there is a problem:

I removed the reduction in the first loop of applyforces, everyone here in the office said that reducing an array was a bad idea and indeed it was:

For some reason some nans started to appear in the first value of the virial, for some multicolvars, in a consistent (same input, same nan) but illogic way (changing a CV option (like asking for componets) made the nan go away in some case). But if I print on screen the array I use to prepare the GPU data **before the gpu loop**, but after moving them to the gpu for example in the line I am writing this comment on, the nan disappears (by printing the whole omp_forces[0]) or changes place (if I print only omp_forces[0][0], for some reason the second component of the virial became nan...)...

In the confusion I moved the GPU virial to its own loop with a safer reduction on a variable instead of the reduction on the array.

The solution i found, of duplicating `gatherForces<GPU>` applies only for multicolvarstemplate, with minimum code duplication. Would you prefer a different solution?

gtribello: Hi @Iximiel 

I just want to check that I understand what you are suggesting here.  My understanding is that you are changing the way that forces are gathered.  There are essentially two methods that you use.  You either do what I did, namely:

```c++
for(unsigned i=rank; i<nactive_tasks; i+=stride) {
        std::size_t task_index = partialTaskList[i];
        ParallelActionsOutput myout( myinput.nscalars,
                                     fake_vals.data(),
                                     derivatives.size(),
                                     derivatives.data() );
        // Calculate the stuff in the loop for this action
        T::performTask( task_index, actiondata, myinput, myout );

        // Gather the forces from the values
        T::gatherForces( task_index,
                         actiondata,
                         myinput,
                         ForceInput( myinput.ncomponents,
                                     myinput.nscalars,
                                     value_stash.data()+myinput.nscalars*task_index,
                                     nderivatives_per_component,
                                     derivatives.data()),
                         forces );
}
``` 

This gathers the forces in two arrays, which internally to the action are called `thread_unsafe` and `thread_safe`.  A reduction is then done on the `thread_safe` array  

The alternative is to do what you are proposing here:

```c++
#pragma acc parallel loop
      for(unsigned i=0; i<nactive_tasks; ++i) {
        std::size_t task_index = partialTaskList_data[i];
        ParallelActionsOutput myout { input.nscalars,
                                      valstmp+input.nscalars*i,
                                      ndev_per_task,
                                      derivatives+ndev_per_task*i};
        // Calculate the stuff in the loop for this action
        T::performTask( task_index, t_actiondata, input, myout );
        if constexpr (!has_custom_gather) {
          // Gather the forces from the values
          T::gatherForcesGPU( task_index,
                              t_actiondata,
                              input,
                              ForceInput { input.ncomponents,
                                           input.nscalars,
                                           value_stash_data+input.nscalars*task_index,
                                           nderivPerComponent,
                                           derivatives+ndev_per_task*i},
                              ForceOutput { omp_forces_data,
                                            omp_forces_size,
                                            forcesForApply_data,
                                            forcesForApply_size }
                            );
        }
      }

#pragma acc parallel loop
      for(unsigned v=0; v<virialSize; ++v) {
        const  unsigned m = input.nindices_per_task*3 + v;
        double tmp=0.0;
#pragma acc loop reduction(+:tmp)
        for(unsigned t=0; t<nactive_tasks; ++t) {
          std::size_t task_index = partialTaskList_data[t];
          auto fdata = ForceInput { input.ncomponents,
                                    input.nscalars,
                                    value_stash_data+input.nscalars*task_index,
                                    nderivPerComponent,
                                    derivatives+ndev_per_task*t};
          for(unsigned i=0; i<input.ncomponents; ++i) {
            tmp += fdata.force[i]*fdata.deriv[i][m];
          }
        }
        omp_forces_data[nthreaded_forces-virialSize+v] = tmp;
      }
    }
```

Now, you only set the `thread_unsafe` forces in the `gatherForces` method.  The `thread_safe forces are gathered in a second set of loops over all the tasks that are performed within the PTM.

If we need to use this second mechanism for the `thread_unsafe` forces that is fine with me.  However, it appears to me that you now have two separate implementations of this mechanism.  The first of these was implemented when you worked on the SecondaryStructure variables and was added in the SecondaryStructure class.  The second of these is now being added inside the `ParallelTaskManager` and relies on using the `virialSize` variable.  

When you set up the ParallelTaskManager objects in the classes  that include a PTM you have a call like this:

```c++
taskmanager.setupParallelTaskManager( natoms_per_task,
                                        3*natoms_per_task + virialSize,
                                        virialSize );
``` 

The third argument here is the size of the `thread_safe` array.  It is translated to a variable called `nthreaded_forces`.  If you rewrite the second loop in the second option above as:

```c++
#pragma acc parallel loop
      for(unsigned v=0; v<nthreaded_forces; ++v) {
        const  unsigned m = forcesToApply.size() - nthreaded_forces;
        double tmp=0.0;
#pragma acc loop reduction(+:tmp)
        for(unsigned t=0; t<nactive_tasks; ++t) {
          std::size_t task_index = partialTaskList_data[t];
          auto fdata = ForceInput { input.ncomponents,
                                    input.nscalars,
                                    value_stash_data+input.nscalars*task_index,
                                    nderivPerComponent,
                                    derivatives+ndev_per_task*t};
          for(unsigned i=0; i<input.ncomponents; ++i) {
            tmp += fdata.force[i]*fdata.deriv[i][m];
          }
        }
       omp_forces[v] = tmp;
      }
    }
```

then the PTM offers a general mechanism.   Surely, it should be possible to make this work for the SecondaryStructure variables and the MultiColvars without having two specialized pieces of code for the two methods.

Iximiel: Before answering, the most important thing to understand is why this change, at cost of repeat myself: reducing on the elements of an array is not guaranteed to work. Before it was working fine, but now in some of the case where it was working is broken and I do not understand why. And this is why I separated the virial from `gatherForces`. For the secondary structure the cause was that asking to reduce the array on all atoms caused the program to crash on entering the loop.

> it appears to me that you now have two separate implementations of this mechanism.

The secondary structure and this implementation are compatible. Moreover, the secondary structure also depends on the second loop on the VirialSize because the atoms and the virial reduce in two fundamentally different ways, I think I am explaining it better in a few lines.

> When you set up the ParallelTaskManager objects in the classes that include a PTM you have a call like this:
>...
>Surely, it should be possible to make this work for the SecondaryStructure variables and the MultiColvars without having two specialized pieces of code for the two methods.

Adding some context: I just tried to add USEGPU to `QuaternionBondProductMatrix`, because it was one of the few CV with the PTM in a cpp, and it appeared different enough to be used to check if this mechanism is easy to apply to other CVs. And I learned a few new things:
in the case of `QuaternionBondProductMatrix` I need both _custom and the GPU `gatherForces`, but the virial loop is not necessary. And the `gatherForces` may look quite different between each other.

In the three example that I have I need different way of reducing the race-prone part of the forces, and even if virial and atomic derivatives looks alike, their sum have different 'boundary conditions'. If I write down just the sum reductions:
```c++
//virial, the race is in v , all task contribute to all v
//for each c, the same m contributes to the same v
omp_forces[v]+= fdata.force[c]*fdata.deriv[c][m];
//secondary structure, the race is in atomID, some task contribute to some atomID
//for each c, the same m might not be contributing to the same atomID
omp_forces[atomID] += fdata.force[c]*fdata.deriv[c][m];
//quaternion () the race is in force_index, a few tasks contribute to a force_index
//encased in a nested for loop i:0->4,j:0->4
omp_forces[j*nquat + force_index ] += fdata.force[i]*fdata.deriv[i][4+j];
```
they are very similar, but their context changes quite a lot.



The virial can be used as template for reducing when the components omp_forces depends on all the tasks

The one for the secondary structure when the components have a optional dependency on the tasks

QuaternionBondProductMatrix is an example of a more specialized reduction (and probably I am saying this because that is the only example I saw with this kind of pattern) -probably can be written in a way that it is similar to the one of secondary structure in the sense that has a "sum if" parameter, but the "if" part is different from the one in secondary structure

src/crystdistrib/QuaternionBondProductMatrix.cpp
```diff
@@ -212,281 +238,205 @@ std::vector<unsigned>& QuaternionBondProductMatrix::getListOfActiveTasks( Action
 }
 
 void QuaternionBondProductMatrix::performTask( std::size_t task_index,
```
Iximiel, line 240: The old version of perform task did opposed some resistance on running on the GPU. Removing the vectors made it possible to run it.

I changed the rest of the function to a more declarative style to reduce the number of loops needed by it, and to remove nearly all the branches nested within the for loops. So now is more GPU-friendly

Iximiel:
@gtribello I was looking at `AdjacencyMatrixBase`, and in particular to `DistanceMatrix`, because it looks like a simple starting point for `USEGPU`.

The problem there is that the `AdjacencyMatrixBase::performTask` declares a bunch of std::vectors. 
The one encapsulated in `AdjacencyMatrixInput` can be swapped for a `View<const Vector>` (I tried and works), this avoid copies and eventual resize, and makes `AdjacencyMatrixInput` more copy-passable.
The other one is the one at the start of the function, the buffer `std::vector<Vector> atoms( nneigh + n3neigh );`, that is more delicate because on the CPU is "free to set up" but on the GPU needs to be pre-allocated before enqueuing the kernels. The solution I am trying for check if it is feasible is: writing a performTaskBuffered that accepts the buffer as input and call it from performTask as usual, where I allocate the buffer, so I do not change PTM-CPU, and directly call the performTaskBuffered version from the PTM-GPU loop. Do you figure something more linear?

gtribello:
Hi @Iximiel 

Firstly, I'm not sure if it is meant to be hyperbole, but two is not what I would call a bunch.  A bunch would need to be at least five.  If I meant a number close to two, I would say a couple.

Pedantry aside, I have a proposed solution for this problem.  I need to go to the doctor's now for this wasp sting, but I hope to be able to update you on this a little later in the day.

Iximiel:
>Pedantry aside, I have a proposed solution for this problem. I need to go to the doctor's now for this wasp sting, but I hope to be able to update you on this a little later in the day.

I am looking forward to see it

Iximiel:
@gtribello this is ready for gpuwork/matrices. It still has the gatherForcesGPU/custom waiting for passing to the new index thing, so that this step do not break anything.
I kept the modification (and the tests) to performTask in QuaternionBondProductMatrix, because I know that function works on the GPU, so it is already set up for the next step

PR #1243, opened 2025-04-16
Added new mechanism for forces in ParallelTaskManager

gtribello:
##### Description

@Iximiel and @GiovanniBussi, I am not entirely happy with the GPU implementations we have thus far.  My concern is that we have thus far implemented the GPU parallelization in three of the methods that use the ParallelTaskManager class.  All three of these implementations use applyForces part of the ParallalTaskManager differently.

* `MultiColvarTemplate` gathers the derivatives on the atoms using the `gatherForces` method I wrote and then uses a mechanism for gathering the virial that is implemented in `ParallelTaskManager`
*  `SecondaryStructureBase` has a custom method for gathering the forces on the atoms that is implemented in `SecondaryStructureBase` and then uses the mechanism for gathering the virial in `ParallelTaskManager`
*  `QuaternionBondProductMatrix` uses a custom method for gathering the forces and does not use the mechanism for gathering the virial in `ParallelTaskManager` because there is no virial.

Our objective when we started this project was to write a `ParallelTaskManager` action that could handle the parallelism across many of the classes.  My view is that if every class uses this method differently, we might as well abandon the abstraction and move to local implementations of the parallelism in each method.  If we are introducing an abstract class, we shouldn't be using it differently in every action we are using it in. 

With that in mind, I am proposing the following rethink of the `applyForces` part of the `ParallelTaskManager`.  Now, rather than having a `gatherForces` method in `MultiColvarTemplate`, `SecondaryStructureBase` and `QuaternionBondProductMatrix`, we instead have a `getForceIndices` method.  This method returns the indexes in `forcestoApply`  the forces for task `task_index` act upon.  This makes it straightforward to do the gathering of forces within `ParallelTaskManager` directly.

I have tried to modify the GPU code so that it also uses this mechanism in a way that is what is compatible with what was done in `final-backpropegation`, but I have obviously not tested it.  Can we investigate whether this is a viable strategy for getting rid of the customisation, please?  If it is I will then port these changes onto the gpuwork/matrices branch and will use it for all the additional functionalities that use the PTM on that branch.

I would like to do this now before we go much further with the current strategy.  If @Iximiel you want to have a meeting to discuss what I am proposing here then I am free all day tomorrow and Friday as well as most days next week.  Let's try to keep it constructive, though.  I am not saying that what you did is wrong.  In hindsight, I didn't fully understand how the GPU works when I wrote the API that was based on `gatherForces`.  I think that we can do better now that I have a slightly better understanding.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/colvar/MultiColvarTemplate.h
+++ b/src/colvar/MultiColvarTemplate.h
@@ -79,11 +79,11 @@ public:
                            const MultiColvarInput& actiondata,
                            ParallelActionsInput& input,
                            ParallelActionsOutput& output );
-  static void gatherForces( std::size_t task_index,
-                            const MultiColvarInput& actiondata,
-                            const ParallelActionsInput& input,
-                            const ForceInput& fdata,
-                            ForceOutput forces );
+  static void getForceIndices( std::size_t task_index,
+                               std::size_t ntotal_force, 
+                               const MultiColvarInput& actiondata,
+                               const ParallelActionsInput& input,
+                               View<std::size_t,helpers::dynamic_extent> force_indices );
 };
 
 template <class T>
@@ -165,6 +165,7 @@ MultiColvarTemplate<T>::MultiColvarTemplate(const ActionOptions&ao):
   // Sets up the index list in the task manager
   taskmanager.setupParallelTaskManager( natoms_per_task,
                                         3*natoms_per_task + virialSize,
+                                        3*natoms_per_task, 
                                         virialSize );
   taskmanager.setActionInput( MultiColvarInput{ usepbc, mode });
 }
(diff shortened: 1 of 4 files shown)


GiovanniBussi:
I am also available, either tomorrow early afternoon or Friday morning

gtribello:
Hello @Iximiel and @GiovanniBussi 

Can we try for a meeting on Friday.  How about 10:00 AM for you and 9:00 AM for me?

If that works, please send a Zoom invite.

I got stung by a wasp and it is looking like it has got infected so I may be spending a considerable portion of tomorrow at the doctor's (crying face)

Iximiel:
> Can we try for a meeting on Friday. How about 10:00 AM for you and 9:00 AM for me?

Ok for me

>I got stung by a wasp and it is looking like it has got infected so I may be spending a considerable portion of tomorrow at the doctor's (crying face)

😱

Iximiel:
@gtribello 

It looks like this is working 😃, I had to massage a little the part in the ifdef guard. But there are some suspect errors on the very first element of the virial of the secondary structure that I have to understand (I am talking about the GPU here). I'll try to checkout also the quaternion-bond multiplication into this to see if works also there. before merging this let me PR the working code into this

tomorrow I'll show you also a less invasive solution I am currently trying and that appear to be working. The difference is that I add a specialize gatherThread to be run on the GPU, mirroring the CPU implementation with some extra space occupied by the raceable forces instead of the indexes.

I need to take some measures between the two implementations and also to check the impact on the CPU implementation

gtribello:
Hi @Iximiel 

I just thought I would write a quick summary of my understanding of the next steps here from our meeting this morning.  So here are the steps in a list:

1. You do a PR on this branch of anything that is required to get this working on the GPU.
2. I will merge your PR on here.  I will then merge this into final-backpropegation.
3. I will merge final-backpropegation into gpuwork/matrices and rewrite things in there to use the new applyForces mechanism.
4. You check that gpuwork/matrices works ok on the GPU for secondarystructure and multicolvars and set up a PR for anything that needs to change in order to get that to work.
5. We merge gpuwork/matrices into final-backpropegation.

Notice that I am waiting on step 1 in order to start on step 2 so the sooner you can get that PR on here done the better.

I hope that is reasonably clear.

gtribello:
Hi @Iximiel 

I just merged the PR on `gpuwork/matrices` is there anything that needs to be changed on **this branch** to get secondarystructure and multicolvar working on the GPU, or can I go ahead and merge this into `final-backpropegation` and get started on merging `final-backpropegation` into `gpuwork/matrices`?

Iximiel:
> Hi @Iximiel
> 
> I just merged the PR on `gpuwork/matrices` is there anything that needs to be changed on **this branch** to get secondarystructure and multicolvar working on the GPU, or can I go ahead and merge this into `final-backpropegation` and get started on merging `final-backpropegation` into `gpuwork/matrices`?

Hi, @gtribello, out of the box:
 - multicolvar works
 - secondarystructureDRMSD sort of works
   - only fails for PARABETARMSD and only in the first non zero element of the force (in other words only a number per frame is wrong), the virial is ok

PR #1244, opened 2025-04-17
Removed allocation of vectors in some performTask methods and used so…

gtribello:
##### Description

Hi @Iximiel 

This is how I propose getting rid of the memory allocation in AdjacencyMatrixBase.  This introduces a variable into ParallelTaskManager called `workspace_size` that tells you how many doubles of extra memory are needed.  I then create a std::vector<double> with that size that I pass to ParallelActionOutput.

Is that something that you can make work with the GPU?

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/adjmat/AdjacencyMatrixBase.h
+++ b/src/adjmat/AdjacencyMatrixBase.h
@@ -50,8 +50,8 @@ public:
   const Pbc* pbc;
   Vector pos;
   std::size_t natoms{0};
-  std::vector<Vector> extra_positions;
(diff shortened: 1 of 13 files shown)


gtribello:
Got some antibiotics for my wasp bite as well just in case you were wondering.

Iximiel:
I noted that you did not change the buffer in SecondaryStructureBase:

To make it run on the GPU I modified to buffer to be a fixed size array, like this:
```c++
// std::vector<Vector> pos( actiondata.natoms );
  std::array<Vector,30> pos;
```
We already discussed about this, but without the buffer context. I do not know if is a good idea to keep it like this or if it is better to also SecondaryStructureBase to use the buffer

PR #1246, opened 2025-04-18
Optimized RMSD and made openACC friendly

GiovanniBussi:
@Iximiel this is an attempt to reimplement RMSD so that it is openACC friendly. Interesting, on a dummy benckmark where I just to the RMSD of a few atoms, I get a significant performance gain on my laptop wrt lapack.

The implementation is still a bit dirty. And let's see if it breaks any test on GHA.

The benchmark I did was:
```
cat > plumed.dat << EOF
r: RMSD TYPE=OPTIMAL REFERENCE=test0.pdb
RESTRAINT ARG=r AT=0.0 KAPPA=1.0
EOF

cat > test0.pdb << EOF
ATOM      1  O   ALA     2      -0.926  -2.447  -0.497  1.00  1.00      DIA  O
ATOM      2  HNT ALA     2       0.533  -0.396   1.184  1.00  1.00      DIA  H
ATOM      3  HT1 ALA     2      -0.216  -2.590   1.371  1.00  1.00      DIA  H
ATOM      4  HT2 ALA     2      -0.309  -1.255   2.315  1.00  1.00      DIA  H
ATOM      5  HT3 ALA     2      -1.480  -1.560   1.212  1.00  1.00      DIA  H
ATOM      6  CAY ALA     2      -0.096   2.144  -0.669  1.00  1.00      DIA  C
EOF

plumed-runtime benchmark --kernel this:/Users/bussi/plumed2/tmp/ref-rmsd/lib/libplumedKernel.dylib
```

I measure a gain of ~ 50%

```
BENCH:  Kernel:      this
BENCH:  Input:       plumed.dat
BENCH:  Comparative: 1.000 +- 0.000
BENCH:                                                Cycles        Total      Average      Minimum      Maximum
BENCH:  A Initialization                                   1     0.004797     0.004797     0.004797     0.004797
BENCH:  B0 First step                                      1     0.001857     0.001857     0.001857     0.001857
BENCH:  B1 Warm-up                                       399     0.006421     0.000016     0.000008     0.000106
BENCH:  B2 Calculation part 1                            800     0.012943     0.000016     0.000009     0.000073
BENCH:  B3 Calculation part 2                            800     0.014405     0.000018     0.000009     0.000096
PLUMED:                                               Cycles        Total      Average      Minimum      Maximum
PLUMED:                                                    1     0.038202     0.038202     0.038202     0.038202
PLUMED: 1 Prepare dependencies                          2000     0.002209     0.000001     0.000000     0.000019
[...]

PR #1262, opened 2025-05-20
Test for compute_quaternion_from_K

Iximiel:
this adds the tests for #1246:
@GiovanniBussi I added the tests: to avoid putting selecting the sign for the eigenvectors in `compute_quaternion_from_K` in the function I just changed it in the test and checked that blas and `compute_quaternion_from_K` return both the same numbers.

In the test there is also a small benchmark to confront the two implementations.

The generated matrices can be improved I think. The rationale for setting them up is to replicate the 3x3 Tensor that it is used to set up the 4x4 matrix in RMSD and `randomRMSDmat` imitates that, with a "reference" and a collection of atoms that are put at a certain distance from the reference, maybe they should put on a sort of 'wobbly stick', since now are evenly distributed.


Iximiel:
@GiovanniBussi apparently, if you susbstitute the for loop for the normalizations with the `/=` operator you can speed this up a little

And I am discovering in doing this that `fixzeroes.sh` does not work on the files I set up as a reference

Iximiel:
the A/=frob instead of the loop brings the total time for the new function from  0.006334 to 0.005075! (in the benchmark of the test)

GiovanniBussi:
The test added here by @Iximiel has been manually added in a modified form directly in #1246, so that this PR can be closed

PR #1264, opened 2025-05-23
Refactored FunctionOfMatrix and FunctionOfVector to use PTM

gtribello:
##### Description

This is a continuation of the work that @Iximiel and I have been doing on refactoring the code so that the new parallel task manager (PTM) is used in more places.  The eventual aim of this work is to do more things with GPUs.  This uses the PTM in another two actions that I wrote in place of the code that I wrote earlier.  The actions used here are FunctionOfVector and FunctionOfMatrix.

Changes to regtests here just delete places where we are printing the derivatives of a sum object.  Note the derivatives output were not compared with a reference command.  In other words, there was just a dumpderivatives command that output a file for which there was no equivalent *.reference version

Making this change appears to improve performance.  In the chart below, I do timings for this input:

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM
PRINT ARG=q6_sum FILE=colvar
BIASVALUE ARG=q6_sum
````

Results are shown with and without the BIASVALUE action so as to toggle forces on and off.

q6-timings

You can see that the new version is faster if derivatives are employed.  It causes a slowdown if derivatives are not employed.  However, given that the calculation with derivatives is much slower, I think that the speed up there is what should inform our decision.

The chart below shows a similar set of benchmarks with this input:

````
q6: Q6 SPECIES=@mdatoms SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0}
lq6: LOCAL_Q6 SPECIES=q6 SWITCH={EXP D_0=1.0 R_0=0.1 D_MAX=2.0} SUM

PRINT ARG=lq6_sum FILE=colvar
BIASVALUE ARG=lq6_sum
````

lq6-timings

The result is essentially the same.  The fact that local Q6 with derivatives is now cheaper than Q6 with derivatives in the old code is pretty cool!

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
Hi @gtribello I just run the test on mine, I get the same test failures only if I export `PLUMED_NUM_THREADS=2`.
I am giving them a more deep look now

PR #1269, opened 2025-05-28
Final steps in creation of new manual

gtribello:
##### Description

This PR finalises everything for the new manual.  I have added back in the installation pages, the change log and the list of MD codes.

The installation pages are in a tutorial on the tutorial pages that is linked to from the manual.  I took the information from the current manual and incorporated it into the tutorial.  I think doing it this way makes it easier to have multiple guides to the installation process.

While I was trying to do the change log I realised that mkdocs can't do nested lists (which we use in the change log).  It also cannot do alerts.  There is an extension to mkdocs (materials for mkdocs) that allows these features and that seems to offer more features in general.  I have thus moved to using material for mkdocs instead of mkdocs.  This was a relatively small change to the code but the appearance of the manual changes.  The landing page now looks like this:

[front-page.pdf](https://github.com/user-attachments/files/20489176/front-page.pdf)

The action list looks like this:

[actionlist.pdf](https://github.com/user-attachments/files/20489187/actionlist.pdf)

(if @Iximiel could work out how to stop the html from splitting the names of actions here that would be great).

And a typical action page looks like this:

[action-example.pdf](https://github.com/user-attachments/files/20489195/action-example.pdf)

You get really nice looking alerts though as you can see here:

changelog

As part of this, I went through the old manual and carefully checked that all the information from it is inside the new manual.  I have thus ensured that we have transferred everything from the old manual to the new. manual.  Files in user-docs whose information has been transferred have thus been deleted.  The only things left in that directory are the old tutorials and the old bibliography (the references in that old bibliography should appear somewhere in the new manual though).

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
@gtribello I suggest you to fix the version of both mkdocs and mkdocs-material in the CI to avoid unexpected changes on version updates

> [actionlist.pdf](https://github.com/user-attachments/files/20489187/actionlist.pdf)
>
>(if @Iximiel could work out how to stop the html from splitting the names of actions here that would be great).

in plain mkdocs seems to adapt itself to the largest element of the columns.
Here maybe we can say force it do do it, tomorrow I'll give it a try

GiovanniBussi:
@gtribello a couple of comments:
- Can we reintroduce shortcuts in the new tool? I think using \issue{} for issues waw very convenient rather than using the full link in markdown. And there might be other use cases
- I looked at the installation manual [here]. I am not a big fan of how it was before, but I find this even more confusing for a number of reasons:
   - The flow chart of the topics is completely messed up (as you can see in the screenshot below)
   - The sentence "Click here to open manual pages for actions discussed in this tutorial," is unrelate to this page.
   - The doc is then duplicated (installation and installation_old). I don't know how easy it will be to maintain this information

image

gtribello:
Hi @Iximiel 

Here is another challenge for you if you are interested.  With the new materials for markdown documentation, it is possible to add a switch that converts the site between light and dark mode.  I have implemented this switch in this branch.  The problem, however, is that the tables don't switch between light and dark mode because I am using datatables.  There is a page on how to set up the dark mode for the tables here:

https://datatables.net/manual/styling/dark-mode

but I can't get that to work with our site.

Also note issue #1273.  Can't get that one fixed either.

Thank you

PR #1270, opened 2025-05-28
Using INSPHERE for implementing the switching function with the openacc workflow

Iximiel:
##### Description

I semi-duplicated the switching function implementation:

Now it can be run both as "a series of ifs" or with inheritance. 
I duplicated because it looked the easiest way of not removing lepton from the Switching function. lepton produces cpu assembly and not ptx (cuda assembly) I believe that it is not (easily) portable to openacc.

With some templating I managed to reduce the duplication of the code only to the interfaces:
I added a common Data structure with all the possible parameters (that can be downsized since lot of parameters are just occupying space if the switch do not uses them) and a series of classes that are namespace-like to assemble them lego-style for doing the calculations.
I added a class that make the new implementation usable like the previous one and used it as the "CPU" switching function. Here the inheritance is used because lepton has the same interface, but the inside is completely different and I did not wanted to store and extra object for the eventual lepton switch.

The non inherited class directly stores the parameters and an enum type with the kind of the function, then with a switch (a C++ one) select the static function to run, but it does not have lepton inside.

The `#pragma GCC diagnostic error "-Wswitch"` is there for saving mental health

I wanted to keep the switching function class the "simpler" possible to copy/move to simplify declaring the support classes for the PTM. Now it do not need to call `set` after being copied, this is shown in the support class `VolumeInSphere`

I should still add an `apply` method like the one in the pbc class.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
@GiovanniBussi @gtribello 
Here a few measurements on the switching function

<details>
  <summary>How I obtained these numbers</summary>


```bash
nsteps=1000
list_of_natoms=(10 500 1000 2000 4000 8000 16000)

export PLUMED_MAXBACKUP=0
export PLUMED_NUM_THREADS=4

for natoms in "${list_of_natoms[@]}"; do
    output=tmp_ter_${natoms}
    plumed benchmark --plumed="plumed.dat" \
        --natoms="${natoms}" \
        --nsteps="${nsteps}" \
        --kernel="{plumed_master_250611}/libplumedKernel.so:{plumed_newSwitch_250612}/libplumedKernel.so" \
        --atom-distribution=sc >"${output}"
    #sed -ne '/Single run, skipping comparative analysis/,$ p' < tmp
    echo "Atoms: ${natoms}"
    grep Comparative "${output}"
    grep Calculating "${output}"
done
```

plumed input:
```
DEBUG DETAILED_TIMERS

cv: COORDINATION GROUPA=@mdatoms R_0=1 NOPBC

PRINT ARG=* FILE=colvar FMT=%8.4f STRIDE=1

```

</details>

I ran it a few times:

image

image

image

Here I am measuring the "standard" switching function, it looks like It is going faster with the new modifications (I suppose that there is a little more inlining than before)

Iximiel:
@gtribello this is ready to go!

---

PS:
If I change to
```
DEBUG DETAILED_TIMERS

f: FIXEDATOM AT=0,0,0
a: AROUND ATOMS=@mdatoms ORIGIN=f SIGMA=0.2 XLOWER=-1.0 XUPPER=1.0 YLOWER=-1.0 YUPPER=1.0 ZLOWER=-1.0 ZUPPER=1.0
x: SUM ARG=a PERIODIC=NO
PRINT ARG=x FILE=colvar
```
As you andd Giovanni suggeste, the time spent in update disappear! So, no problem :D

PR #1287, opened 2025-06-18
Update metatomic documentation for the new manual

Luthaf:
##### Description

This updates the syntax for admonition, and rephrase some of the installation instructions. 

When building the manual locally with metatomic enabled, I get a `2.11 failed` sticker, I guess this is because the model at `path/to/model.pt` does not exists?

Screenshot

##### Target release

I would like my code to appear in release v2.11

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/metatomic/metatomic.cpp
+++ b/src/metatomic/metatomic.cpp
@@ -30,9 +30,11 @@ along with the METATOMIC-PLUMED module. If not, see <http://www.gnu.org/licenses
 /*
 Use arbitrary machine learning models as collective variables.
 
-\note This action requires the metatomic-torch library. Check the instructions
-in [the module page](module_metatomic.md) which includes instructions on how to
-enable this module.
+!!! note
+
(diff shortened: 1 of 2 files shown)


carlocamilloni:
@Luthaf about the example if the file is not too big you could point to something in the regtest folder doing like:
MODEL=regtest/...

Luthaf:
I just removed the model files in #1263, because these are treated as binary files by git, and need to be regenerated relatively often. 

I could add another step to the doc build that generate a model when building docs if all dependencies are present, what do you think?

carlocamilloni:
that could be nice, but only if it is not time consuming because builing the manual is already a relatively long process. @gtribello what do you think?

Luthaf:
Generating the model would be pretty fast (~couple of seconds), but would require additional python dependencies. I can also live with the red sticker in there for now

gtribello:
That sounds fine to me.  If you want to have a chat about it drop me an email and we can arrange a time.

PR #1268, opened 2025-05-27
Now the CI will try to use nvhpc for real

Iximiel:
##### Description

I noted that my modification to the CI with the NVHPC sdk was just installing it and then plumed was compiled with the gnu compilers...

I corrected the mistake and now it also checks that the openacc things gets compiled as expected.

The module `membranefusion` in not compatible with the current nvhpc compiler because nvc++ do not support the custom reductions in openMP. In the -nvhpc- matrix that module is disabled to  end the compilation and try the tests

##### Target release

I would like my code to appear in release __v2.11__

##### Type of contribution

- [x] changes to code or doc authored by PLUMED developers, or additions of code in the core or within the default modules
- [ ] changes to a module not authored by you
- [ ] new module contribution or edit of a module authored by you

##### Copyright

- [x] I agree to transfer the copyright of the code I have written to the PLUMED developers or to the author of the code I am modifying.

- [ ] the module I added or modified contains a `COPYRIGHT` file with the correct license information. Code should be released under an open source license. I also used the command `cd src && ./header.sh mymodulename` in order to make sure the headers of the module are correct. 

##### Tests

- [ ] I added a new regtest or modified an existing regtest to validate my changes.
- [ ] I verified that all regtests are passed successfully on [GitHub Actions](https://github.com/plumed/plumed2/actions).

PR #1292, opened 2025-06-23
New manual/gpu

Iximiel:
##### Description

@GiovanniBussi @gtribello 

I have updated the `gpu.md` file in the new manual with the information on the currently working openacc actions.

I also added some installation instructions, I would keep them there until we have something stabler.
On top of that I modified configure.ac in order to remove it from the scope of #1268 and, mainly, to make easier to setup openacc.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/configure
+++ b/configure
@@ -660,6 +660,8 @@ CPPFLAGS
 LDFLAGS
 CXXFLAGS
 CXX
+PLUMED_ACC_GPU
+PLUMED_ACC_TYPE
 MPIEXEC
 BASH_COMPLETION_DIR
 PYTHON_BIN
@@ -764,6 +766,8 @@ LDSHARED
 PYTHON_BIN
 BASH_COMPLETION_DIR
 MPIEXEC
+PLUMED_ACC_TYPE
+PLUMED_ACC_GPU
 CXX
 CXXFLAGS
 LDFLAGS
@@ -1455,7 +1459,9 @@ Optional Features:
   --enable-af_cpu         enable search for arrayfire_cpu, default: no
   --enable-libtorch       enable search for libtorch, default: no
   --enable-libmetatomic   enable search for libmetatomic, default: no
-  --enable-openacc        enable search for openacc, default: no
+  --enable-openacc        enable search for openacc (use PLUMED_ACC_TYPE and
+                          PLUMED_ACC_GPU for controlling the settings),
+                          default: no
   --disable-openmp        do not use OpenMP
 
 Some influential environment variables:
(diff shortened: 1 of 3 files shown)


Iximiel:
@GiovanniBussi @gtribello can you rereun the -doc-mpi-? it has failed with `../src/lib/plumed: error while loading shared libraries: libplumedKernel.so: cannot open shared object file: No such file or directory` and I find it strange

.github/workflows/linuxWF.yml
```diff
@@ -54,6 +54,10 @@ jobs:
         # path required for pytest:
         echo "$HOME/.local/bin" >> $GITHUB_PATH
         echo "PYTHONPATH=$HOME/opt/lib/plumed/python:$PYTHONPATH" >> $GITHUB_ENV
+        echo "CPATH=$HOME/opt/include:$CPATH" >> $GITHUB_ENV
+        echo "INCLUDE=$HOME/opt/include:$INCLUDE" >> $GITHUB_ENV
+        echo "LIBRARY_PATH=$HOME/opt/lib:$LIBRARY_PATH" >> $GITHUB_ENV
+        echo "LD_LIBRARY_PATH=$HOME/opt/lib:$LD_LIBRARY_PATH" >> $GITHUB_ENV
```
Iximiel, line 60: @carlocamilloni I suppose that removing these lines with the merge commit https://github.com/plumed/plumed2/commit/974267a192957c17adeeee0e74d5fff300b9b531 may have broken the manual build process 

If these are not the culprit, maybe sourcing the sourceme in before `make doc` may be the solution

---

another thing changed is the removal of the statement `LDFLAGS=-Wl,-rpath,$LD_LIBRARY_PATH` (not having the ldpath hardcoded in the executable might be another colprit) from the ./configure call, I think it has been removed with the @Luthaf's PR for torch, but I do not know why?

Iximiel:
> @Iximiel @gtribello @GiovanniBussi this is failing since the last time I merged v2.10 in master, but v2.10 was passing all tests so I do not know which of the changes in v2.10 is givin problems, I do not think is Max code but more the metatensor changes

maybe also these merge should pass trough a PR

.github/workflows/linuxWF.yml
Luthaf: I think the rpath was not properly loading dependencies, so I replaced it with an explicit `LD_LIBRARY_PATH` in the "run test" step below (line 188 of the file)

PR #1306, opened 2025-07-23
removing allocations from startwith

Iximiel:
##### Description

with these modifications I speed up the initialization, I am looking also at a solution that @gtribello mailed me, I will remove the draft after I tried to add that modification and verified that the tests passes

As you can see most of the time is passed in `CaseInsensitiveCompare` and in `startWith`:

before

sc_TORSION_128000nat_500steps_6threads

after:

sc_TORSION_128000nat_500steps_6threads_comp_stringview_findkw_only

The time of master are:
```
BENCH:  Kernel:      this
BENCH:  Input:       sc_cmTORSION_CPU_128000.dat
BENCH:                                                Cycles        Total      Average      Minimum      Maximum
BENCH:  A Initialization                                   1   571.519895   571.519895   571.519895   571.519895
BENCH:  B0 First step                                      1     0.024423     0.024423     0.024423     0.024423
BENCH:  B3 Calculation part 2                            499     8.278312     0.016590     0.016000     0.019506
PLUMED:                                               Cycles        Total      Average      Minimum      Maximum
PLUMED:                                                    1   579.820944   579.820944   579.820944   579.820944
PLUMED: 1 Prepare dependencies                           500     0.001118     0.000002     0.000001     0.000008
PLUMED: 2 Sharing data                                   500     0.275068     0.000550     0.000287     0.000830
PLUMED: 3 Waiting for data                               500     0.001679     0.000003     0.000002     0.000008
PLUMED: 4 Calculating (forward loop)                     500     7.945487     0.015891     0.015347     0.020815
PLUMED: 5 Applying (backward loop)                       500     0.051883     0.000104     0.000092     0.002660
PLUMED: 6 Update                                         500     0.018839     0.000038     0.000030     0.000087
```

```
BENCH:  Kernel:      this
BENCH:  Input:       sc_cmTORSION_CPU_128000.dat
BENCH:                                                Cycles        Total      Average      Minimum      Maximum
BENCH:  A Initialization                                   1   335.808614   335.808614   335.808614   335.808614
BENCH:  B0 First step                                      1     0.040812     0.040812     0.040812     0.040812
BENCH:  B2 Calculation part 1                            299     5.884897     0.019682     0.018033     0.023406
BENCH:  B3 Calculation part 2                            200     3.919282     0.019596     0.018236     0.022326
PLUMED:                                               Cycles        Total      Average      Minimum      Maximum
PLUMED:                                                    1   345.652010   345.652010   345.652010   345.652010
[...]

PR #1307, opened 2025-07-28
A faster parser

Iximiel:
##### Description

this is build on top of #1306, and can replace it.

Now the lines are pre-parsed with `TokenizedLine`, I did not used the `FastStringUnorderedMap` because as not I need this to work (at leas O(log(N)), and because accessing and deleting in a unordered_map is yes O(1), but deleting from a vector is O(N), so we lose the advantage while parsing. I may have a solution, but this is working now and it is WAAAAAAY much faster than before.

Also, I think I did not managed to solve the problem that in a vector read you can `@replicas:`-y the whole vector or the all the single elements but the first (because if you start with `@replicas:` it will flag the whole input as that). And I did not understand if it is an already present problem or something I just created, with the test I know that the new parser is compatible with that, but the solution is to change how the call of  getWords that does the first tokenization of the input line behaves.

@GiovanniBussi what do you think?
There are a few light changes into DEBUG, because it is the ONLY action I found with lowercase keys

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/regtest/basic/rt-make-getwords/output.reference
+++ b/regtest/basic/rt-make-getwords/output.reference
@@ -58,11 +58,11 @@ Replica: 2
 	3 
 "@replicas:{{1,2,3} {0,0,0} {2,4,6}}"
 Replica: 0
-	{1,2, 
+	{1,2,3} 
 Replica: 1
-	{0,0, 
+	{0,0,0} 
 Replica: 2
-	{2,4, 
+	{2,4,6} 
 "@replicas:1 2 3"
 Replica: 0
 	1 
--- a/src/tools/Tools.cpp
+++ b/src/tools/Tools.cpp
@@ -510,10 +510,14 @@ void getWords_replicas(gch::small_vector<std::string_view> & words,
       ++parlevel;
       if(parlevel == 1) {
         init = i+1;
+      } else {
+        ++size;
       }
     } else if(line[i] == closePar) {
       --parlevel;
+      ++size;
       if(parlevel == 0) {
+        --size;
         break;
       }
     } else {


src/cltools/Driver.cpp
GiovanniBussi, line 1: Why has this been moved?

src/generic/Debug.cpp
GiovanniBussi, line 1: I think flags should always be case insensitive. I would remove the special case for DEBUG

src/tools/TokenizedLine.h
```diff
@@ -0,0 +1,123 @@
...
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+   GNU Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public License
+   along with plumed.  If not, see <http://www.gnu.org/licenses/>.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ */
+#ifndef __PLUMED_tools_TokenizedLine_h
+#define __PLUMED_tools_TokenizedLine_h
+#include "Tools.h"
+#include <map>
```
GiovanniBussi, line 26: Why did you use a map here? An unordered map is significantly faster when using strings as keys

src/generic/Debug.cpp
Iximiel: ok, that should take a few minutes

src/cltools/Driver.cpp
GiovanniBussi: OK so this is technically a bug should be backported

GiovanniBussi:
>Also, I think I did not managed to solve the problem that in a vector read you can @replicas:-y the whole vector or the all the single elements but the first (because if you start with @replicas: it will flag the whole input as that). And I did not understand if it is an already present problem or something I just created, with the test I know that the new parser is compatible with that, but the solution is to change how the call of getWords that does the first tokenization of the input line behaves.

I do not understand. The expected behavior is that every time a value is read that begins with `@replicas:` it is parsed in that way. You can do this in two manners:
- The tokenizer (or the FastStringUnorderedMap) is assigned the full string. Later on when you access an element you check if it starts with `@replicas` and, in case, unfold it
- The tokenizer (or the FastStringUnorderedMap) is assigned directly the correct element

I expect them to be equivalent.

Notice that the performance issues are with searching a key over thousands, not when reading it. So, it's not really necessary to optimize that part of the code. It will be in any case executed N times, and not N*log(N) times

src/tools/TokenizedLine.h
GiovanniBussi: But that's exactly what is implemented in FastStringUnorderedMap. Is there a reason not to use it?

Iximiel:
> @Iximiel I am still confused: why didn't you use the `FastStringUnorderedMap`?

@GiovanniBussi I tried, but I did managed to get only a collection of sigsegv
So I decided to start by making this work with an ordered map because I did not need to rewrite the delete operation to delete also from the support vector (I commented the possible solution and the KISS one, that I realized a lot later..., deleting from the support vector slows downs things in any case, becasue I add N operations at the end o N o(log(N)) operation if i use the set and I delete while I am working, when this passes the CI I will make some measurements).
Changing to another map should be easy, as now I just wanted this to work with a reasonable speedup

---

on the other hand: Fedora is failing, but I can't access the log, can you?

src/tools/TokenizedLine.h
Iximiel: the problem is that the logic of the line is erasing the parsed kws,
so erasing from the support vector adds a lot of time because that is an array, pushing to it also adds time (push is O(1) until you need to resize)

Tomorrow I will change to the FSUM and take some measure with the multicolvars. I was thinking about deleting the array only in the end to avoid searching and deleting among its element (and writing strange logic)

GiovanniBussi: > the problem is that the logic of the line is erasing the parsed kws,
so erasing from the support vector adds a lot of time because that is an array, pushing to it also adds time (push is O(1) until you need to resize)

I know this, and indeed I think that's why you were replacing the implementation of FastStringUnorderedMap to use a `set`, or perhaps with an `unordered_set`. This is the right thing to do, no?

Iximiel: If we delete the elements of the unordered_set only at the end the uset might indeed be a good idea: it has the same problems of the unordered_map, but if we do not care and we do not access/find within it, we do not care of hashing its values and we live in the realm of the O(1) only for the insertions in the buckets
and then we delete in the dtor of the FSUM

Tomorrow i will definitively try that

GiovanniBussi:
Regarding the replica syntax, yes I confirm that the first brace is removed by the initial tokenizer, the one that reads the input, and should remane as it is.

In the optimization I think we should keep the current behavior, to avoid breaking input files.

Iximiel:
> To make sure I understand: in all cases where v2.10 works, your version gives the same result. In addition, you allow the replica syntax on individual components of a vector. This might make sense as an extension. Provided it literally breaks none of the existing input files

 - the replica working on the individual component is not a new feature, it is an existing feature I found, it is "tested" in `basic/rt-multi-1?`, and looking at the blame is there since 8 years ago, the first component not working was not working also before this PR
 - Since the test are passing, can I assume that the new parser parses the same things with the same result as before?

GiovanniBussi:
Ok thanks for the clarification so maybe you are right the current behavior is wrong... I will have a look at it since it is strange that we never noticed it

Iximiel:
> but it is slightly slower with the FSUM, it think because it is managing two maps at the same time (and erases through key instead of using an already found iterator - is "Amortized constant." for map vs "Average case: map.count(key), worst case: map.size()" for the unordered_map, looking at the manual)

I measured again, on the workstation, the version that uses map is sligthly faster that the one that uses the FSUM, branch [here](https://github.com/Iximiel/plumed2/tree/nit/longlonglog-dict-fsum), even with the erase that runs with the iterator

src/generic/Debug.cpp
```diff
@@ -64,8 +64,8 @@ void Debug::registerKeywords( Keywords& keys ) {
   Action::registerKeywords( keys );
   ActionPilot::registerKeywords(keys);
   keys.add("compulsory","STRIDE","1","the frequency with which this action is to be performed");
-  keys.addFlag("logActivity",false,"write in the log which actions are inactive and which are inactive");
-  keys.addFlag("logRequestedAtoms",false,"write in the log which atoms have been requested at a given time");
+  keys.addFlag("LOGACTIVITY",false,"write in the log which actions are inactive and which are inactive");
```
GiovanniBussi, line 67: @Iximiel here, for consistency with previous versions, it might be wiser to keep the current capitalization and then automatically make upper case all flags. I am saying this in case people use the new manual prepare an input for an old version. But perhaps it is irrelevant (we switched to case insensitive quite some time ago, in v2.6).

PR #1318, opened 2025-09-10
Parallel task manager accelerated as a plugin

Iximiel:
##### Description

**Note that this PR is built on top of #1092**

@GiovanniBussi, @gtribello, as we discussed last week I started this new approach to the PTM: the accelerated version is now served separately as a plugin.
This has a few advantages:
 - I can test the compilation on the CI without risking NVHPC blocking itself on some omp pragmas
 - You can use Plumed on the CPU compiled with gnu/clang along with the accelerated one
 - I have somehow implemented a mixed precision version of plumed (we have to discuss a few things about it)
 - starting from the overridden PTM in the plugin we can try different new acceleration strategies

I tried to avoid propagating the template declarations the more I could, but I did not always succeed

Looking at the diff some extra changes have slipped in, like the ones in check and some iostream inclusions.

This is only a WIP, only the standard multicolvars are included in the plugin, but I am working on expanding them

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/.github/workflows/linuxWF.yml
+++ b/.github/workflows/linuxWF.yml
@@ -146,18 +146,6 @@ jobs:
         echo "CXX=icpx" >> $GITHUB_ENV
         echo "CC=icx" >> $GITHUB_ENV
         echo "FC=ifx" >> $GITHUB_ENV
-    - name: Install NVHPC compiler
-      if: contains( matrix.variant, '-nvhpc-' )
-      # use this if it does not work
-      # wget https://developer.download.nvidia.com/hpc-sdk/24.3/nvhpc_2024_243_Linux_x86_64_cuda_12.3.tar.gz
-      # tar xpzf nvhpc_2024_243_Linux_x86_64_cuda_12.3.tar.gz
-      # nvhpc_2024_243_Linux_x86_64_cuda_12.3/install
-      run: |
-        curl https://developer.download.nvidia.com/hpc-sdk/ubuntu/DEB-GPG-KEY-NVIDIA-HPC-SDK | sudo gpg --dearmor -o /usr/share/keyrings/nvidia-hpcsdk-archive-keyring.gpg
-        echo 'deb [signed-by=/usr/share/keyrings/nvidia-hpcsdk-archive-keyring.gpg] https://developer.download.nvidia.com/hpc-sdk/ubuntu/amd64 /' | sudo tee /etc/apt/sources.list.d/nvhpc.list
-        sudo apt-get update -y
-        sudo apt-get install -y nvhpc-24-3
-      # nvhpcinstalls his own mpi compilers
     - name: Install MPI
       # install MPI at last since it modifies CC and CXX
       if: contains( matrix.variant, '-mpi-' )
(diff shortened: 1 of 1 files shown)


src/core/PlumedMain.h
```diff
@@ -535,6 +535,8 @@ class PlumedMain:
   void getKeywordsForAction( const std::string& action, Keywords& keys ) const ;
 /// Check if the input has been initialized
   bool hasBeenInitialized() const ;
+/// Returns the images from the dlloader
+  const std::vector <void*>& getDLHandles() const noexcept;
```
Iximiel, line 539: @GiovanniBussi I did this to use the action register's check with the extra handles within FunctionShortcut.
Is this a good idea?

Iximiel:
@gtribello @GiovanniBussi this is nearly done
I have to adjust a little the tests (that as now can only be run locally) but we can check the at least the compilation of the accelerated CV on the CI without disturbing the compilation of the main plumed library

src/core/PlumedMain.h
Iximiel: >     * I do not know how you use the result. This is supposed to be an internal detail, perhaps what you are doing can be directly encapsulated in an additional method of class PlumedMain

I use it in FunctionShortcut.h . By "encapsulated it as a method of PlumedMain", do you mean to create something like `plumed->check` instead of calling the action register singleton while exposing the void* vector?

>     * It is very difficult to see what this PR does because it is massive

I know, I should have moved the tests as the very last thing, because those are eclipsing the other files in the gh UI.
Apart from that, this branch is based on #1092, for the mixed precision things, so part of the changes can be discussed there

GiovanniBussi: >I use it in FunctionShortcut.h . By "encapsulated it as a method of PlumedMain", do you mean to create something like plumed->check instead of calling the action register singleton while exposing the void* vector?

Yes exactly. The implementation based on void* is quite tricky, so I would rather add an ad hoc method to do (what you need to do)

Iximiel:
@gtribello can you help me in getting right the documentation for the new shorcuts?

The Intel error is already corrected, I will push it with the documentation fix

gtribello:
Hi @Iximiel 

I had a look at this.  Can't get the code to run on my laptop any more because of an update, but I have a suggestion on how to solve the issue with the manual.  To be clear, the idea is that the DISSIMILARITY_CPU command would never be used in an input file.  The input would always be:

````
DISSIMILARITY ...
````

If that is the case, then you just need to add the command:

```c++
keys.setDisplayName("DISSIMILARITY");
```

in the `registerKeywords` method of the actions that have a _CPU version.  Adding this command hides the shortcut version from the manual builder.  I hope this works.

Iximiel:
> Hi @Iximiel
> 
> I had a look at this. Can't get the code to run on my laptop any more because of an update, but I have a suggestion on how to solve the issue with the manual. To be clear, the idea is that the DISSIMILARITY_CPU command would never be used in an input file. The input would always be:
> 
> ```
> DISSIMILARITY ...
> ```
> 
> If that is the case, then you just need to add the command:
> 
> ```c++
> keys.setDisplayName("DISSIMILARITY");
> ```
> 
> in the `registerKeywords` method of the actions that have a _CPU version. Adding this command hides the shortcut version from the manual builder. I hope this works.

Thanks! I'll try that with DISSIMILARITY and MATRIX_PRODUCT
I am testing with curl disabled in the manual to speed up things

On my PC it  DISSILMILARITY works, but MATRIX_PRODUCT(_CPU) not, it may be a question of getting the correct suffix?

Iximiel:
Hi @gtribello 

On my pc the manual problem looks solved using your first suggestion.

gtribello:
Oh jolly good @Iximiel.  It looks like it is solved here, too, as the doc-mpi is now building.

PR #1320, opened 2025-09-19
Added the "CUSTOM check" in switching functions

Iximiel:
##### Description

@GiovanniBussi, related to the bug in cosine we talked on Monday, I modified the test I added some times ago for the switching functions:
I added an extra check with the equivalent CUSTOM function for each of the test.
The fastgaussian gives some problem, I may have set the custom up wrongly.

##### Target release

I would like my code to appear in release __2.11__

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Very good! Do I understand correctly that this test verifies that derivatives are equivalent those obtained with CUSTOM?

When you say that there is a problem in fastgaussian, do you mean that there might be a bug in the evaluation of that switching function?

Iximiel:
> Very good! Do I understand correctly that this test verifies that derivatives are equivalent those obtained with CUSTOM?

Yes, it verifies that both the value and the derivatives against CUSTOM

> When you say that there is a problem in fastgaussian, do you mean that there might be a bug in the evaluation of that switching function?

I think that there is a problem in CUSTOM when `r==0` for `FUNC=exp(-x^2/2) R_0=1.0 D_0=0.0 D_MAX=2.6` (remember that $x=\frac{r-d_0}{r_0}$)
but there is also the non zero chance that `GAUSSIAN R_0=1.0 D_0=0.0 D_MAX=2.6` and `CUSTOM FUNC=exp(-x^2/2) R_0=1.0 D_0=0.0 D_MAX=2.6` are not equivalent because of my skill issues

It gives the problem on the derivative of r==0.0 only when calling for "CalculateSquared" by looking at the tests

---

I need to add also the cubic switch, that I forgot for some reason, when I added the first version of this test...

Iximiel:
@GiovanniBussi 
The bug was in the calculateSqr implementation of the fastgaussian: the derivative was wrong.
Now with the new test layout it is easier to understand it:
I corrected the bug in the latest commit, with a comment about it

PR #1327, opened 2025-10-03
Small performance optimisation in transferring matrices between stash and value

gtribello:
##### Description

When you were transferring the matrices between the Value and stash at the start of the parallel task manager a lot of time is spent transferring zeros.  With this change we avoid spending that time on transferring zeros.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/ActionWithMatrix.cpp
+++ b/src/core/ActionWithMatrix.cpp
@@ -98,30 +98,40 @@ void ActionWithMatrix::updateBookeepingArrays( RequiredMatrixElements& outmat ) 
 }
 
 void ActionWithMatrix::transferStashToValues( const std::vector<double>& stash ) {
+  unsigned ncomp = getNumberOfComponents();
   unsigned ncols = getPntrToComponent(0)->getNumberOfColumns();
   unsigned m=0, n=0, nrows = getPntrToComponent(0)->getShape()[0];
   for(unsigned i=0; i<nrows; ++i) {
-    for(unsigned j=0; j<ncols; ++j) {
-      for(unsigned k=0; k<getNumberOfComponents(); ++k) {
-        getPntrToComponent(k)->set( m, stash[n] );
-        n++;
-      }
-      m++;
+    unsigned ncr = getPntrToComponent(0)->getRowLength(i);
+    for(unsigned k=1; k<ncomp; ++k) {
+      plumed_assert( ncr == getPntrToComponent(k)->getRowLength(i) );
     }
+    for(unsigned j=0; j<ncr; ++j) {
+      for(unsigned k=0; k<ncomp; ++k) {
+        getPntrToComponent(k)->set( m+j, stash[n+j*ncomp+k] );
+      }
+    }
+    n += ncols*ncomp;
+    m += ncols;
   }
 }
 
 void ActionWithMatrix::transferForcesToStash( std::vector<double>& stash ) const {
+  unsigned ncomp = getNumberOfComponents();
   unsigned ncols = getConstPntrToComponent(0)->getNumberOfColumns();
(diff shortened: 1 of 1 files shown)


gtribello:
Hi @Iximiel 

Can you try to make the flame graph that you sent me earlier with this branch.  I hope that this should address some of the slow bits of code you were seeing.

Thanks

gtribello:
OK @GiovanniBussi and @Iximiel 

I put a sanity check inside a ifndef NDEBUG to hopefully get a little more performance from this change.  If it passes now (which is should) we should be OK to merge.

gtribello:
Don't merge this yet @GiovanniBussi .  I have another idea to resolve the issue that @Iximiel is raising with the flame graphs.  I will try to implement and will push the result here.  Thanks.

gtribello:
Hi @Iximiel 

Could you try that calculation again and create a new flame graph using this updated version, please? Now, the amount of time spent in the transferStashToValue functions should be significantly less.

Also, just remind me, how many OpenMP threads are you running the calculations on for figures 8 and 15?  

Thanks

gtribello:
Hi @Iximiel 

It might also be worth examining how well the Q6 calculations perform when you attempt to run them on the GPU now as well.  This would not be for the paper.  That push_back was probably really impacting the speed for all the test calculations we did.

gtribello:
Hi @Iximiel 

As you can see, I merged the changes that you made into here.  Thanks for looking at this and optimising a few things.  

There are perhaps more things to be done to optimise the code, but we have to stop at some point and finish the tutorial paper.  We can discuss this more with the others next Monday, but I would like to have the paper submitted by the end of October.  My preference is thus to stop working on optimising the code until the paper is done.  We then need to run all the calculations for the paper with this new and faster version of the code to get updated timings for actions.

I would hold off re-running most of the calculations until Monday, after we have had that discussion about what to do with the paper.  However, it would be great if you could regenerate Figure 15 using this new version of the code.  To finish the first draft of the text, I need to know if the work we have done on optimising the code has ensured that there is any size of sphere where using the mask is faster than not using it.  If I know the answer to this question, I can then complete the text of the first draft.  

My question is thus: is it possible to get a new version of Figure 15 before Friday?  If I can get that figure by Friday, I can then finish the text of the paper so we have a first draft for next week.

src/core/Value.cpp
```diff
@@ -277,7 +274,24 @@ std::size_t Value::getIndexInStore( const std::size_t& ival ) const {
...
+  const auto base = task * getNumberOfColumns();
+  if (hasDeriv) {
+    for(unsigned k=base; k<base+ncol; ++k) {
+      if(std::fabs(data[k*(1+ngrid_der)])>0.0) {
+        return true;
+      }
+    }
+    return false;
+  }
+  return std::any_of(&data[base],&data[base+ncol],[](double x) {
```
Iximiel, line 288: The funny thing is the AI complains about the commented out code, but not at the very nasty bug that I proudly produced :wink:

```suggestion
  return std::any_of(&data[base],&data[base]+ncol,[](double x) {
```

@gtribello if you accept this suggestion trough the GH UI we solve the bug that the debug runs are catching

Fun Fact" compiling plumed with -g and running the tests cost you ~50 GB of space

Iximiel:
> I would hold off re-running most of the calculations until Monday, after we have had that discussion about what to do with the paper. However, it would be great if you could regenerate Figure 15 using this new version of the code. To finish the first draft of the text, I need to know if the work we have done on optimising the code has ensured that there is any size of sphere where using the mask is faster than not using it. If I know the answer to this question, I can then complete the text of the first draft.

@gtribello, by looking at the figures 8 and 15 your idea is working quite well now: in the 8 the y axis shifted lower, and in the 15 we can finally see what you were expecting

PR #1336, opened 2025-10-17
Removing warnings from the always-on modules

Iximiel:
##### Description

I am trying to avoid to compile plumed with warnings. I am starting from the internal `always` modules, then to the `default-on` modules.
I forced the `-Werror` flag in the Makefile core and tools with the basic warnings, like in the CI, plus  -Wshadow (because it made me find a suspect thing that I showed in #1335, and it helps being more intentional in functions and methods)

By rereading part of the code I fond a few things that merit some attention: I started to flag them with a comment starting with `//NOTE:`. I will open some mini-review for those bits of code here in GH

`Werror` breaks intel because it wants `-qopenmp` instead of `-fopenmp`, but I think I can find a solution (in configure.ac) for that quickly.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


src/tools/RMSD.cpp
```diff
@@ -124,27 +135,30 @@ const std::vector<Vector>& RMSD::getReference() const {
...
-  unsigned n=reference.size();
-  plumed_massert(this->align.size()==align.size(),"mismatch in dimension of align/displace arrays");
-  this->align=align;
+void RMSD::setAlign(const std::vector<double>& alignTo,
+                    const bool normalize_weights,
+                    const bool remove_center) {
+  const unsigned n=reference.size();
+  //NOTE:QUESTION HERE: I am looping on the n from the reference, but I am checking that alignTo and align have the same size, not that reference and align have the same size!!!!
+  plumed_massert(align.size()==alignTo.size(),"mismatch in dimension of align arrays");
+  align=alignTo;
```
Iximiel: In the body of the function below I am looping on `align` with this `n` taken from `reference`, but here I am checking that the input `alignTo` and the member `align` have the same size, not that `reference` and `align` have the same size!!!!

Same question for setDisplace

src/tools/TypesafePtr.h
```diff
@@ -71,34 +71,34 @@ class TypesafePtr {
...
-    SizeLike(long int size): size(std::size_t(size)) {}
-    SizeLike(long long int size): size(std::size_t(size)) {}
+    SizeLike(short unsigned dim): size(dim) {}
+    SizeLike(unsigned dim): size(dim) {}
+    SizeLike(long unsigned dim): size(dim) {}
+    SizeLike(long long unsigned dim): size(dim) {}
+    SizeLike(short dim): size(std::size_t(dim)) {}
+    SizeLike(int dim): size(std::size_t(dim)) {}
+    SizeLike(long int dim): size(std::size_t(dim)) {}
+    SizeLike(long long int dim): size(std::size_t(dim)) {}
```
Iximiel: this struct appears in some other places with the very same code

Iximiel:
@GiovanniBussi I "solved" (in heavy quotes) the problem with the of `crystdistrib/rt-bops-shortcut` by reducing the forces precision to 2 digits
Is that ok?

PR #1346, opened 2025-11-17
[Not to be merged] - NL with the linked cells algorithm

Iximiel:
@GiovanniBussi, @carlocamilloni, @gtribello
As we discussed this morning: this is the work in progress 

I will probably separate some of the commits about the test organizations in a standalone PR because can exists on their own.

You can see that some of the test that are using the NL with two lists are failing now (I suspect that is because the list with cells is sighthly bigger)

I changed the LC to return the list of elements in cells, so that the same cell configuration can be applied on two lists of atoms like in the loop I am using  for the NL with two list.
This does not change the interface as it used in other CV


Iximiel:
#1354 will be used instead of this

PR #1348, opened 2025-11-17
Multiple updates to metatomic integration

### LINKED ISSUE #1337: METATOMIC: SELECTED_ATOMS keyword does not work as intended
In the [documentation](https://www.plumed.org/doc-v2.10/user-doc/html/_m_e_t_a_t_o_m_i_c.html), it says that the ```SELECTED_ATOMS``` keyword can be used as ```SELECTED_ATOMS=11-20``` just like the atom selector but actually it is [parsed as a vector](https://github.com/plumed/plumed2/blob/527744f9371afd1ae5365db43f0a5312b763bc8b/src/metatomic/metatomic.cpp#L533) and thus must can only be written in expanded form e.g.  ```SELECTED_ATOMS=11,12,13,14,15,16,17,18,19,20```

Luthaf:
##### Description

Fix #1337 and update the code to work with the latest releases and use the latest utilities. 

The main change is added support for multiple variants of the same output in one model, i.e. having a single file that can produce any of `features/soap`, `features/dihedral`, `features/my_custom_thing`, and pick one at runtime.

##### Target release

I would like my code to appear in the next patch release for v2.10

##### Type of contribution

##### Copyright

##### Tests


Luthaf:
Finally got CI to pass, this took a while … Let me know if this is good on your side as well!

Issue #1358, opened 2025-12-12
plumed/wrapper/Plumed.h:3132:10: error: 'unique_ptr' in namespace 'std' does not name a template type (plumed 2.10.0)


truatpasteurdotfr:
I am trying to build openmm-plumed (git) for x86_64/debian13/openmm-8.4.0/cuda-13.1/plumed-2.10.0 which is failing with:

```
[ 44%] Building CXX object CMakeFiles/OpenMMPlumed.dir/serialization/src/PlumedForceProxy.cpp.o
In file included from /opt/openmm-plumed/openmmapi/include/internal/PlumedForceImpl.h:38,
                 from /opt/openmm-plumed/openmmapi/src/PlumedForceImpl.cpp:32:
/usr/local/include/plumed/wrapper/Plumed.h:3132:10: error: 'unique_ptr' in namespace 'std' does not name a template type
3132 |     std::unique_ptr<char[]> dynamic_buffer;
      |          ^~~~~~~~~~
/usr/local/include/plumed/wrapper/Plumed.h:1801:1: note: 'std::unique_ptr' is defined in header '<memory>'; this is probably fixable by adding '#include <memory>'
1800 | #include <array> /* array */
  +++ |+#include <memory>
1801 | #include <initializer_list> /* initializer_list */
In file included from /opt/openmm-plumed/openmmapi/include/internal/PlumedForceImpl.h:38,
                 from /opt/openmm-plumed/openmmapi/src/PlumedForce.cpp:34:
/usr/local/include/plumed/wrapper/Plumed.h:3132:10: error: 'unique_ptr' in namespace 'std' does not name a template type
3132 |     std::unique_ptr<char[]> dynamic_buffer;
      |          ^~~~~~~~~~
/usr/local/include/plumed/wrapper/Plumed.h:1801:1: note: 'std::unique_ptr' is defined in header '<memory>'; this is probably fixable by adding '#include <memory>'
1800 | #include <array> /* array */
  +++ |+#include <memory>
1801 | #include <initializer_list> /* initializer_list */
make[2]: *** [CMakeFiles/OpenMMPlumed.dir/build.make:79: CMakeFiles/OpenMMPlumed.dir/openmmapi/src/PlumedForce.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[...]

PR #1361, opened 2026-01-14
Changed the order of the includes and when include <memory> in the wrapper

Iximiel:
##### Description

Following the problems raised in #1358 I made a few modifications in Plumed.h, these changes solve the issue:

since some `unique_ptr` capabilities are used if the compiler supports c++11 even if `__PLUMED_WRAPPER_LIBCXX11` is not activated, I moved the  `#include <memory>` in the `#if __cplusplus > 199711L` include block.

I removed the `#include <memory>` from the C++17 include block, since if `#if __cplusplus >= 201703L` passes the `#if __cplusplus > 199711L` passes too.

The movement of the includes is just an aesthetic choice, I can revert to the previous order if you prefer.

##### Target release

I would like my code to appear in release __v2.10+__

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/wrapper/Plumed.h
+++ b/src/wrapper/Plumed.h
@@ -1777,29 +1777,29 @@ __PLUMED_WRAPPER_EXTERN_C_END /*}*/
 #include <new> /* bad_alloc bad_array_new_length (C++11) */
 #include <typeinfo> /* bad_typeid bad_cast */
 #include <limits> /* numeric_limits */
+//C++11 functionalities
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
@Iximiel do you think this means we can close #1358 then?

Iximiel:
> @Iximiel do you think this means we can close #1358 then?

@GiovanniBussi yes

PR #1363, opened 2026-01-21
Making the CI compatible with a maintained python version (Reprise of #1350)

Iximiel:
>[!NOTE]
>I will likely reopen this as a standard PR soon I have finished
>The original PR is #1350 

I am doing some researches on testing with python 3.10.

Since I am getting a segmentation fault, following [this blogpost](https://blog.xmatthias.com/post/pytest-debug-segfault/), I started trying to use `pytest --collect-only` to have a list of tests and then "bisect" it to find the culprit.

The problem is that instead of getting the list of error I get a segmentation fault...

So I dug a little by manipulating the test collection. Turns out that if I hide `test_input_builder.py` I do not get the segmentation fault in the collection phase (I still get it while running the tests, but that is a _future me_ problem for now). This is strange because in #1350 it works (at least not in dbg mode).

I did some "manipulation" to `test_input_builder.py` and the culprit for the segmentation fault seems to be the initialization of the `plumed.InputBuilder`.

And this Is what I get when I try to run something similar directly in the interpreter:
```
>>> import plumed
>>> ib = plumed.InputBuilder()
+++ Loading the PLUMED kernel runtime +++
+++ PLUMED_KERNEL="/---/installdir/lib/libplumedKernel.so" +++
```

funnily this one below does not segfault, it loads the kernel twice (one for plmd, and one for guessing where to load the vim syntax file in ib) but goes smoothly
```python
import plumed
plmd = plumed.Plumed()
ib = plumed.InputBuilder()
```

---

###EDIT

I am breaking this PR in multiple more focused pr:

- [ ] #1366
- [ ] #1367

### DIFF
--- a/python/plumed.c
+++ b/python/plumed.c
@@ -1606,13 +1606,12 @@ struct __pyx_memoryviewslice_obj;
  * 
  * cdef class Plumed:             # <<<<<<<<<<<<<<
  *      cdef cplumed.plumed c_plumed
- *      cdef int initialized
+ *      #cdef int initialized #not used
 */
 struct __pyx_obj_6plumed_Plumed {
   PyObject_HEAD
   struct __pyx_vtabstruct_6plumed_Plumed *__pyx_vtab;
   plumed c_plumed;
-  int initialized;
 };
 
 
@@ -1726,7 +1725,7 @@ struct __pyx_memoryviewslice_obj {
  * 
  * cdef class Plumed:             # <<<<<<<<<<<<<<
  *      cdef cplumed.plumed c_plumed
- *      cdef int initialized
+ *      #cdef int initialized #not used
 */
 
 struct __pyx_vtabstruct_6plumed_Plumed {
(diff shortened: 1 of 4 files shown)


GiovanniBussi:
Do you mean it is a bug in the python interpreter? I think it should not happen

Iximiel:
On my pc I get the error
```python
ib=plumed.InputBuilder()
p=plumed.Plumed()
```
but not if I setup the IB like this.
```python
p=plumed.Plumed()
ib=plumed.InputBuilder()
```

In the CI this seems not to be a problem. On my laptop with a cranky WSL I do no get the error even without the declared plumed in the tests. So the `os.environ` error it may be and edge case of what my workstation does when calling plumed_finalize? I still can't explain the one we get in the CI in `File "/opt/hostedtoolcache/Python/3.10.19/x64/lib/python3.10/unittest/case.py", line 549 in _callTestMethod`.

PR #1366, opened 2026-02-02
_guessplumedroot now deletes the temporary file used to store the plumed output

Iximiel:
##### Description

In the [mkstemp documentation](https://docs.python.org/3.9/library/tempfile.html#tempfile.mkstemp) there is the following sentence:
> Unlike TemporaryFile(), the user of mkstemp() is responsible for
> deleting the temporary file when done with it.

So I changed the routine to use the higher level NamedTemporaryFile, to it will automatically delete the tmp file every time I run some test (or the user tries to create a new input with the InputBuilder)

##### Target release

I would like my code to appear in release __v2.9__

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/python/plumed.pyx
+++ b/python/plumed.pyx
@@ -708,21 +708,21 @@ def _guessplumedroot(kernel=None):
        dir from there.
     """
     try:
-        import tempfile
-        log=tempfile.mkstemp()[1]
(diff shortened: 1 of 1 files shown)


GiovanniBussi:
LGTM!

PR #1376, opened 2026-03-09
Fixed CONVERT_TO_FES to make it more back compatible with earlier versions of PLUMED. 

gtribello:
##### Description

CONVERT_TO_FES is now a shortcut action in PLUMED 2.10 that uses CUSTOM to convert to take the negative logarithm of a histogram and to multiply that by k_B T.  However, I made a mistake when adapting the MINTOZERO command to work with the new version of PLUMED.  To be clear, the shortcut that does this calculation works like this:

````
# Use -k T log P(s) to get the free energy from the histogram (h)
fu: CUSTOM ARG=h FUNC=-2.5*log(x) PERIODIC=NO
# Find the location of the minimum in the free energy surface that we have just calculated
m: FIND_GRID_MINIMUM ARG=fu 
# Subtract the minimum in the free energy from the grid so that the minimum has a free energy of zero.
f: CUSTOM ARG=fu,m.optval FUNC=x-y PERIODIC=NO
````

Because I am too clever for my own good, FIND_GRID_MINIMUM uses spine interpolation and a conjugate gradient algorithm to find the coordinates of the actual minimum of the free energy surface.  However, the earlier implementation of MINTOZERO just searched the value of the free energy on the various grid points, found the lowest and used that as the origin.

With this change, I have thus adjusted the shortcut so that if you use the MINTOZERO option, you just find the lowest free energy point on the grid and use that for the position of the origin.  The behaviour with this option thus recovers what was done in old versions of the code.

If you want to do the actual minimisation, you can use the FINDMINTOZERO, which does the conjugate gradient minimisation.

In rt-fes-periodic there were two CONVERT_TO_FES commands that use the MINTOZERO option.  I changed one of these commands to FINDMINTOZERO (this is the one where the reference is unchanged).  For the other, I kept it as MINTOZERO and checked that the values that you get out with this option are the same as what they would be in older versions of PLUMED.  I thus ensured that we are properly maintaining back compatibility.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/dimred/ArrangePoints.cpp
+++ b/src/dimred/ArrangePoints.cpp
@@ -338,7 +338,7 @@ void ArrangePoints::optimize( std::vector<double>& pos ) {
   if( mintype==conjgrad ) {
     int code = mycgminimise.minimise( cgtol, pos, &ArrangePoints::calculateFullStress );
     if( code>0 ) {
-        error("failure in conjugate gradient minimisation"); 
+      error("failure in conjugate gradient minimisation");
     }
   } else if( mintype==pointwise ) {
     unsigned nvals=getPntrToArgument( dimout )->getShape()[0];
--- a/src/dimred/ProjectPoints.cpp
+++ b/src/dimred/ProjectPoints.cpp
@@ -238,7 +238,7 @@ void ProjectPoints::getProjection( const unsigned& current, std::vector<double>&
   }
   int code = myminimiser.minimise( cgtol, point, &ProjectPoints::calculateStress );
   if( code>0 ) {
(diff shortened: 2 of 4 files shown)


gtribello:
@GiovanniBussi 

This works by finding the grid point where the free energy has the lowest value.  I then do the CG starting from there.  This is what happens if you use the  FINDMINTOZERO option

I wanted to make this change because Omar had a crash when he used the MINTOZERO option because the CG wasn't converging.  This wouldn't have happened in old versions of the code because you are not doing the CG.  I thus thought that introducing the new flag for turning this option makes sure people know what they are doing.

Issue #11, opened 2013-07-26
PCA colvars


GiovanniBussi:

GiovanniBussi:
Implemented in a507eb51813638e70629482216b099235da260f8

PR #1385, opened 2026-03-26
TT-MetaD

strandn:
##### Description

This pull request involves an implementation of tensor train (TT)-metadynamics (TT-MetaD). Details of the method can be found in https://arxiv.org/abs/2603.13549. The relevant code is in the "ttsketch" module (src/ttsketch). Instructions for setup and compilation are written in at the top of src/ttsketch/TTMetaD.cpp.

##### Target release

Latest possible version (master/v2.11?).

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
thanks @strandn! While we wait for the workflow to be approved I would like to ask you some cosmetic changes, (we need to update this part in the developer manual...):
 - You should add a few files (see for example metatomic):
   - you should move the installation instructions in a `module.md`
   - you should create a `module.yml` with some information about you and your TT-metad
 - Plumed has a "cite" function you can use to cite your paper and your sources (like ITensor and so on): `this->cite(ref)` in the ctor of the action
 - be prepared to write `std::vector` a lot. You are using `using namespace std;` and codecheck will not pass it
 - we will probably ask you to modify the configure.ac with the necessary settings to setup plumed with ITensor and add a small script to install it on the CI, so that your TT-metad can be tested there. I'll help you with those, if needed

src/ttsketch/BasisFunc.cpp
```diff
@@ -0,0 +1,322 @@
...
+    return result;
+  }
+}
+
+double BasisFunc::fourierd(double x, int pos) const {
+  if(pos == 1) {
+    return 0.0;
+  }
+  // k is the same harmonic index as in fourier()
+  double k = floor(0.5 * pos);
```
Iximiel: So, you need need this floored!
```suggestion
  double k = pos >> 1;
```
see [this](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TA3cAGQImbAA5PwAjbFIQAE5yAAd0JWIXJi9ffwMUtOchELDIthi4xIdsJwyRIhZSIiy/AJ57bEcCplr6oiKI6NiE%2BzqGppzWpRHe0P7SwfiASnt0H1JUTi5LAGZQ1F8cAGozLY9UNhYiJGPcMy0AQVu7zBWoxgOk0nR1pSUAfWJsGwIM8fK9sAd1OQQWCDgBrBbHGz3R7Qt6CVYEWKYYEvN6Qg6hIjvVILI4AdiRdwO1IOqPBsOOABEDjR6OgyBA8gAqLQAOgArAitpSaQdSNgiKsmO9Pt8/gCgZD4YjHmYyYzHlwlvRuPz%2BAEuDpyOhuB5bLYDkoVmtwds%2BOQiNotUtYSB%2BVpDNxpPw2G6PQajSauPwlCAPY7DVryHBYCgMDh8MQyJRqHRGKwONx7YJhGIJJwZHJhMo1JpI%2BR9BZDMZTObrG0Ohk3ExPN5moEW30SmUhnl0kIxi1kql%2B0wuwNyg2qp1uqM2%2BMp9UhLPpsUJ0NJj1BwZNw1x3Nyksrat1txtrt9raTmcLlctjdkfc6TKvtgfv8iICcaC8VDcfShUpFF/xZa1MVIbFn3xQliSUUk1RFGlnwZLZmTya5rh4FV7lFcVJVIaUPlfd8FQgJVANVdVNW1XV9SdY1TTrKxLWtdYjgsLYLH4CMdAWJYkGwFgcDiCAaK4b1yF9d1yADfggxDMMHSdPjyFdaR%2BV5AA2eIeHiLRNK0LR%2BQsCxdK2T0uC2Ojy3kpTIyWGNkDQdA2CSBhYhTCAMFc9y4mAHhTIEBhP1IUMICieiolCeoAE8s34KLWFIGKAHkol0ad4vIbyOGEFKmHoOLyxwKIfGADwJHoUNeH4HBzhMSRioIcVqgANzfejsHUKofE/LLCXaej6AIKJSFirwcHoohSAIX0avIdrSCiVJsEZQEa2GkxlIEIxgCUAA1TEAHcUqSZgspzURxEkQtLpLDR6P0VojC28xrFsQwRtDSAlnQJJOmqgBaFKrONRaZpwb7RMXTpm1bbIh2CGZu0GVo%2B06bc0ZHTp9x7CZ2mnGopkxmGiZ6XHUeGLd5yHXdV1mPGj1YgsxL1GT6KDCEAA5NMBzTpAOYBUFQA4At5HgDggRMSFIdith4BZuO2iA4xctzGAoKgvLV3yQH8wK0xCsKIvLRLYqys3krSjKnCynLmCIfLCvokqyoq%2BgqqyuqNo2I1CBa5x2uqo0up6vr5oGnVy2G0bxqwX3uJmub7UW5aVDW%2BrgE20B7J2lg9sO7ATrOg1s3kPMbtkO6VAe8tK2rV6mM%2BqIod%2B/6MiBkG5PBghIfgI8CaXAIIHcEmkbXA9cmxjISfRjIKbifHG2XYmaZ3QeZymBed1XhGd/J5H1wV5YT04CxWeswNuG53n%2BcF4XRYscXJelsg5fPpXc5VkBCBoGhPMuhXAsVd5D3TLEaegSBQzPSgb/GgRAYpnUUpAlQ6hposD%2BrDYM5BSBQJAFWXBSg4EIKQR6QhXV0GYKbMGC%2B7MbLcEZAQP%2BBxC5HViDfPmAshYizFhLYGWwDitSUBwu%2B3DH7P34Z/Xi/FBLCWoC6N00heRbC0FzLQ8QtjxHiNIMkZI1Fkk0hZCSvoeCGToVfbB0C7LSIslxcxcluBSOdAtWIaRXDSCAA%3D) clang is smarter and there is nearly no difference, but gcc produces quite a lot more code.

Iximiel:
@strandn 

The ci will not work until we setup the installation process for ITensor.

please run make astyle 

First of all:

> You should add a few files (see for example metatomic):
> 
>     * you should move the installation instructions in a `module.md`
> 
>     * you should create a `module.yml` with some information about you and your TT-metad

You should start by adding in the .ci directory a script for installing ITensor (see the arrayfire one), maybe write an ad-hoc options.mk for that. Then you should make the linux runner of the ci call it in the `-mpi-` jobs (you'll need to correct the varous /path/to).
A possible step to add (below the pytorch one) is the following (untested):
```yaml
    - name: Configure ITensor
      if: ${{ contains( matrix.variant, '-mpi-' ) }} 
      run: |
        echo "LDFLAGS=$LDFLAGS -L/path/to/ITensor/lib" >> $GITHUB_ENV
        echo "CPPFLAGS=$CPPFLAGS -I/path/to/ITensor -DITENSOR_USE_HDF5 -DITENSOR_USE_OMP -D__ITENSOR_LAPACK_WRAP_h" >> $GITHUB_ENV

        #temporary, the following lines will be addresses in configure.ac
        echo "CXXFLAGS=\"-O3 -fconcepts\"" >> $GITHUB_ENV
        echo "LIBS=\"-litensor -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -lhdf5 -lhdf5_hl\"" >> $GITHUB_ENV
        # this will be uncommented after the configure.ac modifications
        #echo "PLUMED_CONFIG=$PLUMED_CONFIG --enable-libitensor" >> $GITHUB_ENV
```
Probably the mkl libraries will need to be removed from the CI, since those are configuration dependent.

Tell me if you need any help to setup this

Edit:
https://rhysd.github.io/actionlint/ is a good tool for checking if the workflow is setup in a semantically correct way

strandn:
@Iximiel First of all, thank you very much for your messages and your assistance so far. I am very sorry for the inconvenience. This is my first time dealing with git rebasing. I think the culprit of the issue is the fact that I have two separate local "ttsketch" branches syncing to the same remote (GitHub) branch (one in my personal laptop and the other in my university's cluster). A few days ago, I successfully rebased (I thought) from one of my local branches but since then, blindly did `git pull` from the other local branch. That seems to have done a merge, not a rebase. This created a merge commit that brought back my entire commit history (the 800+ commits I thought I'd squashed away). When I pushed that merge commit, it rewrote the remote and my squash, and seemed to have resulted in some very undesired behavior (such as my branch deviating from plumed2/master in unexpected ways). I will be more mindful in the future. Let me roll back some changes to recover a recent stable version. Again, apologies.

Iximiel:
@strandn do not worry, git is a difficult but necessary evil (see the famous readme of the first commit of git for the opinion of Linus).
I get in similar situations often. Force pushing a branch cause problems if is synced on other computers. In your case a good practice could be to treat the ttsketch branch as the "Truth", and try things on a secondary branches on each machine that will be merged onto ttsketch when you are satisfied with the changes, or simply throw them away when you do not like the new things. Trust me on this: it will save you some scares and headaches! The extra branches can stay local unless needed on the other computer.

A good idea for moving things around could be to trying to use a TUI or GUI for git, I use lazygit, but there are good free addons for vscode and the usual gitk.

Before rebasing/resetting or any other thing that change the history, remember to backup the branch you are going to work with by creating a new one, that will also help in case of missteps. Sorry for repeating this.

Iximiel:
Hi @strandn, we are in the right direction :D 

the `install.itensor` script seems ok, but I need you to give permissions to the `install.itensor` script (`chmod +x` should be enough) and commit that change, so that the CI will run it.

When this passes the CI I will help you in modifying configure.ac

For the errors in codecheck, I suggest you to do `make codecheck` in the ttsketch directory to have a local check there (if you run it in the main directory will take more that 1 hour, locally will be a lot shorter), so you can check that you solved them without rerunning the CI and waitng for 90 minutes.

As usual if you have some doubts, ask questions!

carlocamilloni:
The code is still not compiling because of missing include files, please verify (hdf5 stuff)

Iximiel:
@strandn create a personal git repository on github with the installer for itensor and a very simple workflow that runs just that it in the same environment of the plumed CI to check if it is working.
In this way you will be faster at checking if the installer work and you do not need us to activate the CI for you at each iteration.

If you are slightly blocked, please ask us for direction.

.github/workflows/linuxWF.yml
```diff
@@ -68,6 +68,23 @@ jobs:
...
+      if: ${{ contains( matrix.variant, '-mpi-' ) }}
+      run: |
+        .ci/install.itensor
+    - name: Configure ITensor
+      if: ${{ contains( matrix.variant, '-mpi-' ) }}
+      run: |
+        ITENSOR_DIR="$HOME/opt/itensor"
+        HDF5_CFLAGS=$(pkg-config --cflags hdf5-serial 2>/dev/null || pkg-config --cflags hdf5 2>/dev/null || echo "-I/usr/include/hdf5/serial")
+        HDF5_LIBDIR=$(pkg-config --variable=libdir hdf5-serial 2>/dev/null || pkg-config --variable=libdir hdf5 2>/dev/null || echo "/usr/lib/x86_64-linux-gnu/hdf5/serial")
+        echo "LDFLAGS=$LDFLAGS -L${ITENSOR_DIR}/lib -L${HDF5_LIBDIR}" >> $GITHUB_ENV
```
Iximiel: Try this:
```suggestion
        HDF5_LIBDIR=$(pkg-config hdf5-serial --libs-only-L 2>/dev/null || pkg-config hdf5 --libs-only-L 2>/dev/null || echo "-L/usr/lib/x86_64-linux-gnu/hdf5/serial")
        echo "LDFLAGS=$LDFLAGS -L${ITENSOR_DIR}/lib ${HDF5_LIBDIR}" >> $GITHUB_ENV
```

Iximiel: I saw that the -L of hdf5 does not get set correctly, hopefully this will solve the problem

strandn:
@Iximiel Thank you again for all your help so far, once again. It seems we are moving in the right direction. I was curious to see if I could try to fix the workflow issues myself. As a preliminary step, I wanted to make sure I can still compile my local "ttsketch" branch on my local machine, using the same commands I have always been using:
`./configure --enable-modules=ttsketch LDFLAGS=-L/home/nstrand/ITensor/lib CPPFLAGS="-I/home/nstrand/ITensor -DITENSOR_USE_HDF5 -DITENSOR_USE_OMP -D__ITENSOR_LAPACK_WRAP_h" LIBS="-litensor -lmkl_intel_lp64 -lmkl_sequential -lmkl_core -lhdf5 -lhdf5_hl" CXXFLAGS="-O3 -fconcepts;
make -j 4`
But unfortunately, I now unexpectedly get the following error message:
`ActionPilot.cpp: In constructor ‘PLMD::ActionPilot::ActionPilot(const PLMD::ActionOptions&)’:
ActionPilot.cpp:30:7: error: ‘*(PLMD::Action*)((char*)this + 16).PLMD::Action::keywords’ is used uninitialized [-Werror=uninitialized]
   30 |   if( keywords.exists("STRIDE") ) {
      |       ^~~~~~~~
cc1plus: all warnings being treated as errors
make[6]: *** [../maketools/make.rules:16: ActionPilot.o] Error 1
`
Keep in mind I have synced (pulled) a number of changes made by others in the main plumed2/master branch before attempting this compile. Do you have any intuition for how to resolve this? Thank you very much in advance.

Iximiel:
The CI compiles, there is just a very tiny _runtime_ error in ARG: I published a comment directly in the the text to address it.

I am not sure I can help on the error you get locally from ActionPilot: `-Werrors` is active only in core modules and not in contributed modules, so it should not be your fault.
`keywords` is initialized in the `Action` ctor, so it might be a false positive (unless is a local merge problem, since on the CI compiles smoothly). What is your compiler?


Edit:
I think that when this little problem with ARG is solved we can finally see the test running.
Then we will start with configure.ac and configure (I need you to have autoconf exactly at the version 2.69, you can get the install script directly from the .ci directory, if you need it

src/ttsketch/TTMetaD.cpp
```diff
@@ -0,0 +1,1086 @@
...
+  void calculate() override;
+  void update() override;
+  static void registerKeywords(Keywords& keys);
+};
+
+PLUMED_REGISTER_ACTION(TTMetaD, "TTMETAD")
+
+void TTMetaD::registerKeywords(Keywords& keys) {
+  Bias::registerKeywords(keys);
+  keys.use("ARG");
```
Iximiel:
```suggestion
```
As of v2.10 (I think) `ARG` is compulsory from bias and do no more reserved from ActionWithArguments, You do not need to call `use` on it

strandn:
@Iximiel Thank you for your reply. The local compilation is fixed by switching to an older version of gcc (I switched from v15 to v10).
I've confirmed your suggested change; hopefully the tests go well this time. I have checked that autoconf v 2.69 is available locally.

Iximiel:
@strandn look at the error message in the test results in debug mode:
```
make -C ttsketch testclean
make -C rt-ttmetad testclean
make test
Wed May  6 06:15:46 UTC 2026
Running regtest in /home/runner/work/plumed2/plumed2/regtest/ttsketch/rt-ttmetad
++ Test type: driver
++ Arguments: --plumed plumed.dat --trajectory-stride 500 --timestep 0.002 --igro traj.gro
++ Processors: 0
WARNING using plumed from /home/runner/opt/lib/plumed
/home/runner/work/plumed2/plumed2/regtest/ttsketch/rt-ttmetad/tmp
Checking for ttsketch module
../../scripts/run: line 189: 168215 Aborted                 (core dumped) $mpi $valgrind $plumed driver $arg > out 2> err
FAILURE: exit code 134
From line 519, file mps/mps.cc

Cannot create QN conserving MPS with bond dimension greater than 1 from a SiteSet
FAILURE
Diff for COLVAR:
```

PS:
We are failing in the tests: this means that the build part works :D 
The docker are failing because we need to setup configure, and then on the technical side we are good to go!

---

> @Iximiel Thank you for your reply. The local compilation is fixed by switching to an older version of gcc (I switched from v15 to v10).


@GiovanniBussi we might need to check also some newer gcc

strandn:
> make -C ttsketch testclean
> make -C rt-ttmetad testclean
> make test
> Wed May  6 06:15:46 UTC 2026
> Running regtest in /home/runner/work/plumed2/plumed2/regtest/ttsketch/rt-ttmetad
> ++ Test type: driver
> ++ Arguments: --plumed plumed.dat --trajectory-stride 500 --timestep 0.002 --igro traj.gro
> ++ Processors: 0
> WARNING using plumed from /home/runner/opt/lib/plumed
> /home/runner/work/plumed2/plumed2/regtest/ttsketch/rt-ttmetad/tmp
> Checking for ttsketch module
> ../../scripts/run: line 189: 168215 Aborted                 (core dumped) $mpi $valgrind $plumed driver $arg > out 2> err
> FAILURE: exit code 134
> From line 519, file mps/mps.cc
> 
> Cannot create QN conserving MPS with bond dimension greater than 1 from a SiteSet
> FAILURE
> Diff for COLVAR:
> ```
> 
> PS: We are failing in the tests: this means that the build part works :D The docker are failing because we need to setup configure, and then on the technical side we are good to go!

This error seems to be specific to the -debug-mpi- CI. After some digging, my understanding is that when plumed is built with --enable-debug-glibcxx, a flag is activated that changes the internal layout of standard C++ containers throughout all compiled code. ITensor is a pre-compiled library that was not built with this flag, so its internal data structures use a different layout. When plumed's ttsketch code interacts with ITensor at runtime, the mismatch likely caused undefined behavior (the above error really shouldn't happen/doesn't make sense since there are no "QN" data types). There seems to be a spurious error deep inside ITensor. Additionally, ITensor itself cannot be recompiled with this flag due to an incompatibility in its own source code (I tried several times to no avail). My conclusion is that ttsketch probably should not be included in debug builds that use this flag, and should be excluded from -debug-mpi-. Do you have any other thoughts?

src/ttsketch/TTHelper.cpp
```diff
@@ -0,0 +1,183 @@
...
+  // Z = int tt(x) dx (partition function); normalize to get probability density rho
+  auto Z = tt(1) * basis_int0[0];
+  for(int i = 2; i <= d; ++i) {
+    Z *= tt(i) * basis_int0[i - 1];
+  }
+  auto rho = tt;
+  rho /= elt(Z);
+  // ei[k]  = E[x_k],  eii[k] = E[x_k^2],  eij[k][l] = E[x_k * x_l] for k < l
+  std::vector<double> ei(d), eii(d);
+  std::vector<std::vector<double>> eij(d, std::vector<double>(d));
```
Iximiel: You can use PLMD::Matrix here, to compact the memory like you did few lines below, unless you need this `std::vector<std::vector<>>` for a specific function call

Iximiel:
That error is common in all the -debug- iterations, from https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode_using.html :

>To use the libstdc++ debug mode, compile your application with the compiler flag -D_GLIBCXX_DEBUG. Note that this flag changes the sizes and behavior of standard class templates such as std::vector, and therefore you can only link code compiled with debug mode and code compiled without debug mode if no instantiation of a container is passed between the two translation units.

Do you pass any stl container to ITensor? or just plain data? Maybe passing the underline array (`.data()`) and the size will solve the problem?

**The fastest solution might be building ITensor with the debug flags**, like we do with boost, since you are compiling it in the CI, you are controlling the flags, and eventually you may need to do it also with hdf5.


<details>
<summary>Itensor is doing this:</summary>

If this can help IT is failing at the first `if` in the following function, I do not know your code enough to go deeper, QN means that it is sparse, right?
```c++
void
new_tensors(std::vector<ITensor>& A,
            SiteSet const& sites,
            int m)
    {
    auto N = length(sites);
    auto a = std::vector<Index>(N+1);
    if(hasQNs(sites))
        {
        if(m==1) for(auto i : range1(N)) a[i] = Index(QN(),m,tinyformat::format("Link,l=%d",i));
        else Error("Cannot create QN conserving MPS with bond dimension greater than 1 from a SiteSet");
        }
    else
        {
        for(auto i : range1(N)) a[i] = Index(m,tinyformat::format("Link,l=%d",i));
        }
    A[1] = ITensor(sites(1),a[1]);
    for(int i = 2; i < N; i++)
        { 
        A[i] = ITensor(dag(a[i-1]),sites(i),a[i]); 
        }
    A[N] = ITensor(dag(a[N-1]),sites(N));
    }
```
</details>

strandn:
> Do you pass any stl container to ITensor? or just plain data? Maybe passing the underline array (`.data()`) and the size will solve the problem?

> The fastest solution might be building ITensor with the debug flags, like we do with boost, since you are ? > compiling it in the CI, you are controlling the flags, and eventually you may need to do it also with hdf5.

Passing .data() and size: The issue isn't that we're passing std::vector directly to ITensor. The problem is that ITensor's own types (like SiteSet) contain STL containers internally, and their constructors are inline. So they get compiled into our translation unit with _GLIBCXX_DEBUG active, giving those containers a different internal layout than what libitensor.a expects.

I looked deeper into this. Calling PrintData(sites) just before constructing the MPS (executed via plumed-compiled inline code) shows the SiteSet is correct:

site 1 = (dim=5|id=755|"n=1,Site")
site 2 = (dim=5|id=408|"n=2,Site")

But inspecting the same object in GDB at the point of failure (inside libitensor.a):

(gdb) print length(sites)
$2 = 0

The same SiteSet reports 0 sites to libitensor.a, causing hasQNs to weirdly return true and triggering the misleading error. The SiteStore pointer is valid and non-null; it's the internal vector that's being misread due to the layout difference.

Building ITensor with debug flags: I tried adding -D_GLIBCXX_DEBUG to ITensor's build flags. It fails to compile because ITensor's TenIter iterator is missing a const qualifier on operator*(), which LIBCXX_DEBUG's iterator validation requires:

  itensor/tensor/teniter.h:63:5: note: candidate: ... (near match)
  itensor/tensor/teniter.h:63:5: note:   passing
      'const itensor::TenIter<double*, itensor::VecRangeT<0> >*' as 'this'
      argument discards qualifiers

This *might* be a fixable bug in ITensor (adding const to operator*()), but patching a third-party library in CI feels non-ideal. I would say excluding ttsketch from -debug-mpi- seems like the cleanest option. Happy to hear if you have other thoughts.

Iximiel:
> > Do you pass any stl container to ITensor? or just plain data? Maybe passing the underline array (`.data()`) and the size will solve the problem?
> 
> > The fastest solution might be building ITensor with the debug flags, like we do with boost, since you are ? > compiling it in the CI, you are controlling the flags, and eventually you may need to do it also with hdf5.
> 
> Passing .data() and size: The issue isn't that we're passing std::vector directly to ITensor. The problem is that ITensor's own types (like SiteSet) contain STL containers internally, and their constructors are inline. So they get compiled into our translation unit with _GLIBCXX_DEBUG active, giving those containers a different internal layout than what libitensor.a expects.

:cry:, thank for the dumpster dive!

> This _might_ be a fixable bug in ITensor (adding const to operator*()), but patching a third-party library in CI feels non-ideal. I would say excluding ttsketch from -debug-mpi- seems like the cleanest option. Happy to hear if you have other thoughts.

This looks like not a problem that we can fix/workaround here: let's start with opening an issue on ITensor :smiling_imp:. Jokes apart, @GiovanniBussi I think their idea of excluding TTsketch from the -debug- runs might be the fastest solution, but has to be well documented with a comment in the CI aimed at forgetful future us

Iximiel:
@strandn it's time to `configure.ac`:
I would suggest a shameless copy of what has been done for metatomic:

add an `--enable-libitensor` and an `--enable-libhdf5` (that is activated by if libitensor is true, like libtorch and libmetatomic)
both the two `--enable-lib` will need to define the `__PLUMED_HAS_...`, and the ttskech module will need to have its code protected by the relative `#ifdef __PLUMED_HAS_...`

Iximiel:
@strandn for passing cppcheck throw also an ifdef for `__PLUMED_HAS_LIBHDF` somewhere in your module (you can use an `#if defined(xxx) && defined(yyy)` for a oneliner, like in metatomic.cpp:118)



Add a `PLUMED_CHECK_CXXFLAG([-fconcepts])` as the very first step when you test for libitensor, so the check will work without asking for -std=c++20 in configure.ac just for itensor.

Then write a LOUD comment around `PLUMED_CHECK_CXXFLAG([-fconcepts])` for _future us_, specifying that when we pass to C++20 we might not need that check anymore

configure.ac
```diff
@@ -1031,6 +1033,44 @@ if test $libmetatomic = true ; then
...
+      auto i = itensor::Index(2,"i");
+      auto j = itensor::Index(2,"j");
+      auto T = itensor::ITensor(i,j);
+      return 0;
+    }
+  ], [__PLUMED_HAS_LIBITENSOR], [ itensor openblas ], [true])
+  CXXFLAGS="$save_CXXFLAGS"
+  if test "$__PLUMED_HAS_LIBITENSOR" = yes ; then
+    PLUMED_CHECK_CXXFLAG([-fconcepts])
+  fi
```
Iximiel: You should not need to force fopenmp and fconcepts, because openmp is tested much more robustly at line [765](https://github.com/plumed/plumed2/blob/9b13f6fb405331e5692f13cd539dc941db7c2cb1/configure.ac#L765) (on master) and you are checking for fconcept as the very first thing of this clause:


```suggestion
  PLUMED_CHECK_CXXFLAG([-fconcepts])

  PLUMED_CHECK_CXX_PACKAGE([libitensor],[
    #include <itensor/all.h>
    int main() {
      auto i = itensor::Index(2,"i");
      auto j = itensor::Index(2,"j");
      auto T = itensor::ITensor(i,j);
      return 0;
    }
  ], [__PLUMED_HAS_LIBITENSOR], [ itensor openblas ], [true])
```

strandn:
@Iximiel It feels like we are close! I am a bit puzzled by the errors occurring in -debug-mpi- and -nvhpc-mpi-.

In the former, rt-grid and rt-protein regtests are triggering segfaults:

> Running regtest in /home/runner/work/plumed2/plumed2/regtest/python/rt-grid
> ++ Test type: python
> ++ Arguments: ./python-script.py
> ++ Processors: 0
> WARNING using plumed from /home/runner/opt/lib/plumed
> /home/runner/work/plumed2/plumed2/regtest/python/rt-grid/tmp
> Checking for python dependency
> ../../scripts/run: line 189: 160700 Segmentation fault      (core dumped) PYTHONPATH="$root/python:$PYTHONPATH" $($plumed_nompi config python_bin) $arg > out 2> err
> FAILURE: exit code 139

Clearly, these tests are not explicitly related to my own contributions.

In the latter, there appears to be a compatibility issue between nvc++ and the -fconcepts flag.

Do you have any suggestions on how to get around these issues and/or if there is a preferred pattern to follow to address them? Thank you in advance!

Iximiel:
> In the former, rt-grid and rt-protein regtests are triggering segfaults:

That's kind of strange, since the rt-protein test passes, and it is very similar to the other 2, and in the standard -debug- everything passes

For the nvhpc I think that the easiest solution is to exclude ttskectch also from that CI branch, because such a compiler incompatibility it is not easy to solve

It also looks like you forgot to push the configured `configure`

strandn:
@Iximiel I just noticed something odd. I was going through all the tests that pass and the ttsketch test seems to get skipped in all of them. For example, in docker / docker (fedora39):

> #11 1410.4 + test ttsketch/rt-ttmetad/ NOT APPLICABLE
> #11 1410.4 + check file ttsketch/rt-ttmetad/report.txt for more information

Iximiel:
> @Iximiel I just noticed something odd. I was going through all the tests that pass and the ttsketch test seems to get skipped in all of them. For example, in docker / docker (fedora39):
> 
> > #11 1410.4 + test ttsketch/rt-ttmetad/ NOT APPLICABLE
> > #11 1410.4 + check file ttsketch/rt-ttmetad/report.txt for more information

That's because you have correctly setup the tests with `plumed_needs=libitensor` :)

That it working as intended: means that even with a --enable-modules=all, if the necessary libraries are not present that test does not run, since ttskecth can't be executed. If you look more, you can see that ttsketch tests are not the only tests to be skipped there.
The docker tests are there to test "less standard" configurations, and are not setup to mimic 100% the standard linuxWF run.

You can add a stub class with an immediate error like this https://github.com/plumed/plumed2/blob/daf88e11023ce6e2c964a5d41ac13a2a6b7278ea/src/metatomic/metatomic.cpp#L129 (I think it is better to use the `error` method than a direct throw) to inform the user about the installation/
And also using the trick of putting the registration for the action and for the keywords outside the ifdef guard https://github.com/plumed/plumed2/blob/daf88e11023ce6e2c964a5d41ac13a2a6b7278ea/src/metatomic/metatomic.cpp#L990, so that one can build a complete manual without installing the libraries necessary for running


---

coverage is failing because libitensor can't be enabled:

>checking whether ccache mpic++ accepts -fconcepts... yes
>checking libitensor without extra libs... no
>checking libitensor with  -litensor -lopenblas... no
>configure: WARNING: cannot enable __PLUMED_HAS_LIBITENSOR

please delete temporarily  this line https://github.com/plumed/plumed2/blob/daf88e11023ce6e2c964a5d41ac13a2a6b7278ea/.github/workflows/linuxWF.yml#L126 and see what is happening at configuration time

strandn:
@Iximiel We've made more progress. The only remaining failure is rt-symfunc in linux / linux (debug-mpi). This test is unrelated to ttsketch (it uses the adjmat and symfunc modules via the Python PLUMED bindings). Note that the test passes locally when run with the same --enable-debug --enable-debug-glibcxx flags using locally-built Python bindings. It does not crash, and make test passes cleanly. In the CI, rt-protein and rt-grid (which failed last week) now pass.

Iximiel:
I have no clues... It may be a compiler version issue?

Can you try to not install ITensor with the same logic that does not activate ttsketch in debug mode?

carlocamilloni:
@strandn good job! @GiovanniBussi i think this can be merged

PR #1401, opened 2026-04-22
Exploiting VectorTyped to avoid copies during the NL construction

Iximiel:
##### Description

First of all I added a set of test that can be useful to understand how the NL is breaking in all the possible situations (serial, serial+omp, mpi, mpi+omp)

Then I exploited the trick PLUMED uses to transfer `Vector`s wit MPI to avoid creating an extra unsigned array to be gathered and then copied into the local neigborlist.
In a serial or non mpi run this skips a significant portion of code since it works directly on the `neighbors_` array. In a MPI run the code still uses a temporary array, but communicate directly into the main `neighbors_` array, skipping the last copy.

I think this will lower the RAM tax that the NL imposes on the PC for larger systems

I have a fast benchmark on drag races of 60 seconds (how many steps in 60 seconds) with 4 omp threads:

| Atomi | LC     | LC-v   | Δ% LC→LC-v | NL     | NL-v  | Δ% NL→NL-v |
|-------|--------|--------|------------|--------|-------|------------|
| 125   | 506813 | 557075 | +9.92%     | 445501 | 456190| +2.40%     |
| 1000  | 42082  | 44953  | +6.82%     | 38427  | 40167 | +4.53%     |
| 8000  | 6705   | 6732   | +0.40%     | 3201   | 3342  | +4.40%     |
| 27000 | 1559   | 1586   | +1.73%     | 501    | 517   | +3.19%     |
| 42875 | 946    | 960    | +1.48%     | 201    | 221   | +9.95%     |

And with 2 MPI processes with 3 opemMPtreads:

| Atomi | LC     | LC-v   | Δ% LC→LC-v | NL     | NL-v  | Δ% NL→NL-v |
|-------|--------|--------|------------|--------|-------|------------|
| 125   | 329986 | 335234 | +1.59%     | 278411 | 278566 | +0.06%     |
| 1000  | 29640  | 29800  | +0.54%     | 25404  | 25970  | +2.23%     |
| 8000  | 3909   | 3879   | −0.77%     | 1941   | 2001   | +3.09%     |
| 27000 | 901    | 894    | −0.78%     | 281    | 301    | +7.12%     |
| 42875 | 531    | 501    | −5.65%     | 121    | 121    |  0.00%     |

`NL` runs this:
```
[...]

PR #1402, opened 2026-04-23
Lowering memory usage in the NL

Iximiel:
##### Description

This is a work in progress, as I am not sure that this is speeding things up but at least it should not slow down the run.

The idea is that each atom in a NL will have a limited number of neighbors, so I should reserve the memory in a linear fashion instead of in a squared fashion. See the image below (LC="Linked Cells" NL="Neighbor List", "**-v"= this branch):

memory_usage

From the point of view of saving memory this is working (for 42875 atoms we work with 230 MB instead of 14GB!), see the graph above. I am still doing some measurement to understand if this is slowing down or not the run.
If the number of possible neighbors is underestimated the push back will trigger a reallocation, but that should happen a few times and only in the first few NL recalculations I think, since when the array is reallocated or elongated it _should_ grow at least of its size, from what I understood. I do not know If adding an extra command in the input line for specifying the estimation of the number of neighbors per atom could be useful

I did not extended the "optimization" to the use of the NL without the update of the list with the stride since recalculating the Neigbours slows things significantly down in the regime in which storing all the couples is faster than updating the NL with a stride (with less than 1000 atoms).

What do you think?

(the test are a cherry pick of the ones in #1401)

##### Target release

I would like my code to appear in release __v2.11__?

##### Type of contribution

##### Copyright

##### Tests


Iximiel:
@GiovanniBussi 

this triggers an error in maze and PIV, that is my definitely fault. because they access the NL while is not initialized yet

---

A curiosity: since I do not know well the idea behind the maze module, I would like to ask a question about how the COM in the `maze::Optimizer` is calculated:
```c++
Vector Optimizer::center_of_mass() const {
  const unsigned nl_size = neighbor_list_->size();

  Vector center_of_mass;
  center_of_mass.zero();
  double mass = 0;

  for (unsigned int i = 0; i < nl_size; ++i) {
    unsigned int i0 = neighbor_list_->getClosePair(i).first;
    center_of_mass += getPosition(i0) * getMass(i0);
    mass += getMass(i0);
  }

  return center_of_mass / mass;
}
```
It is the COM of the first elements of the couples of the NL, is that the wanted behavior? In single list mode one of the atoms will inevitably be excluded by the calculation (since in the couple lower triangular matrix the diagonal is excluded).

(on another note the link in the manual for more info about the module is broken,you can find that link here:https://www.plumed.org/doc-master/user-doc/html/module_maze/)

GiovanniBussi:
Thanks @Iximiel. Perhaps @jakryd can comment on the maze module?

Iximiel:
@GiovanniBussi 
In the meantime I devised a workaround that it is not much intrusive.
Hopefully this won't break anything similar in the nest

Iximiel:
@GiovanniBussi I removed the workaround, and I added the no-check method (in style at/operator[] in vector) leaving the old call as a "safer" call, so only additions to the API. I re-did the measurements a few times, and apparently there is no difference between master and this (thanks to the branch predictor, I think).

During the benchmarks I noted that the performance of this branch is more consistent than master regardless what happens on the PC.

I need to add a few tests for the new methods, and I think it is done.

Looking at git grep, in the core modules, getClosePair is used only in coordinationBase and in pycv, I would change it into the "less safe" option.


---

>I did not extended the "optimization" to the use of the NL without the update of the list with the stride since recalculating the Neigbours slows things significantly down in the regime in which storing all the couples is faster than updating the NL with a stride (with less than 1000 atoms).

Quick translation: without the NL algorithm activated (aka with stride==0) the memory optimization is not enabled, since it would be using `getIndexPair` at each invocation of `get*Pair`, and this kills performances in any regime.

GiovanniBussi:
OK. Related to this:

> Looking at git grep, in the core modules, getClosePair is used only in coordinationBase and in pycv, I would change it into the "less safe" option.

~Please post the timings after the change, so that we can see if the speedup is confirmed. Thanks!~

Sorry, I see this is only for memory footprint. So as long as tests pass I think we can merge it. Thanks!

jakryd:
Thanks for letting me know @Iximiel. What should I change in the NL part to make it more compatible with your updates? Sorry, I haven't had the time to catch up on all the recent optimizations. If leaving this as it's currently implemented does not cause any problems, I would not touch the maze module temporarily, because I've been working on a large update.

Iximiel:
> OK. Related to this:
> 
> > Looking at git grep, in the core modules, getClosePair is used only in coordinationBase and in pycv, I would change it into the "less safe" option.
> 
> ~Please post the timings after the change, so that we can see if the speedup is confirmed. Thanks!~
> 
> Sorry, I see this is only for memory footprint. So as long as tests pass I think we can merge it. Thanks!

@GiovanniBussi as now I have the pc doing some extra benchmarks, the timings are nearly identical for the safe and unsafe version, but I need to update the tests with a fast check on `ready()/getUpdatedPair()` before getting this ready.
We need to decide what to do with the memory warning.
(I'll post a table as soon the PC stops the benchmark for the other branch)

> Thanks for letting me know @Iximiel. What should I change in the NL part to make it more compatible with your updates? Sorry, I haven't had the time to catch up on all the recent optimizations. If leaving this as it's currently implemented does not cause any problems, I would not touch the maze module temporarily, because I've been working on a large update.

Hi @jakryd, there are no needed changes in your code: I rolled back the modification to keep the backward compatibility of the API.
My doubt on how maze calculates the COM is there because I do not know the idea behind maze: do you want to calculate the COM only to the first element of each couple passed to the optimizer?.

If you want to change something because you are paranoid about the performances, you can change the loop to run safely the unchecked version. I do not think it is necessary, looking at the measurement I did in coordination using `getUpdatedPair` and `getClosePair` show that there is not significant impact on the performances, especially in a loop as small as the one you use in the COM:
```c++
if (neighbor_list_->ready()){
  for (unsigned int i = 0; i < nl_size; ++i) {
    //without internal check (safe only if ready() is true)
    unsigned int i0 = neighbor_list_->getUpdatedPair(i).first;
...
  }
} else {
  for (unsigned int i = 0; i < nl_size; ++i) {
    //with internal check
    unsigned int i0 = neighbor_list_->getClosePair(i).first;
...
  }
}
```

Iximiel:
> @GiovanniBussi as now I have the pc doing some extra benchmarks, the timings are nearly identical for the safe and unsafe version

I confirm what I said, I did some runs, fewre number of atoms for doing more than one run of 120 seconds:
<details>
<summary> With 2mpix3omp </summary>

meanplotmpi
</details>

<details>
<summary> With 4omp </summary>
meanplot
</details>

GiovanniBussi:
OK so this proposed change:
- decreases the memory footprint
- does not affect performance.

Correct?

If so I can merge. Thanks!

Iximiel:
> OK so this proposed change:
> 
>     * decreases the memory footprint
> 
>     * does not affect performance.
> 
> 
> Correct?

yes!
but let me rebase on master for avoiding the current problems of github

PR #1403, opened 2026-04-24
Using Linkcells to accelerate the NL

Iximiel:
##### Description

Hello, I think this is the last (or at least the latest) PR in the NL series.

Fist of all I need to do some extra modification to the code (like compacting and move the new template function in the LC header, and rebasing this on the modifications of #1401, if it gets accepted), but I wanted to show the current state of this to get some feedback.

I tried to combine linkCells with the standard NL algorithm, to see if things will be speed up.

Along the way I rationalized the LinkCells loops into a template function and I tried to parallelize the LC version of the NL.

Here a few graphs, they follow the #1401 nomenclature and I used the same commands `**-v` is this PR, and `**-noomp` is this PR with the openmp code for the linkcell commented out.

These are the results on the NL algorithm:

nllc_plot_NL

The LC-NL "linearizes" the time and, contrary to LC -see below-, this seems to benefit from the omp parallelization.
As we were predicting, you can see the old implementation start faster but scales worse than the new implementation. So I was thinking of keeping the two implementations and decide when use one or another, with the default behavior derived from some measurements that can be overridden by the user input.

---

These are the timing on LC the `-v` has openmp, the `-noomp` has the new interface but no openmp

nllc_plot_LC

I do not think that adding openmp is a good idea to LC, but that was not the point o these modifications. Maybe because it is working directly in the result array.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


GiovanniBussi:
Let's leave this dangling.

This is providing speedup in some corner cases, but in order to use it properly we should either (a) implement proper heuristic or (b) allow the user to control it with an env var or better from the input file.

Iximiel:
Here I made some measurements, as usual -v is this branch, with 125 atoms this  is slower that the plan NL with the simple cubic, with more atoms it starts to scale more linearly, thanks to the LinkCells:
NLwithLC

and the histogram to confront the results:
NLwithLCHisto

In the last commit I added a environmental variable to control this and the architecture in the NL to add this option, I added a toggle method to not add any extra element to the constructor, since changing the option has no impact on the setup, and it will be easier to implement when the NL is used

Iximiel:
@GiovanniBussi the nvhpc test and the pycv failed on checking out the repo, I think that rerunnning them will make them pass

I do not understand what is happening to the mac tests

GiovanniBussi:
@Iximiel only 5 tests are crashing. Are the tests using any of the features you implemented here? There might be numerical issues. Differences seem significant though (I quickly skimmed them)

Iximiel:
I do not understand why they pass on linux and intel and not on mac
they are all about the neighbor list

PR #1406, opened 2026-04-28
Multi gyration cv

gtribello:
##### Description

I made some changes to the way the gyration CV functions after the discussion with Roman on Monday.  The idea was to all users to use the multi colvar framework for GYRATION.  In other words, I changed the GYRATION command so you can now do:

````
gy: GYRATION ATOMS1=1-6 ATOMS2=7-12 ATOMS3=13-18 TYPE=RADIUS
PRINT ARG=gy
```` 

You can thus compute gyration radii for three sets of atoms with a single PLUMED command.  Furthermore, if you do the calculation in this way, you avoid virtual function calls, and the multiple calculations of the gyration radii are parallelised.

While I was making these changes, I noticed that we didn't have any regression tests that tested the MASS_WEIGHTED version of GYRATION.  I thus added a test of this functionality.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### CODE REVIEW

**Comment by gtribello** [comment]
The problem here is with rt-make-fortran.  It seems very weird.  @Iximiel can you take a look at this?  Please take a look at the work I have done in colvar/Gyration.h.  I am not sure if I have done everything right with making the templates work correctly.  Thanks

File: src/colvar/Gyration.h [hunk 1]
```diff
@@ -0,0 +1,471 @@
...
+// masses as weights or with unit weights
+#define NUMBER_OF_GYRATION_TYPES 11
+
+namespace PLMD {
+namespace colvar {
+
+template <typename T=double>
+class Gyration : public Colvar {
+private:
+  enum CV_TYPE:unsigned {RADIUS=0, TRACE=1, GTPC_1=2, GTPC_2=3, GTPC_3=4, ASPHERICITY=5, ACYLINDRICITY=6, KAPPA2=7, GYRATION_3=8, GYRATION_2=9, GYRATION_1=10,
```
Iximiel, line 39: This is a very small point:
since I saw you are modernizing by putting the "CV_TYPE::" when you call it, you can go all in and declare this a `enum class`
```suggestion
  enum class CV_TYPE:unsigned {RADIUS=0, TRACE=1, GTPC_1=2, GTPC_2=3, GTPC_3=4, ASPHERICITY=5, ACYLINDRICITY=6, KAPPA2=7, GYRATION_3=8, GYRATION_2=9, GYRATION_1=10,
```

This adds the advantage to raise a warning if a swithc/case statement is not complete

src/colvar/Gyration.h
```diff
@@ -0,0 +1,471 @@
...
+// The enum then contains double the number of types defined here as you can do every one of them with
+// masses as weights or with unit weights
+#define NUMBER_OF_GYRATION_TYPES 11
+
+namespace PLMD {
+namespace colvar {
+
+template <typename T=double>
+class Gyration : public Colvar {
+private:
```
Iximiel: also this is not important, but works exactly as before (it maybe faster because it is not using the preprpcessor, but it may raise some warnings because is typed)
```suggestion
namespace PLMD {
namespace colvar {

template <typename T=double>
class Gyration : public Colvar {
private:
// The number of types of gyration are defined here
// The enum then contains double the number of types defined here as you can do every one of them with
// masses as weights or with unit weights
static constexpr unsigned NUMBER_OF_GYRATION_TYPES = 11;
```

src/colvar/Gyration.h
Iximiel: Now it works, so take this new comment for a next iteration: you can use the tools/BitmaskEnum.h:
```suggestion
  enum CV_TYPE:unsigned {RADIUS=1, TRACE=1<<1, GTPC_1=1<<2, GTPC_2=1<<3, GTPC_3=1<<4, ASPHERICITY=1<<5, ACYLINDRICITY=1<<6, KAPPA2=1<<7, GYRATION_3=1<<8, GYRATION_2=1<<9, GYRATION_1=1<<10, WITH_MASS=1<<15};
```
**If** the tools that are in that header make all the various if statements easier to work with
 (and you do not need the `NUMBER_OF_GYRATION_TYPES` macro)

src/colvar/Gyration.h
```diff
@@ -0,0 +1,471 @@
...
+                         RADIUS_MASS=NUMBER_OF_GYRATION_TYPES+RADIUS,TRACE_MASS=NUMBER_OF_GYRATION_TYPES+TRACE,
+                         GTPC_1_MASS=NUMBER_OF_GYRATION_TYPES+GTPC_1, GTPC_2_MASS=NUMBER_OF_GYRATION_TYPES+GTPC_2, GTPC_3_MASS=NUMBER_OF_GYRATION_TYPES+GTPC_3,
+                         ASPHERICITY_MASS=NUMBER_OF_GYRATION_TYPES+ASPHERICITY, ACYLINDRICITY_MASS=NUMBER_OF_GYRATION_TYPES+ACYLINDRICITY, KAPPA2_MASS=NUMBER_OF_GYRATION_TYPES+KAPPA2,
+                         GYRATION_3_MASS=NUMBER_OF_GYRATION_TYPES+GYRATION_3, GYRATION_2_MASS=NUMBER_OF_GYRATION_TYPES+GYRATION_2, GYRATION_1_MASS=NUMBER_OF_GYRATION_TYPES+GYRATION_1
+                        } rg_type{RADIUS};
+  bool nopbc;
+  std::vector<double> value;
+  std::vector<double> derivs;
+public:
+  using precision = T;
```
Iximiel: This is more on the point about youd doubts:
Like this it will always work in double mode, regardless of the T, unless that is intentional, you should rename most of the `double` into `precision` or to `T`
```suggestion
  std::vector<precision> value;
  std::vector<precision> derivs;
public:
  //move this on top of the class so that the two vectors above will know what precision is
  using precision = T;
```

Iximiel (approved):
> The problem here is with rt-make-fortran. It seems very weird.

I agree.

I think the test failing is due to intel changing things (maybe it is finally come the time in which intel has decided to not use ifx as the alias for flang anymore?), or maybe is a fluctuation, because it is the very first time we see it that `ifx` is not there...

> @Iximiel can you take a look at this? Please take a look at the work I have done in colvar/Gyration.h. I am not sure if I have done everything right with making the templates work correctly. Thanks

I think that unless you want to force the double precision whatever is the `T` you should rename some `double` into T or `precision` in the cv, keeping some explicit double where you do not want to compromise on the precision.

I did not resist pointing at some petty things in the enum. But those can be ignored if this works

edit: readability

gtribello:
Hi @Iximiel 

Thanks for the comments.  I had some questions based on what you wrote that I wanted some further clarification on.

1. For the error with rt-make-fortran: does your comment mean that we are having the same problem with other branches of plumed?  For example, if I were to run the tests on the master version of PLUMED would I see the same problem now?  In other words, is this a general problem or it something that I have introduced?

2.  I am not sure if I understood the comment about the BitmaskEnum for the enum.  I included the NUMBER_OF_GYRATION_TYPES variable because all the options in the enum are duplicated as you can do the gyration with unit weights or the masses of the atoms as weights.  I hoped that if I made this clear in the way I implemented things, it would be easy to extend if anyone wants to add another type at some point in the future.  I am not sure if this is still clear if I use the BitmaskEnum.

3. I think I have worked out what to do with the doubles in the calculateCV function.  What should I do if I have PLMD::Vectors, though?  How do I make these typed and not simply arrays of doubles?

4. If I change the enum to an enum class, PLUMED no longer compiles.  Also, you don't use enum class in Distance.h. You just use enum.  You get the warning about case statements not being complete when you just use an enum when you compile on my machine.  I saw this because in the earlier version of the code, there was a TOT option in the enum that wasn't actually used anywhere in the code, and which I removed.  I thus would rather not make this change as I cannot see the advantage.

src/colvar/Gyration.h
```diff
@@ -0,0 +1,471 @@
...
+  case GTPC_1:
+  case GTPC_2:
+  case GTPC_3:
+  case GTPC_1_MASS:
+  case GTPC_2_MASS:
+  case GTPC_3_MASS: {
+    int pc_index = cvin.mode-2; //index of principal component
+    if( pc_index>=NUMBER_OF_GYRATION_TYPES ) {
+      pc_index = pc_index - NUMBER_OF_GYRATION_TYPES;
+    }
```
Iximiel, line 364: Another small point:

this is robust to the enum changes, because it does not rely on GTPC_1 being 2 (not astyled)
```suggestion
  case GTPC_1:
  case GTPC_2:
  case GTPC_3:
  case GTPC_1_MASS:
  case GTPC_2_MASS:
  case GTPC_3_MASS: {
  unsigned pc_index = 0;
  switch(cvin.mode)
  case GTPC_1:
  case GTPC_1_MASS:{pc_index = 0; break;}
  case GTPC_2:
  case GTPC_2_MASS:{pc_index = 1; break;}
  case GTPC_3:
  case GTPC_3_MASS:{pc_index = 2; break;}
  default: error(you should not be here)
  }
```

Iximiel:
Hi @gtribello, I should have been clearer with some suggestions:

>     1. For the error with rt-make-fortran: does your comment mean that we are having the same problem with other branches of plumed?  For example, if I were to run the tests on the master version of PLUMED would I see the same problem now?  In other words, is this a general problem or it something that I have introduced?

Seems to be common also in other branches: ifx is missing, and I think I can blame intel

>     2. I am not sure if I understood the comment about the BitmaskEnum for the enum.  I included the NUMBER_OF_GYRATION_TYPES variable because all the options in the enum are duplicated as you can do the gyration with unit weights or the masses of the atoms as weights.  I hoped that if I made this clear in the way I implemented things, it would be easy to extend if anyone wants to add another type at some point in the future.  I am not sure if this is still clear if I use the BitmaskEnum.

That is a suggestion for a next iteration, to store the "has mass" as a single bit in the enum, and avoid the have the extra NUMBER_OF_GYRATION_TYPES, and so you can count on having up to 31 gyration types instead of """"only"""" 15, and you do not need to update that 11.

>     3. I think I have worked out what to do with the doubles in the calculateCV function.  What should I do if I have PLMD::Vectors, though?  How do I make these typed and not simply arrays of doubles?

You can use VectorT<float> (that is an alias for VectorTyped<float,3>, for other sizes you can add the alias in Vector.h or write explicitly VectorTyped<type,num>).

>     4. If I change the enum to an enum class, PLUMED no longer compiles.  Also, you don't use enum class in Distance.h. You just use enum.  You get the warning about case statements not being complete when you just use an enum when you compile on my machine.  I saw this because in the earlier version of the code, there was a TOT option in the enum that wasn't actually used anywhere in the code, and which I removed.  I thus would rather not make this change as I cannot see the advantage.

That should have been a suggestion for a deeper modification, maybe it is better now to drop the typesafety to have the feature, but I still think that can help in having some extra guard rails in the code.
In my bad defense I did `Distance.h` some time ago. But I should have checked that multicolvartemplate uses an unsigned for the `mode` parameter in the ColvarInput parameter, adding the `class` makes the enum "typesafer" and makes the compiler be more attentive when you compare it with other integers, making you cast it explicitly. It is more robust, but has is toll in terms of code written.


As now it does not compile because  NUMBER_OF_GYRATION_TYPES  is unsigned and confronted with an signed integer

PR #1418, opened 2026-05-19
Fix 1417

### LINKED ISSUE #1417: Issue with rounding of timestep in PLUMED v2.10 +
@gtribello I managed to reproduce what we discussed today. At the end of this message, there's a simple test showing the problem. I have in my local clone of v2.9, but I will not push it until it is solved for v2.10 and upwards.

The input below should create a `COLVAR` file which looks like:
```
#! FIELDS time c
 20000.000000 1.000000
#! FIELDS time c
 20000.000000 1.000000
```

Instead, for v2.10 and up, it looks like
```
#! FIELDS time c
 20000.000000 1.000000
#! FIELDS time c
 20000.000950 1.000000
```

I think I know the reason. [This line](https://github.com/plumed/plumed2/blob/daf88e11023ce6e2c964a5d41ac13a2a6b7278ea/src/core/PlumedMain.cpp#L746) is not updating the value stored in the `ActionToPutData* ts`. In fact, I tried to fix it by doing something like `ts->setValuePointer("timestep", & rounded )`, but it cannot work, because the action can only store a float. So, in order to have no error, I have the redefine a local `float` variable and pass a pointer to it, but this will reintroduce the rounding issue.

So, this seems something unavoidable with the current design of having the timestep stored in an ActionToPutData. Did I get it right? Can you envision an easy fix? 

Test code below.

```
#include <vector>
#include "plumed/wrapper/Plumed.h"
#include <iostream>

using namespace PLMD;

template<typename T>
void run(){

  auto natoms=10;
  std::vector<T> masses;
  std::vector<std::array<T,3>> positions;
  std::vector<std::array<T,3>> forces;
  T box[3][3];
  T virial[3][3];
  
  Plumed p;

  p.cmd("setRealPrecision",int(sizeof(T)));
  p.cmd("setNatoms",natoms);
  p.cmd("setTimestep",(T)0.002);
  p.cmd("init");
  
  p.cmd("readInputLines",
    "c: CONSTANT VALUE=1.0 \n"
    "PRINT ARG=c FILE=COLVAR RESTART=YES\n"
  );

  // dummy settings, not really used
  positions.resize(natoms);
  masses.resize(natoms);
  forces.resize(natoms);

  for(unsigned i=0;i<natoms;i++) for(unsigned j=0;j<3;j++) positions[i][j]=0.0;
  for(unsigned i=0;i<natoms;i++) for(unsigned j=0;j<3;j++) forces[i][j]=0.0;
  for(unsigned i=0;i<natoms;i++) masses[i]=1.0;

  p.cmd("setStep",10000000);
  p.cmd("setMasses",&masses[0]);
  p.cmd("setPositions",&positions[0][0]);
  p.cmd("setForces",&forces[0][0]);

  p.cmd("setBox",&box[0][0]);
  p.cmd("setVirial",&virial[0][0]);
  p.cmd("calc");

}

int main(){
  run<double>();
  run<float>();
}

```

GiovanniBussi:
##### Description

Attempt to fix #1417 

@gtribello can you please check if I did it as intended?


GiovanniBussi:
@gtribello there is an error when customizing units. Maybe the logic of `ActionToPutData::updateUnits` should be made available outside ot the `ActionToPutData`  class?

PR #1419, opened 2026-05-20
Fix 1417 again

gtribello:
##### Description

@GiovanniBussi I think this should fix isssue #1417.  I took your original fix kept the first two commits so kept your test.  I removed the third commit that had the changes to the code, as they were not fixing the problem.  I thus closed your PR and replaced it with this one.

The way the code was working (and why it was not doing what was expected) was as follows:

1. You get a pointer containing the timestep from PLUMED and set a PLMD::Value equal to the timestep.
2. When you set the PLMD::Value  you do all the schenanigans with the float to double conversion.
3. You then update the units, which resets the value of the PLMD::Value to the value from the MD code (multiplied by the conversion factor), so all the schenanigans that were done are lost.

I have changed it so in point 3, you multiply the quantity stored in the PLMD::Value by the conversion factor, so it should work now.  Note that this resetting of the units is performed only if you pass a quantity that is fixed throughout the simulation and set before the MD loop starts.  

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Spoke too soon.  Let me investigate.

gtribello:
OK, @GiovanniBussi, I hope this should be a better fix.  The only thing is that I now also do the same thing that is done with the timestep when you transfer the value of k_B T.  It is probably less critical in that case, but making that change is hopefully harmless.

Not sure why codecheck is failing.  It looks like it is not able to download something or other, and that it is not because of a change I have made.

GiovanniBussi:
Yes I agree this solution is better (and cleaner)

gtribello:
Do you understand what has gone wrong with code check?

GiovanniBussi:
> Do you understand what has gone wrong with code check?

There has been a glitch in the ftp download of autoconf, it seems to be working now. Maybe you can wait for this test to run

gtribello:
OK will do.  If this all passes then I will merge.  OK?

GiovanniBussi:
Yes thanks!

PR #1432, opened 2026-07-16
This fixes a bug that appears if you use HISTOGRAM with READ

gtribello:
##### Description

This PR fixes a small bug in the master version of PLUMED that I found when I was reviewing masterclass 21.2.

If you are using driver to analyze the data in colvar file only using the --noatoms option then the pbc information is not transferred to PLUMED and you get a segmentation fault. As you can see I have resolved this problem in KDE.h by checking if the PbcAction is a null pointer and not calling any function from it if it is.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
@gtribello This looks fine but we need a regtest that fails before and pass after the change to be sure that the bug is not inadvertently reintroduced in the future

gtribello:
HI @carlocamilloni 

I have added the retest.  Here is what happens if you try the test with the master version of PLUMED that doesn't have this fix 

segfault

PR #1434, opened 2026-07-28
Restore setNumOMPthreads in the GROMACS 2025 patch

### LINKED ISSUE #1407: PLMD::OpenMP::getNumThreads() returns only 1
In my Bias, I am trying to use OpenMP.
However, calling `OpenMP::getNumThreads()` I always receive **1**, although the underlying Gromacs run reports _**10** OMP threads per each one of the parallel simulations_.

Is this related to #1380 ? i.e. Does Gromacs batch need more items to include?

Thank you

Amr

aalhossary:
Fixes #1407.

Up to `patches/gromacs-2024.3.diff` the GROMACS patch told PLUMED how many OpenMP threads
GROMACS was using (in the `/* PLUMED */` block of `src/gromacs/mdrun/runner.cpp`):

```cpp
int nth = gmx_omp_nthreads_get(ModuleMultiThread::Default);
plumed_cmd(plumedmain, "setNumOMPthreads", &nth);
```

When the 2025.0 patch was rewritten around the native PLUMED support that GROMACS 2025 ships,
that call was not carried over, so `PLMD::OpenMP::getNumThreads()` returns 1 for every GROMACS
2025 run and anything parallelised over OpenMP silently runs serially. As discussed in #1407,
setting `PLUMED_NUM_THREADS` by hand is not a workable substitute: the right number is not known
in advance on heterogeneous hardware, and it changes once GROMACS' dynamic load balancing
redistributes the work.

This restores the call in the `PlumedForceProvider` constructor, behind the same API-version gate
the 2024.3 patch used.

### Placement and safety

- `gmx_omp_nthreads_init()` is called in `runner.cpp` well before `mdModules_->initForceProviders()`,
  so the thread count is initialised by the time the force provider is constructed.
- `src/gromacs/applied_forces/{colvars,densityfitting,qmmm}` already include `gromacs/mdlib/`
  headers, so the added include does not violate GROMACS' module layering.

### Verification

- `plumedforceprovider.cpp.preplumed` is byte-identical to the GROMACS file at **v2025.0, v2025.1,
  v2025.2, v2025.3 and v2025.4**, so the hunk applies across the whole 2025 series. Confirmed by
  applying it to real v2025.0 and v2025.4 sources with the same flags `patch.sh` uses
  (`patch -u -l -b -F 5 -N`).
- Built GROMACS 2025.4 with this patch applied (clang 20.1.1, `-DGMX_USE_PLUMED=ON`, runtime mode)
  and ran `gmx mdrun -ntmpi 1 -ntomp N -plumed` with `PLUMED_NUM_THREADS` unset. `PLUMED.OUT`
  reports, with and without the restored block (same build tree, same kernel, only this hunk
  differing):

  | `-ntomp` | without the block | with the block |
  |---------:|------------------:|---------------:|
  | 1        | 1                 | 1              |
  | 4        | **1**             | 4              |
  | 8        | **1**             | 8              |

### Two related gaps, not addressed here

The same 2024.3 block contained two further items that are also absent from the 2025 patch:

1. `plumed_cmd(plumedmain, "setGpuDeviceId", &deviceId)` (gated on API > 9);
2. the warning that `-update gpu` combined with `-plumed` produces wrong results unless PLUMED
   only acts on neighbour-list-search and/or file-writing steps.

Both would need new fields plumbed through `PlumedOptions`, the way `replex_` and `ms_` already are.
Happy to open a separate issue or PR if you confirm these were dropped unintentionally rather than
deliberately.

### DIFF
--- a/patches/gromacs-2025.0.diff/src/gromacs/applied_forces/plumed/plumedforceprovider.cpp
+++ b/patches/gromacs-2025.0.diff/src/gromacs/applied_forces/plumed/plumedforceprovider.cpp
@@ -44,6 +44,7 @@
 #include "gromacs/domdec/domdec.h"
 #include "gromacs/domdec/domdec_struct.h"
 #include "gromacs/math/units.h"
+#include "gromacs/mdlib/gmx_omp_nthreads.h"
 #include "gromacs/mdrunutility/handlerestart.h"
 #include "gromacs/mdrunutility/multisim.h"
 #include "gromacs/mdtypes/commrec.h"
@@ -113,6 +114,15 @@ try : plumed_(std::make_unique<PLMD::Plumed>()),replex_(options.replex_)
             int res = 1;
             plumed_->cmd("setRestart", &res);
         }
+    }
+
+    if (plumedAPIversion_ > 5)
+    {
+        /* tell PLUMED how many OpenMP threads GROMACS is using, so that it can
+           parallelise its own work; without this PLMD::OpenMP::getNumThreads()
+           always reports 1 unless PLUMED_NUM_THREADS is set by hand */
+        int numOmpThreads = gmx_omp_nthreads_get(ModuleMultiThread::Default);
+        plumed_->cmd("setNumOMPthreads", &numOmpThreads);
     }
 
     if (isMultiSim(options.ms_))


aalhossary:
**N.B. — the count is read once, and that is deliberate.**

The value is fetched in the `PlumedForceProvider` constructor rather than refreshed per step, because on the
GROMACS side it cannot change afterwards: `gmx_omp_nthreads_set()` has exactly one production call site,
inside `gmx_omp_nthreads_init()` itself (every other caller is a unit test), and that init runs once in
`runner.cpp`. Dynamic load balancing redistributes DD domain volumes and the PP/PME split — it does not
re-thread the ranks — so `ModuleMultiThread::Default` is constant for the whole run.

Nothing here is frozen on the PLUMED side either, should that assumption ever change: `setNumOMPthreads`
is a plain assignment into the `OpenMP` singleton, `getNumThreads()` is read at point of use (right before
the parallel region) rather than cached at action construction, and `cmd_setNumOMPthreads` carries no
init-state guard — so it can legitimately be re-issued from `calculateForces()` if GROMACS ever gains the
ability to re-thread mid-run. Only worth doing behind a "value changed" check, since a per-step
`cmd("...")` pays a string lookup.

Two consequences worth flagging for reviewers:

* `PLUMED_NUM_THREADS` still wins. `getNumThreads()` consults the environment on its *first* call and
  latches, and that first call happens after this constructor-time set — so the existing manual override
  keeps working and explicit user intent still beats auto-detection.
* This passes `ModuleMultiThread::Default`, which is what the 2024.3 patch used, but GROMACS may assign
  different counts to Nonbonded/PME/Update. Happy to switch to a different module's count if you think
  another one better reflects what PLUMED should get.

aalhossary:
**CI note** — the three red jobs look unrelated to this change. This PR only edits a file under
`patches/gromacs-2025.0.diff/`, which PLUMED never compiles: no object files are produced from
`patches/` in a build, and the only references to it in the build system are a bash-completion
dependency on `patches/*.sh` and an install-time copy in `src/lib/Makefile`. That tree is consumed
solely when a user runs `plumed patch` against a GROMACS source tree, at which point GROMACS — not
PLUMED — compiles it.

The three failures:

* `linux (-nvhpc-mpi-)` fails during *dependency installation*, before PLUMED is configured or built:
  ```
  Err:4 https://developer.download.nvidia.com/hpc-sdk/ubuntu/amd64  nvhpc-24-3 24.3
    404  Not Found
  E: Failed to fetch .../nvhpc-24-3_24.3_amd64.deb  404  Not Found
  ```
  NVIDIA appears to have removed or rotated the HPC SDK 24.3 package. This should hit every PR until
  the workflow pins a newer nvhpc.

* `linux (-debug-)`: segmentation fault (exit 139) in `regtest/python/rt-protein`.

* `linux (-intel-)`: abort (exit 134) in `regtest/adjmat/rt-matmult`, reporting
  `corrupted size vs. prev_size`.

The latter two are memory faults inside PLUMED's own kernel — the Python interface and the `adjmat`
module respectively — in code paths this PR does not touch.

One caveat in the interest of full disclosure: the most recent green run of these jobs on `master` is
from 2026-07-20, and no other PR has exercised them since, so I could not compare against a same-day
baseline. Happy to rebase or to have the jobs re-run if you would like that confirmed.

aalhossary:
Superseded by #1435 — closing this one.

Two things were wrong here, both my fault:

1. **Wrong base branch.** `CONTRIBUTING.md` asks for bug fixes on the oldest maintained branch where the
   bug appears, and `patches/gromacs-2025.0.diff` is present on `v2.10` as well as `master`. I opened
   this against `master`. Retargeting in place was not an option: `master` is 766 commits ahead of
   `v2.10`, so simply switching the base would have turned the diff into an unreviewable wall. #1435 is
   the same single commit cherry-picked onto `v2.10` (applies cleanly — the only difference in that file
   between the branches is the `PLUMED_LOG_FILE` block from #1420, which sits below the insertion point).
2. **I did not fill in the pull request template**, including the copyright transfer. #1435 has it
   complete.

The technical content is unchanged, plus one clarification a reviewer would reasonably have asked for:
setting the thread count once at construction is sufficient, because `gmx_omp_nthreads_set()` has a
single production call site inside `gmx_omp_nthreads_init()` — every other caller is a unit test — so the
value cannot change mid-run. `PLUMED_NUM_THREADS` still takes precedence where set.

Apologies for the churn, and thanks for your patience.

PR #1436, opened 2026-07-29
kenref: new module for kinetic ensemble refinement against NMR data

aalhossary:
##### Description

This adds a new module, `kenref`, implementing **kinetic ensemble refinement**: a bias that restrains an
ensemble of replicas so that NMR observables computed *across* the ensemble reproduce experimental
values. It provides one action, `KENREF`, which is simultaneously a `Bias` and an `ActionAtomistic` —
it acts on coordinates directly and takes no `ARG`.

Three energy models are selectable by name via the `MODEL` keyword:

| `MODEL`    | restrains                                            |
|:-----------|:-----------------------------------------------------|
| `SIGMA`    | cross-relaxation rates (interproton NOE build-up)    |
| `PLATEAUS` | NOE plateau values                                   |
| `RELAX`    | longitudinal / transverse relaxation                 |

`RELAX` needs at least two replicas since the observable is an ensemble average; `SIGMA` and `PLATEAUS`
also work with one.

The numerics live in an external library, **KEnRef** (<https://github.com/Smith-Group/KEnRef>, BSD
3-Clause), which we maintain. The method and this implementation are published in:

* A. Alhossary and C. A. Smith, *Restraining Interproton Angular and Distance Dynamics with KEnRef*,
  **J. Phys. Chem. B 130**, 3076-3087 (2026), doi:[10.1021/acs.jpcb.5c08554](https://doi.org/10.1021/acs.jpcb.5c08554)
  — introduces KEnRef; corresponds to KEnRef version 1.0.0.
* C. A. Smith, A. Mazur, A. K. Rout, S. Becker, D. Lee and B. L. de Groot, *Enhancing NMR derived
  ensembles with kinetics on multiple timescales*, **J. Biomol. NMR 74**, 27-43 (2019),
  doi:[10.1007/s10858-019-00288-8](https://doi.org/10.1007/s10858-019-00288-8)
  — the kinetic ensemble refinement method KEnRef implements.

Both DOIs are also recorded in `src/kenref/module.yml` and cited in `src/kenref/module.md`. The module is `default-off` and is compiled only with `--enable-kenref`
(equivalently `--enable-modules=+kenref`), which finds `libkenref_core` through `pkg-config` or builds
it by delegating to KEnRef's own CMake. A PLUMED built without the flag is completely unaffected — no
new dependency, no module.

### Two things I would particularly like your opinion on

**1. The action is deliberately split across the two repositories.** `src/kenref/KEnRefBias.cpp` here
holds the *stable* half — `registerKeywords`, the PLUMED↔Eigen glue, `calculate()`, and the action
registration. The one-time *constructor*, which changes whenever KEnRef gains an energy model or alters
its parameter schema, is hosted in the KEnRef repository and compiled in through a small forwarding
translation unit, `KEnRefBias_setup.cpp`.

The intent is that adding an energy model to KEnRef never requires a pull request against PLUMED, which
also keeps the maintenance burden on us rather than on you. The cost is one file in this module that is
a two-line `#include` of a source file from the dependency. I have documented why in the file header and
in `module.md`, but if you would rather the whole action lived here, say so and I will inline it.

**2. Regression tests are intentionally thin, for the same reason.** `regtest/kenref/` contains one
model-agnostic smoke test: the action parses, builds its indexing from the experimental data, runs, and
produces its declared components. The numerical correctness of each energy model is validated in
KEnRef's own suite against reference values from the original R implementation. Asserting energies here
would mean a PLUMED-side change every time a model is added. If you would prefer at least one
value-asserting test pinned in-tree, I am happy to add one.

### Notes on the build integration

- The library check uses **plain `pkg-config` shell calls, not `PKG_CHECK_MODULES`**. That macro lives
  in `pkg.m4`, and PLUMED does not use `aclocal` — there is no `aclocal.m4` in the tree, and plain
  `autoconf` regenerates `configure` byte-for-byte. Calling it would have left `PKG_CHECK_MODULES`, and
  every `AC_SUBST` nested inside it, *unexpanded* in the generated `configure`. (I hit exactly this, so
  the comment in `configure.ac` is there to stop the next person reintroducing it.) Plain `pkg-config`
  is also what the `BASH_COMPLETION_DIR` probe already does.
- `configure` was regenerated with **autoconf 2.69**, matching the committed script: I verified that
  2.69 reproduces the current `configure` byte-for-byte before adding anything, so the diff here is
  purely additive (277 lines added, 0 removed). Regenerating with 2.72 instead produces ~10k lines of
  unrelated churn.
- The module resolves its include flags **once, at configure time**, into `Makefile.conf`, rather than
  calling `pkg-config` at Makefile parse time. This is deliberate: `sudo make install` re-parses the
  module Makefile as root with `PKG_CONFIG_PATH` stripped, so a parse-time lookup returns empty and the
  build dies. It mirrors how the link flags already flow (`configure` → `LIBS` → `DYNAMIC_LIBS`).
- `KENREF` removes the `ARG` keyword that `Bias` registers, guarded by `keys.exists("ARG")`. It biases
  coordinates, not CVs, so it has no arguments; without this, an input needs a bare `ARG=` on master
  while 2.10 requires it to be absent, and no single input file can serve both.

### Testing

- `./configure` **without** `--enable-kenref`, and with `libkenref_core` absent from `PKG_CONFIG_PATH`:
  builds clean, and `plumed manual --action KENREF` correctly reports the action does not exist.
- `./configure --enable-kenref` against an installed prefix: the module builds, `KENREF` registers, and
  `regtest/kenref` passes.

##### Target release

I would like my code to appear in release __v2.11__ (master — this is a new feature, not a bug fix).

##### Type of contribution

##### Copyright

  The module is released under the LGPL, like the other contributed modules. The external KEnRef library
  it links against is BSD 3-Clause; that is permissive and adds no restriction on PLUMED. Both are
  stated in `src/kenref/COPYRIGHT`.

##### Tests


carlocamilloni:
@aalhossary great job, all seems to be wokring correctly. @GiovanniBussi @gtribello @maxbonomi we should only think about the separatly hosted part of the code. I think it is fine and could be a  approach in particular for ML related stuff. I do not see any particular issue. What do you think about it?

aalhossary:
@GiovanniBussi @gtribello @maxbonomi I'd appreciate your input on @carlocamilloni 's question

gtribello:
Hello

This looks to me like a sensible way of managing the interface between PLUMED and kENRef.  It looks like it is similar to the way the interface between PLUMED and metatomic is being managed.  I would be fine with seeing this incorporated into PLUMED.

carlocamilloni:
I agree, when you are ready remove the draft and we can merge it

aalhossary:
> I agree, when you are ready remove the draft and we can merge it

Actually I was waiting for #1443 to be solved, because Plumed has at least 2 problems when dealing with domain decomposition:
1. Improper use of MPI communicator(s), which I mentioned in #1443
2. improper handling of PBC if the initial protein structure was broken across the boundaries.

Both of them are found to be touched (directly or indirectly in https://gitlab.com/gromacs/gromacs/-/merge_requests/4622, However, it seems that the responsibility is shared between Plumed and Gromacs. 

So, you are right. Let's first merge KEnRef without domain decomposition. Then we can handle these issues later.

PR #1448, opened 2026-08-21
Bug fix for MultiColvar for codes that don't pass charges to PLLUMED

gtribello:
##### Description

This PR fixes a bug that was identified on the maillist.  The problem is if you use a code like CP2K (or simplemd) that does not pass charges to PLUMED then you cannot calculate any MultiColvars - even the ones you should be able to calculate.  Adding an if condition to check if the charges are passed resolves the issue.

I have also added a regtest to check that we don't readd the bug in future versions.

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
Are the charges set to zero inside multicolvar now if not set by the code?

gtribello:
Yes @carlocamilloni and I have made it so that if you try to compute a DIPOLE with a code that doesn't pass charges to PLUMED then PLUMED will crash.  I have done this both in this PR on v2.10 and PR #1449, which is the same thing on master.

PR #1450, opened 2026-08-24
Changes to avoid sparse matrix lookup operations unless they are absolutely necessary that make FunctionOfMatrix faster

gtribello:
##### Description

PLMD::Value handles two types of matrix:

* Dense matrices where you have access to every element.
* Sparse matrices where each row has values in a subset of the columns.

The sparse matrices occur naturally in the context of calculating a contact_matrix.  All the central atoms that are specified have some neighbours so a small subset of the columns in each row are non-zero.  However, there are likely  no central atoms that have no neighbours so no row is entirely filled with zeros.

If you have a sparse matrix you have to do some expensive look-up operations to get element i,j.  Obviously, you wan to avoid these operations if you have a dense matrix!  This change ensures that we do indeed avoid performing these operations when the matrix is dense.  We weren't avoiding those expensive lookup operations in older versions of the code, which was adding (unecessary) computational expense.

To make this work I had to make a small change in LinkCells.  In essence, before this change if you computed a contact matrix that was not sparse (because the cell was large relative to D_MAX) the order the atoms are in is  shuffled because of the link cell procedure.  The net result is that you have a dense matrix where you have to do expensive look-up operations.  The change to LinkCells essentially ensures that the atom positions are kept in the right order if the number of atoms in the neighboring cells is not less than the total number of atoms.  This change ensures that you also don't need to do lookup operations if your CONTACT_MATRIX is dense.  

In other words, this change ensures that we are only doing the sparse matrix lookup operations when they are actually needed.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/src/core/Value.cpp
+++ b/src/core/Value.cpp
@@ -261,7 +261,7 @@ double Value::get(const std::size_t ival, const bool trueind) const {
     plumed_dbg_massert( ival<getNumberOfValues(), "could not get value from " + name );
   }
 #endif
-  if( shape.size()==2 && trueind ) {
+  if( shape.size()==2 && ncols<shape[1] && trueind ) {
     const unsigned irow = std::floor( ival / shape[1] );
     const unsigned jcol = ival%shape[1];
     // This is a special treatment for the lower triangular matrices that are used when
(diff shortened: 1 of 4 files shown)


carlocamilloni:
@gtribello Is this covered in the regtest?

gtribello:
Yes and no.  There are regtests that check if the contact matrices are built in the right way both without and without sparsity.  However, it is diffiult to add a check to see if a matrix is being treated as sparse when it isn't sparse as the only thing that really affects is performance.  In other words, if you treat a dense matrix as if it is sparse then the code slows down massively.  I don't think we ever had a satisfactory way of testing for such things.

Issue #1452, opened 2026-09-02
Five possible defects in the matrix/vector CV framework


erikfransson:
I'm not very familiar with plumed, but I was trying it out and doing some automatic small tests to explore it further.
It found the following potential issues, listed below, so I thought I'd share in case this is useful for developers.

## Setup
Setup: PLUMED 2.10.1, `plumed driver --ixyz` and MD through the GPUMD interface, `UNITS LENGTH=A TIME=fs ENERGY=eV`, aarch64 (GH200), GCC 14.3, CUDA 13.0.
All values were checked against an independent Python implementation (matscipy neighbour lists), all derivatives with `plumed driver --debug-forces`.
Test cells:  perovskite crystals, 720-2000 atoms, orthorhombic boxes 27-47 A.

## 1. MATRIX_PRODUCT returns wrong derivatives

`s: MATRIX_PRODUCT ARG=w,wt` (w a CONTACT_MATRIX, wt its TRANSPOSE), followed by switching functions and SUM.
Values are correct.
Forces are wrong by 32 % and the virial by 82 % on a rattled crystal.
Every other action in the same chain (CONTACT_MATRIX, TRANSPOSE, matrix CUSTOM, SUM) is correct to about 3e-6.

## 2. COORDINATIONNUMBER silently drops negative switch values

`COORDINATIONNUMBER ... SWITCH={CUSTOM FUNC=sin(3.14*x)/(3.14*x)*sin(0.2618*x)/(0.2618*x) R_0=1 D_MAX=12 NOSTRETCH}` (a sinc kernel, sign-alternating).
Per-atom sums come back as if every negative pair contribution had been discarded: a channel whose correct sum is -98.9 reports +19.6.
Shifting the same function positive, `(sin(q*x)/(q*x)+0.218)*window`, reproduces the reference exactly, so the negative contributions are being dropped, not mis-evaluated.

## 3. DISTANCE_MATRIX applies CUTOFF to derivatives but not to values

`DISTANCE_MATRIX GROUPA=... GROUPB=... CUTOFF=12.0` in a 27 A box.
Every minimum-image pair contributes to downstream values, not only pairs within 12 A: an all-pair Python sum reproduces the PLUMED result to 6 decimals, the within-cutoff sum does not.
The source shows why: `DistanceMatrix::calculateWeight` returns the distance unconditionally and only skips `addAtomDerivatives` when the pair is beyond the cutoff.
So elements beyond the cutoff carry values with silently zero derivatives.

[...]

PR #1453, opened 2026-09-03
Fix 1452

gtribello:
##### Description

This is some initial work to address the comments in issue #1452.  These changes address issues 2 and 3.

On issue 2 I have changed CONTACT_MATRIX so that negative values of the switching function are not ignored. This was a matter of changing an 

```c++
if( val<epsilon )  
```

to 

```c++
if( fabs(val)<epsilon )
```

The changes to the regtest symfunc/rt-structure-factor are required because we now need to set D_MAX in such a way that the negative values of the sine are ignored to reproduce the old result.  I also added a test of a switching function which can take negative values in here.

Issue 3 has not been fixed but is now documented more clearly.  DISTANCE_MATRIX was implemented as an afterthought - CONTACT_MATRIX is expected to be the more commonly used action.  As the manual now explains, if you want to do the calculation described in the issue you should use the following input:

````
d5: DISTANCE_MATRIX GROUPA=1-10 GROUPB=1-250 CUTOFF=1.0
# Apply a switching function to determine the elements in the matrix d5
# where the distance is less than the cutoff
cut: CUSTOM ARG=d5 FUNC=switch(1-x) PERIODIC=NO
# Taking the element-wise product in the next command gives us a matrix
# where every element that is greater than the cutoff is zero.
d5cut: CUSTOM ARG=d5,cut FUNC=x*y PERIODIC=NO
# We can now calculate the average distances by multiplying these matrices by
# a vector of ones and thus summing the rows.
ones: ONES SIZE=10
totdist: MATRIX_VECTOR_PRODUCT ARG=d5cut,ones
ndist: MATRIX_VECTOR_PRODUCT ARG=cut,ones
average: CUSTOM ARG=totdist,ndist FUNC=x/y PERIODIC=NO
DUMPATOMS ATOMS=1-10 ARG=average FILE=avdist.xyz
`````

The CUTOFF keyword is setting the cutoff that should be used for the link cells.  Distances that are greater than the cutoff are still evaluated and stored.  If we set them equal to zero then the CUSTOM functions that act on the matrix elements are evaluated wrongly.  PLUMED now also outputs a warning every time some uses DISTANCE_MATRIX and CUTOFF.  The log also makes clear that the CUTOFF is a cutoff to be used in a link cell calculation and that distances larger than the cutoff will be evaluated.

__I am tempted to change the CUTOFF keyword to LINKCELL_CUTOFF to make this less misleading.  That is a syntax change though.  What do you think?__ 

##### Target release

##### Type of contribution

##### Copyright

##### Tests


carlocamilloni:
@gtribello I was checking the issue and it is related to plumed 2.10 so I think it would be worth to fix things also there where meaningful

gtribello:
This fixes the remaining problems that were identified in issue #1452.  In short:

Problem 1: PLUMED actually can't handle taking the product of two sparse matrices.  This version actually crashes if you try to do it.  This is something that I would like to work on in the future but I think that protecting users from using functionality that doesn't work will do the trick for now.

Problem 4 + 5: The underlying problem for both of the issues identified here was when you have an input like this:

````
ax: COORDINATIONNUMBER ...
   SPECIESA=1,2,3,4,9,10,11,12 SPECIESB=5,6,7,8,13,14,15,16,21,22,23,24,29,30,31,32 
   SWITCH={CUSTOM FUNC=1/(1+exp((x-2.3)/0.2)) R_0=1 D_MAX=3.0 NOSTRETCH}
....
bx: COORDINATIONNUMBER ...
    SPECIES=1,2,3,4,9,10,11,12 
    SWITCH={CUSTOM FUNC=1/(1+exp((x-2.4)/0.2)) R_0=1 D_MAX=3.0 NOSTRETCH}
....
xi: CUSTOM ARG=ax,bx VAR=ax,bx FUNC=ax+bx PERIODIC=NO
```` 

Notice function acts element wise on the two vectors of coordination numbers and that the coordination numbers are functions of different sets of atoms.  When you have an input like this one the forces are evaluated wrongly.  All the changes that Daniele and I made to the way the forces are calculated between versions 2.10 and 2.11 means that this is already fixed in version 2.11.  **I added a regtest to make sure it doesn't break again, though.**

Last thing that was pointed out was that if you use CUSTOM like this:

````
CUSTOM ARG=x VAR=e FUNC=e*e PERIODIC=NO
````

PLUMED will run but will give the wrong values because the variable that the user has tried to call e is replaced with e (the base of the natural logarithm).  I have thus added a check to ensure that CUSTOM crashes with an error whenever one of the VAR is given a name that is same as one of the constants in lepton.

The only thing remaining to address if whether we should change CUTOFF in DISTANCE_MATRIX to LINKCELL_CUTOFF.  I have checked and no-one has used this command in the nest inputs.  There are a couple of uses in the tutorials site but one is me and the other is using it through a shortcut.  Based on that making the change would have a relatively small impact but that isn't everything, obviously.

gtribello:
OK @carlocamilloni 

I think should be all the issues here sorted and the change to the syntax we discussed in the meeting this morning also sorted.

gtribello:
I would be ready to merge this if you are.

PR #1454, opened 2026-09-04
Fixes for problems 2 and 3 from issue 1452 for version 2.10

gtribello:
##### Description

This is a version of the initial work that has been done to address the problems raised in issue #1452 that can be merged into v2.10.  The work is similar to the work done in PR #1453.  I have not, however, adjusted the documentation for the DISTANCE_MATRIX action here because it is not as complete as the documentation for the master version of PLUMED. 

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
HI @carlocamilloni 

I think I have also sorted everything with that issue here as well.  I thus think if this passes the tests it can be merged into v2.10.

There then needs to be a merge of v2.10 into master but I will look after that once these two separate PRs are merged.  I know that it is a bit messy but the way the derivatives work in v2.10 and v2.11 are rather different so I want to make sure nothing from 2.10 that doesn't need to be there is transferred to master.

Thanks

PR #1455, opened 2026-09-07
Plugins documentation

gtribello:
##### Description

This is my suggestion for how to include documentation for the plugins that Daniele wrote in the PLUMED manual.  

My idea was to add a table at the bottom of the modules page that has the plugins in.  This will look like this:

plugins-page

(I can adjust the authors in the table to whatever people feel they should be.)

If you then go to the pycv page you see the following:

pycv

which is a rendered version of whatever was in the `README.md` that is in this particular plugins directory with the titles that are shared by the other pages in the manual.  I can't find things in the nest that use the plugin.  For the tutorials page, I am allowing plugin developers to use a custom search term to search through the lessons to find the appropriate tutorials.

I was looking through the README files and I think that we may want to refine them a bit with Daneiele.  I think that the one for the openaccPTM is not correct anymore.  It may be worth doing this as part of this PR.

It is also worth noting that as I did this I noticed the following announcement:

https://squidfunk.github.io/mkdocs-material/blog/2026/02/18/mkdocs-2.0/

To be clear, we are using materials for mkdocs to build the documentation so this may well affect us.  Any thoughts on what we should do here.

##### Target release

##### Type of contribution

##### Copyright

##### Tests

### DIFF
--- a/new-manual/build_manual.py
+++ b/new-manual/build_manual.py
@@ -460,7 +460,7 @@ LOAD FILE=/path/to/myplugin.so
 
 Once the library is loaded you can then use the functionality that is contained within it in your PLUMED input.
 
-Typically, functionality is included in a plugin rather than a library special compilation options are required to build the associated functionality. For example,
+Typically, functionality is included in a plugin rather than a module when special compilation options are required to build the associated functionality. For example,
 much of the code that allows you to run PLUMED on GPUs is provided through plugins because additional compilation options are required. The code in these plugins is 
 also more experimental than the code in the modules. This sadly often means that it is tested less frequently.  Consequently, if you do try to use these plugins for production calulculations,
  you are strongly recommended __to run the test suite after compilation__.  To run these test suits you change to the `regtest` directory that will be contained in the plugin directory.  Once in that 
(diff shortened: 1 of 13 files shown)


carlocamilloni:
@gtribello I think this is very nice and clear. I would also add a mention to this subsection in the introduction here: Introduction

PLUMED Version 2.11

[...]

[PLUGINS]

Please also note that some developers prefer not to include their codes in PLUMED. To use functionality that has been written by these developed you can use the [LOAD](https://www.plumed.org/doc-master/user-doc/html/LOAD/) command.

[...]

gtribello:
OK @carlocamilloni 

I added the following text on the index page:

_There are also some features that are released within PLUMED as plugins. You can read about these plugins by scrolling to the bottom of [this page](modules.md). The code within these 
plugins is compiled separately to PLUMED. The dynamic library that results compilatiing a plugin can then be loaded at runtime using the [LOAD](LOAD.md) command._

carlocamilloni (approved):

gtribello:
I have another commit on this as I have also rewritten the documentation for the openaccPTM plugin.  I think that Daniele didn't have time to do this before his position finished.  I have just sent him my draft.  If he says it is OK then I will add it here and we can merge.

gtribello:
OK I think this is done now.

PR #1460, opened 2026-09-18
Forces in the style of CudaCoordination for parallel task manager stuff

gtribello:
##### Description

I have rewritten the force calculation in the parallel task manager for actions that involve calculating all the elements of a matrix in the way that Daniele implemented the forces in the CudaCoordination.  This means that we can avoid using lots of memory to store multiple copies of the forces.  My hope is that this will make computing Steinhardt parameters on the GPU feasible.  At the moment, I think Daniele and I concluded it wasn't.

To be clear, currently you activate this new way of caluclating the forces by using a flag. If that flag is not present the forces are calculated in the way that they used to be computed.  This change thus has no negative effects on performance.  In fact, while I was going through the process of changing the code I found some things that could be optimised.  Consequently, as you can see from the graph below the newer version of the code is faster than the old one:

results-single-node

I am computing the local steinhardt parameter example from the J Phys Chem tutorial.  The black line are the timings for the current master and the red line is the timings for the code in this PR.

Now with the current master I ran on 4 threads (on my laptop) with only this version of the code.  The black line below is the old implementation of the forces.  The red line is what you get when you use the new flag to turn on the force gathering in the style of the cuda coordination.  

results-4-nodes

Unfortunately, the new implementation is slower than the new one with OpenMP threading.  At the same time, the calculations take longer if I use openMP threads on my laptop, which makes no sense.  I thus want to try all this on the CPU and GPU of a super computer.

**Oh... I also fixed the product of two sparse matrices.  Now if you have a CV that is a product of two sparse matrices together the forces are computed correctly.**

##### Target release

##### Type of contribution

##### Copyright

##### Tests


gtribello:
Hello @carlocamilloni, @GiovanniBussi and @maxbonomi 

I am a bit stuck with this PR in two regards and I wondered if you had any ideas that might help. The two problems are as follows:

1. Currently, there are a lot of failures for github actions when the code is compiled with Intel.  I have tried to compiling this code with the intel compilers that are currently installed on the supercomputers here.  I cannot reproduce the errors that are seen on GitHub on my local machines, though.  I am just using this configure command:

```bash
./configure --enable-modules=all CXX=icpx CC=icx FC=ifx
```

is there any other flag that I should be adding there to reproduce the configuration that is being used on GitHub?  I would also note that we don't have the latest intel compiler on the machines here so it may something to do with the particular version of the intel compiler we are using.  If anyone has any ideas for reproducing the configuration locally it would be really useful to hear them.

2. I want to test whether I have broken Daniele's implementation of the parallel task manager on the GPUs.  I am also hoping that the work that I have done may have made it work better.  The problem I am having is that I need to request an interactive job on one of the GPU nodes of the local supercomputer.  Everytime I have tried this so far my job has sat in a queue for 2-3 hours and then I have gotten frustrated and given up.  Can you folks give me access to a machine that would allow me to get this task done a little quicker?

What I have found out thus far, though.  

1. I have managed to speed up the CPU implementation of the local q6 parameter by a tiny bit.   

old_vs_new_timings

So the black, red, blue and magenta lines are for calculations done on 1, 2, 4 and 8 threads respectively.  The dashed lines are the timings for the current master.  The solid lines are the timings for this commit.

2. This speed up came about because I found places where I was adding zeros to forces and because I redid the matrix multiplication in a way that ensures that the elements that are needed for doing the multiplication are contiguous in memory.  I can use a flag to use the new code I wrote for calculating the forces in the way Daniele did in CudaCoordination.  This graph shows the effect of using this flag when doing a calculation on the CPU:

ntd_vs_standard

The colours are the same as for the last figure here and the dashed lines here don't use the flag.  You can see that the calculation is quickest if you don't use the CudaCoordination approach for calculating the forces.  This may be different on the GPU, though, as I think the reduction is more expensive.

